PR #75, opened 2015-02-19
Update oxidized.gemspec

funzoneq:
Require rugged for lib/oxidized/output/git.rb

### DIFF
--- a/oxidized.gemspec
+++ b/oxidized.gemspec
@@ -17,4 +17,5 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'asetus',  '~> 0.1'
   s.add_runtime_dependency 'slop',    '~> 3.5'
   s.add_runtime_dependency 'net-ssh', '~> 2.8'
+  s.add_runtime_dependency 'rugged',  '~> 0.21.4'
 end


ytti:
What if user does not use git output? I've not made my mind on this, but should we require that everything that might be needed is required? Or should we have minimal dependencies in gemspec and optional dependencies are checked runtime?

funzoneq:
What's the use of a package manager if you don't use it? I expect gem to install all the requirements of the app. And if it's not used in some cases, then that's not a problem for me. In that case, the gem will just take up some space on my disk.

ytti:
To continue package manager analogy, shouldn't apt only install software you need, not software you might need? Rugged isn't simplest to install, as it requires working compile environment and few libraries. Perhaps user who does not have full gcc stack installed is not thrilled to install them, just to satisfy install requirement which does not match her runtime requirement?

Again, I'm undecided here. It used to be like you propose, but I got some complaints for the other direction and made it this way, where the program barfs and tells you what to install. 
I can easily go back to original behavior if it is what majority wants.

fhibler:
I agree with funzoneq. It should install all software which is required to use every feature included in the software.

PR #102, opened 2015-04-13
Add the Cumulus feature

FlorianDoublet:
I added the Cumulus feature, where I run all the commands in the same block thanks to "launch", because the is lot of command to run.
For the same reason, I added comment for a clearer config file.

### DIFF
--- a/lib/oxidized/model/cumulus.rb
+++ b/lib/oxidized/model/cumulus.rb
@@ -1,0 +1,76 @@
+class Cumulus < Oxidized::Model
+  
+  prompt /^((\w*)@(.*)([>#]\s)+)$/
+  comment  '# '
+  
+  
+  #add a comment in the final conf
+  def add_comment comment
+    "\n###### #{comment} ######\n" 
+  end
+
+  cmd :all do |cfg|
+    cfg.each_line.to_a[1..-2].join
+  end
+  
+  #show the persistent configuration
+  pre do
+    cfg = add_comment 'THE HOSTNAME'
+    cfg += cmd 'cat /etc/hostname'
+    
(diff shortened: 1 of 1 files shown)


ytti:
Great! Some comments.


You can run 'cmd' directly without block:

https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/timos.rb#L34

Also why are you running comment via shell printf? Why not just generate the string in ruby?

def add_comment comment
  "##### #{comment} #####\n" 
end

pre do
  cfg   = '"'
  cfg += add_comment 'THE HOSTNAME'
  cfg += cmd 'cat /etc/hostname'
  ...
  cfg
end


You might also want to check if it supports exec in ssh (it's openssh, so it should):
https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/junos.rb#L47

This means, shell is never started for SSH just bunch of exec channels (one for each command).

FlorianDoublet:
Thanks for your comments, it's corrected except for the "exec". If I set it to true, it doesn't work anymore.

ytti:
aah shame, worth a shot. Really curious they managed to break exec channels, since I've understood it's just OpenSSH. Well, no matter, let's not use it then :)

ytti:
I think we're sending empty command to your Cumulus Linux via cmd ''

can you try to wrap it inside pre or post instead?

pre do
  ...
end

or 

post do 
  ..
end

(their function is that if you have other 'cmd', anything in pre appears on top of config, and anything in post appears in bottom of config)

I think this should work, and if so, it removes one unnecessary empty command being send to your devices.

FlorianDoublet:
Thanks a lot, it work and it's better. Corrected.

PR #103, opened 2015-04-14
Add a pre_logout for some devices

FlorianDoublet:
On an ironware device, 'logout' is not recognized, but also with just one " pre_logout 'exit' " it doesn't work.
I don't know if add two pre_logout is the best way to handle it, but actually I don't see an other way to do it.
Added an other post_login to handle pager.


ytti:
I think this is fine. Boxes will complain about commands they don't recognize, but I don't see why this would be fundamentally issue.
If someone disagrees, we welcome pull request which solves it.

PR #104, opened 2015-04-16
Add support for HP Comware SMB switches

sid3windr:
Some SMB HP switches (V1910, V1920) run Comware, but have the CLI hidden behind a "secret" toggle - it needs to be activated by _cmdline-mode on. The password to enable it varies per model, so this patch introduces a new var, comware_cmdline to contain the password.

Unfortunately we cannot easily autodetect this, as only the "summary" command can give us the model number, but this command is paged and we cannot disable the pager until we are in the full CLI.

Tested on HP V1910-24G-POE.

### DIFF
--- a/lib/oxidized/model/comware.rb
+++ b/lib/oxidized/model/comware.rb
@@ -21,6 +21,17 @@ class Comware < Oxidized::Model
   end
 
   cfg :telnet, :ssh do
+    # enable command-line mode on SMB comware switches (HP V1910, V1920)
+    # autodetection is hard, because the 'summary' command is paged, and
+    # the pager cannot be disabled before _cmdline-mode on.
+    if vars :comware_cmdline
+      post_login do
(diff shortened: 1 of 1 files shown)


ytti:
Yeah, if someone is bothered by this approach, they can discover it via 'summary' and page via 'expect' and send PR. Thanks!

ujemvi:
Hi guys!
I'm having trouble with a SMB switch that fits right here. It's a V1910-24G.
The system connects well to the switch, but it seems that it's not using the '_cmdline-mode on' command. All I got when I look at the config is this:
```
#          ^
#  % Unrecognized command found at '^' position.
#          ^
#  % Unrecognized command found at '^' position.
         ^
 % Unrecognized command found at '^' position.
```
Any idea?

ytti:
You probably should open issue and put @sid3windr there

PR #106, opened 2015-04-17
Changed prompt (too strict)

FlorianDoublet:
For some devices, prompt is too strict (the begin of the prompt could be BR-SSH@ ... for example)


ytti:
I feel this is too liberal, maybe \S?

FlorianDoublet:
Why not, but in fact I just saw that I've got a switch with this prompt : " SWSRV10-PIC-RB01# " so something like \S.*[#>] should be better ?

PR #107, opened 2015-04-20
Added a less strict prompt

FlorianDoublet:
Match with more devices

### DIFF
--- a/lib/oxidized/model/ironware.rb
+++ b/lib/oxidized/model/ironware.rb
@@ -1,6 +1,6 @@
 class IronWare < Oxidized::Model
 
-  prompt /^(\S*)([>#])+$/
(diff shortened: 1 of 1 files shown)


ytti:
I think the proposed:
prompt /^(\S*)@?([\S\s])+([>#])+\s?$/

Is more complex way to say:
prompt /^.+[>#]\s?$/                                                                  

I believe those are equivalent, if not. I'd be curious to see example when they are not. I.e. when would  the simpler match, where the more complex would not.

FlorianDoublet:
That's right, corrected.

ytti:
wup, not exatly strict, but we have some other platforms too where there are quite liberal ways for format to be.

PR #114, opened 2015-04-29
add feature about versions and diffs for oxidized-web

FlorianDoublet:
this pull request is for this pull request in oxidized-web : https://github.com/ytti/oxidized-web/pull/11

### DIFF
--- a/lib/oxidized/nodes.rb
+++ b/lib/oxidized/nodes.rb
@@ -148,6 +148,35 @@ module Oxidized
         end
       end
     end
+    
+    public
+    
+    def version node, group
+      with_lock do
+        i = find_node_index node
+        output = self[i].output.new
+        raise Oxidized::NotSupported unless output.respond_to? :fetch
+        output.version node, group
+      end
+    end
+    
+    def get_version node, group, oid
+      with_lock do
+        i = find_node_index node
+        output = self[i].output.new
+        raise Oxidized::NotSupported unless output.respond_to? :fetch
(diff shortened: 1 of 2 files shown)


ytti:
It looks interesting, but I'm not sure if this is the direction we want to take. 

Also the commit it not against the latest github version, if you look at the git.rb it's the 'single_repo' configuration option (it decides if groups are separate repos or directories in repo).

FlorianDoublet:
updated.

laf:
Is this something that is being considered for merging? It would be handy to have diff support within oxidized-web.

ytti:
I worry about the complexity added in nodes.rb and git.rb, while I also view that most people would do this in their usual git frontend.
But perhaps lot of people actually dont have gitlab, github, stash and no wish to install.

So it feels like I may be alone in my opposition. If so, I am perfectly ok that it is integrated. @nopedial can do it, when he has time to look into it.

Id like to refactor it maybe a bit, but as I wont have time for it any time soon, Im not gonna block integration as-is.

laf:
Understand the concerns :)

I'd certainly vote to include this but happy either way, just don't want to spend time setting something up if this was to be merged in :)

SaaldjorMike:
Would certainly be a nice feature :+1:

Rosiak:
Would really like this too! :+1:

anden4380:
:+1:

darkounet:
:+1:

PR #117, opened 2015-05-06
Added support for http as a source

laf:
This is my first experience with Ruby so I'm sure this won't be correct so any feedback is appreciated :)

Adds support for http as a source so that lookups of devices can be done against an API. Maintains all the mappings as per other sources (I hope).


ytti:
I think it's fine, I might myself perhaps separated more of the load method to separate methods, so that each method is smaller, does less and is easier to read through.
But by no means is this problematic PR. Thank you,

PR #122, opened 2015-05-24
Remove 'show inventory' from Cisco SMB model, since they do not recog…

SaaldjorMike:
…nize the command

```
SWITCH#show inventory
% Unrecognized command


ytti:
Merging this as @thetamind didn't reply.

PR #129, opened 2015-06-10
Add log for update at /var/log/oxidized-update for git output

FlorianDoublet:
Add update for a node in log file

### DIFF
--- a/lib/oxidized/output/git.rb
+++ b/lib/oxidized/output/git.rb
@@ -104,7 +104,9 @@ class Git < Output
         :parents    => repo.empty? ? [] : [repo.head.target].compact,
         :update_ref => 'HEAD',
       )
+      
       index.write
+      true
     end
   end
 end
--- a/lib/oxidized/worker.rb
+++ b/lib/oxidized/worker.rb
@@ -37,8 +37,10 @@ module Oxidized
         msg = "update #{node.name}"
         msg += " from #{node.from}" if node.from
         msg += " with message '#{node.msg}'" if node.msg
-        node.output.new.store node.name, job.config,
+        if node.output.new.store node.name, job.config,
                               :msg => msg, :user => node.user, :group => node.group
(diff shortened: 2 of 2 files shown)


ytti:
Unsure if this is the correct approach, this is not output module agnostic.

I think we want to make #store return false if it didn't update config and true if it did. Then in: https://github.com/ytti/oxidized/blob/master/lib/oxidized/worker.rb#L40

We can do

if node.output.new.store ...
  Log.info "Configuration updated for #{node.name}"
end

At which point output model only needs to conform to this concept, and it'll support logging.

Also. Why not use existing logging infrastructure which logs into syslog? Why implement separate files? I would greatly favour having everything consistently via syslog.

FlorianDoublet:
Yes sorry, I wanna put it in an other file to get only the update but in fact it's true that we don't care.
Updated

ytti:
I don't think you actually need to change the git.rb at all.

We are doing

if ...
  mangle
end


This means we're by default returning true or false from 'if' statement as last value. So I think it should work without touching git.rb, please test.

FlorianDoublet:
I look at it but it don't work without the returns

FlorianDoublet:
maybe because if there is an update the last value will not be the "if" value but " index.write " ?

ytti:
Only reason why it might not work, is if 'index.write' returns nil or
false. So adding one line 'true' in the if block should fix if that is the
case:

if blaablaa
  blaablaa
  true
end

(no need for false, or return)

On 10 June 2015 at 12:42, Florian Doublet <notifications@github.com> wrote:

> maybe because if there is an update the last value will not be the "if"
> value but " index.write " ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ytti/oxidized/pull/129#issuecomment-110673642>.
>



-- 
  ++ytti

FlorianDoublet:
True. it's a java reflex to set return, sorry.
Updated

PR #130, opened 2015-06-10
lib/oxidized/jobs.rb:29 NaN (FloatDomainError)

funzoneq:
```
sudo -u oxidized -H /usr/local/bin/oxidized -d
NaN
/usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/jobs.rb:29:in `to_i': NaN (FloatDomainError)
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/jobs.rb:29:in `new_count'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/jobs.rb:25:in `duration'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/jobs.rb:13:in `initialize'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/worker.rb:7:in `new'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/worker.rb:7:in `initialize'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/core.rb:19:in `new'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/core.rb:19:in `initialize'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/core.rb:11:in `new'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/core.rb:11:in `new'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/cli.rb:9:in `run'
	from /usr/local/share/gems/gems/oxidized-0.7.1/bin/oxidized:9:in `<top (required)>'
	from /usr/local/bin/oxidized:23:in `load'
	from /usr/local/bin/oxidized:23:in `<main>'
```

### DIFF
--- a/lib/oxidized/jobs.rb
+++ b/lib/oxidized/jobs.rb
@@ -26,7 +26,7 @@ module Oxidized
     end
 
     def new_count
-      @want = ((@nodes.size * @duration) / @interval).to_i
+      @want = (@nodes.size * @duration) / @interval
       @want = 1 if @want < 1
       @want = @nodes.size if @want > @nodes.size
       @want = @max if @want > @max
--- a/lib/oxidized/jobs.rb
+++ b/lib/oxidized/jobs.rb
@@ -10,7 +10,7 @@ module Oxidized
       @nodes     = nodes
       @last      = Time.now.utc
       @durations = Array.new @nodes.size, AVERAGE_DURATION
-      duration AVERAGE_DURATION
+      @duration  = AVERAGE_DURATION
       super()
     end
 
@@ -26,7 +26,7 @@ module Oxidized
     end
 
     def new_count
-      @want = (@nodes.size * @duration) / @interval
+      @want = ((@nodes.size * @duration) / @interval).to_i
       @want = 1 if @want < 1
       @want = @nodes.size if @want > @nodes.size
       @want = @max if @want > @max


ytti:
Under what situation does this arise? When interval is configured as "123"? I'm happy to merge, just curious.

funzoneq:
@ytti Actually in my case ```@duration``` turned out to be NaN.

ytti:
Do we know why? What was configured interval?

Right now the #to_i fix will just set it 0, I think we'll still have issue then.

funzoneq:
@ytti `@duration` turned out to not be initialised. I fixed it in this PR. But now there is an error with:

```
* Environment: developmentundefined method `<=' for nil:NilClass
* Listening on tcp://0.0.0.0:8888
/usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/jobs.rb:42:in `work': undefined method `<=' for nil:NilClass (NoMethodError)
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/worker.rb:15:in `work'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/core.rb:37:in `run'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/core.rb:30:in `initialize'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/core.rb:11:in `new'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/core.rb:11:in `new'
	from /usr/local/share/gems/gems/oxidized-0.7.1/lib/oxidized/cli.rb:9:in `run'
	from /usr/local/share/gems/gems/oxidized-0.7.1/bin/oxidized:9:in `<top (required)>'
	from /usr/local/bin/oxidized:23:in `load'
	from /usr/local/bin/oxidized:23:in `<main>'
```

```
def work
      # if   a) we want less or same amount of threads as we now running
      # and  b) we want less threads running than the total amount of nodes
      # and  c) there is more than MAX_INTER_JOB_GAP since last one was started
      # then we want one more thread (rationale is to fix hanging thread causing HOLB)
      if @want <= size and @want < @nodes.size
        @want +=1 if (Time.now.utc - @last) > MAX_INTER_JOB_GAP
      end
    end
```

I think this happens because ```size``` doesn't exist. And by the way I read it, it should be the amount of threads?

ytti:
The error is complaining that we're calling method <= for nil value.

So @want <= size

Is just syntactic sugar for @want.<=(size)

<= being method name, and size being argument for said method.

What the traceback tells, we're calling #size method for object whose class is Nil.

Simply @want is nil, it should not be.

ytti:
Your duration fix definitely is broken, you're not calling the method at all.

'duration' is method which we are calling, now you're assigning. 

'@duration' is variable.

when we do

duration ARGUMENT

we're passing ARGUMENT to the

def duration XYZ
  ..
end

as you can see @duration is assigned in that method. So your fix unfortunately breaks something which is working, it does not explain your issue.

ytti:
Any update on this? I can't accept this pull request as it'll break stuff.

funzoneq:
Can't reproduce myself anymore. Weird.

PR #131, opened 2015-06-13
For JunOS, trigger on commit complete instead of commit started

cfiske:


cfiske:
Hmm, in hindsight I guess you lose the ability to grab the username, even though the completion message is a more reliable trigger point.

ytti:
Also if the config isn't actually changed, then the git backend won't commit, as it compares the hash. So I think we're good?

cfiske:
So I'm going back and forth on this in my mind.

My original reason for considering the change is that, if we log in and pull the config before the commit finished, we don't get the new config (or have to handle that situation in the code somehow to make it wait for it). For Juniper SRX firewalls with complex rulesets a commit can take quite some time to finish.

In terms of the username the 'show config' output on Juniper names the last person to make a configuration change anyway, so we could just grab it from there and have that info whether it was triggered by a syslog message or not.

ytti:
If worry is

a) commit happens
b) oxidized is asked to poll
c) oxidized gets config
d) commit finishes

that is, oxidized won't see the change, because it was too fast. Then I think we could add state to the syslog monitor, that it'll grab username from current message, then it'll wait until commit finishes, and then triggers oxidized only after it has finished. Meaning we still get username from syslog, but we won't call oxidized until it has finished,

PR #139, opened 2015-06-24
Updated nagios_check_failing_nodes.rb

darkounet:


darkounet:
when oxidized haven't backup a switch, the node['last']['status'] doesn't exist

Issue #143, opened 2015-07-05
fetch fails when single_repo: true and groups are used


jwhited:

ytti:
fixed by 513e137

PR #144, opened 2015-07-05
fix fetching when groups are used and single_repo: true

jwhited:
https://github.com/ytti/oxidized/issues/143

### DIFF
--- a/lib/oxidized/output/git.rb
+++ b/lib/oxidized/output/git.rb
@@ -50,13 +50,17 @@ class Git < Output
   def fetch node, group
     begin
       repo = @cfg.repo
-      if group
+      if (group && !@cfg.single_repo?)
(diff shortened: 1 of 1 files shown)


ytti:
I would probably have written it something like this:

  def fetch node, group                                                                                                
    repo = @cfg.repo
    repo = File.join File.dirname(repo), group + '.git' if group and not @cfg.single_repo?
    repo = Rugged::Repository.new repo
    index = repo.index
    index.read_tree repo.head.target.tree unless repo.empty?
    file = node
    file = File.join(group, node) if group and @cfg.single_repo?
    repo.read(index.get(file)[:oid]).data
  rescue
    'node not found'
  end
                                                                                                                      
Not tested though. I'm happy to accept this version too, depending on your preference

jwhited:
sure, i'll test and update

PR #145, opened 2015-07-06
fix fetching when groups are used and single_repo: true

jwhited:
https://github.com/ytti/oxidized/issues/143


ytti:
Thanks Jordan!

Issue #115, opened 2015-05-02
Please add support for UBNT EdgeMax devices


seitzbg:
Please add support for UBNT EdgeMax devices.

I can provide a login if needed to one for development.

ytti:
Hit me up at saku@ytti.fi with login details and I'll see what I can do.

ytti:
Still waiting for login details

ytti:
Fixed by  5f89276

PR #146, opened 2015-07-09
Added support for EdgeOS

laf:
Adds support for #115.

Only been able to test this on ER-X but these are just based on Vyatta so taken the code from that model.


ytti:
Thanks! Closes #115

PR #152, opened 2015-07-30
Store stats in SQL, and use it to re-build ring queue. Need sqlite3.

FlorianDoublet:
Store stats in SQL, and use it to re-build ring queue. Need sqlite3.

### DIFF
--- a/lib/oxidized/config.rb
+++ b/lib/oxidized/config.rb
@@ -28,6 +28,7 @@ module Oxidized
   CFGS.default.rest          = '127.0.0.1:8888' # or false to disable
   CFGS.default.vars          = {}             # could be 'enable'=>'enablePW'
   CFGS.default.groups        = {}             # group level configuration
+  CFGS.default.database      = 'oxidized_database.db'
 
   CFGS.default.input.default    = 'ssh, telnet'
   CFGS.default.input.debug      = false # or String for session log file
--- a/lib/oxidized/job.rb
+++ b/lib/oxidized/job.rb
@@ -4,11 +4,12 @@ module Oxidized
     def initialize node
       @node         = node
       @start        = Time.now.utc
+      
       super do |node|
         @status, @config = node.run
         @end             = Time.now.utc
         @time            = @end - @start
-      end
+      end  
     end
   end
 end
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -83,7 +83,7 @@ module Oxidized
         :ip        => @ip,
         :group     => @group,
         :model     => @model.class.to_s,
-        :last      => nil,
+        :last      => last,
         :vars      => @vars,
       }
       h[:full_name] = [@group, @name].join('/') if @group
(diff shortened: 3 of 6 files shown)


ytti:
Quickly looking I'm not sure we're threading anymore, as the collection appears to be outside the thread in job.rb, I think it'll block until finished then thread just collects information from hash.

I'm also wondering if we should use Sequel ORM instead of raw SQL.

FlorianDoublet:
I know what you mean. It's a good question, but after all, it's the tread who call Store, so are you sure we're not threading anymore ? And is there a way to be sure ? 
I can calculate it just as before and then send it to Store, but it was to minimize the lines.

I can't tell you if Sequel ORM is better than raw SQL, I don't know it. But SQLite is a kind of reference for program database, no ?

ytti:
I don't think it's thread which calls the store. Only the 'super do' part is threaded, outside it, it'll block, I'm pretty sure.
There is way to be sure, set rotation time so small that single thread won't do it. And with debugging on, it should report how many threads running and how many wanted.

FlorianDoublet:
Hum true, but if it's just the problem moving the call of Store into super should fix it. I will check for it

ytti:
SQLite also itself does not support multiple writes, so the code probably isn't thread safe, as many threads try to connect to the same SQLite database, the one doing it right now, will block it.
It might be better for thread to just return config/status, and then the main thread handle all the SQL work alone, then there is no competition to the SQL data.

Except of course if WWW side wants to write there at the same time, then you'll need to implement mutex locking.

FlorianDoublet:
What do you call "main thread", the class Core ?

ytti:
Main thread is the thread which starts all the jobs, walks through job list etc, essentially the worker.

It checks if enough jobs are running, collects stopped jobs and handles their output. This worker could handle the config storing of threads.

FlorianDoublet:
Ok thanks a lot. And so, we don't need a mutex I guess if Worker handle sql injection ?

ytti:
No. Unless oxidized-web tries to write to the SQL as well, then we'll need to mutex locking there.

FlorianDoublet:
Updated, hope it's better

MrRJ45:
@ytti @FlorianDoublet

Does this need testing? If so I will implement the changes tomorrow.

FlorianDoublet:
It's normally alright.

ytti:
I've been lazy, I've not really reviewed the latest version. I'll try to get to it. I also kind of would like to refactor the oxidized core stuff (node, nodes, worker, manager...) but it seems unlikely I'll find time any time too soon.

MrRJ45:
I've just implemented this, so I'll let you know of any issues :)

Only problem so far is it couldn't create the SQLite DB, I changed config.rb to this and that resolved it:

```ruby
CFGS.default.database      = File.join(Config::Root, 'oxidized_database.db')
```

FlorianDoublet:
Oh sorry,  I didn't have this issue..  Thanks for your  review!

MrRJ45:
I guess its probably fine it your DB already exists, I didnt create it first, so I think the issue is only when it tries to create the empty DB :)

MrRJ45:
@FlorianDoublet Just to give you a heads up I am seeing an issue with this patch at the moment, see Issue #177

ytti:
I really should focus on this. But I can't accept it right now. This is not the implementation I'm exactly looking for. Sorry, I know it was lot of work. Direction is where I want to go, but implementation details are not.

danilopopeye:
@FlorianDoublet thanks again for the time and code 👏 I'm closing this as is too outdated to be merged easily 😐

PR #153, opened 2015-08-06
Using socket instead of resolv

casdr:
I found there were some problems with local devices and DNS. I replaced Resolv with IPSocket and it worked.

### DIFF
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -1,6 +1,6 @@
 module Oxidized
-  require 'resolv'
   require 'ostruct'
+  require 'socket'
   require_relative 'node/stats'
   class MethodNotFound < OxidizedError; end
   class ModelNotFound  < OxidizedError; end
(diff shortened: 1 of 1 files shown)


ytti:
This does not make sense to me.
http://ruby-doc.org/stdlib-1.9.3/libdoc/resolv-replace/rdoc/IPSocket.html

```ruby
               # File resolv-replace.rb, line 8
def getaddress(host)
  begin
    return Resolv.getaddress(host).to_s
  rescue Resolv::ResolvError
    raise SocketError, "Hostname not known: #{host}"
  end
end
```

it should be doing exactly the same thing. Except perhaps with choosing DNS server. 
I'm also uncertain if it now raises different errors or same, here the raised error is used:

```
./oxidized/nodes.rb:          rescue Resolv::ResolvError => err
```


Can you try the old code, but without calling #new? Just Resolv.getaddress?

itscassa:
Can you try the old code, but without calling #new? Just Resolv.getaddress?

Nope, didn't work. I'm having the same issue.

ytti:
What error IPSocket raises when it can't resolve?

ytti:
Bump? What error do you get? What kind of problems? I'm hesitant to make this change until I understand what is broken.
Can you replicate the problem in irb/pry shell?

# irb
require 'resolve'
Resolv.new.getaddress 'google.com'

ytti:
bump

supertylerc:
@casdr, any update?  What are the issues with "local devices"?  What is a local device?  Are you talking about hosts stored in `/etc/hosts`?

casdr:
Seems like this was fixed, see #314

PR #156, opened 2015-08-07
Added Fortigate Foundry support

itscassa:


ytti:
We try to call the models by the operating system name. As vendor can have very many operating systems, it's confusing to call them by vendor name. Is 'foundry' really name of the operating system? Foundry was the company Brocade bought.

itscassa:
It is actually FortiOS, but different from the current one.

ytti:
There are two FortiOS? Confusing.

Are both still being developed is or one of them retired and another one still developed?

Is there way to differentiate them? Is there version number difference?

@aakso might also know.

PR #167, opened 2015-08-26
Add support for MRV Master-OS

kwibbly:
This model adds support for MRV Master-OS on Opti-Driver with software release 5.5.1.  I don't have access to other platforms running Master-OS to check, but it should probably work there as well.


RichiH:
@ytti If you have questions regarding this, you can also poke me.

ytti:
looks good, merging

Issue #39, opened 2014-08-12
Persistent `Last Status` and `Last Time` in web interface


RichiH:
Can you cache those values and re-use them so they are not lost when restarting `oxidized`?

will-h:
Data persistance-across-restarts would be super-handy (I just came here to open a ticket about this ;). I have a service check checking for each node to ensure we have a recent copy of the config, and of course it alerts if Oxidized has been restarted, since return time is now 'never'. 

Of course, the webinterface could return mtime of the device's config file, but this is probably not a great solution.

ytti:
The status for each node includes last success and last fail, so you could skip alarm if a) both are never or b) success is less than 2-3 * iteration cycle, this might be sufficient workaround for now.

But still, I think this is relates to the most important refactor oxidized needs, to move state from memory into sqlite or similar. It's not hard thing to do, and brings other benefits to us as well.

RichiH:
Jftr, doing two factor alarming is possible but more effort and less
resilient.

Richard

Sent by mobile; excuse my brevity.

darkounet:
:up:

darkounet:
or maybe a way to store state in a .yml file or .json ?

darkounet:
Any update ? :D

ytti:
Nope, sorry. I'm not working on this currently. I want to have node queue, node statistics etc in disk, not in memory. It's probably gonna take me 3-5 working days, which I'm not ready to contribute at present.

MrRJ45:
Having rewritten some bits I now have all of a nodes stats in the db. I am ignoring the max of 10 stats, do we want to keep stats in the db for a long time?

What else do we think needs to be persisted?

MajesticFalcon:
BUMP

ytti:
This is not gonna happen before I'll refactor the core, and I have no idea when that might happen.

casdr:
Is there any progress / planning on this? We have around 500 devices in Oxidized and this will help debugging and testing a lot.

PeepOks:
Is there any ETA. for this feature. It would be very helpful if last-status is not lost after oxidized service reload.

ytti:
I have no plans to work on this right now. It's important, but mainly the stats keeping would be side-effect of important change of moving state to sqlite, reducing memory footprint significantly in larger installations.

Issue #177, opened 2015-09-06
Jobs Problem?


MrRJ45:
I'm seeing an issue at the moment where nodes have "never" against them, and they wont ever backup.

If I manually start the backup then it works with no problems. It looks like no threads are running as I dont see "jobs wanted" in the log.

Any ideas where to start debugging this? :(

ytti:
Have they been attempted unsuccessfully? After retries failed nodes are moved back of the queue, and won't be tried until iteration starts again. 
What is your interval? Are they not attempted even after interval passes?

MrRJ45:
They haven't been attempted at all, "last => nil" and they are successful when I manually pull the config.

Interval is 14400 - every 4 hours, assuming that is in seconds!?

It may have broke when I reinstalled the server, I cant see why though.

ytti:
Yeah that is seconds. Strange. Can you recreate it easily, if so, can you give me access to the environment where the issues can be recreated?

MrRJ45:
It wouldnt be that easy to set it up, but I can look if need to.

Running pry on jobs.rb I see this, is that normal?

```
From: /var/lib/gems/2.1.0/gems/oxidized-0.7.2/lib/oxidized/jobs.rb @ line 41 Oxidized::Jobs#work:

    36: def work
    37:   # if   a) we want less or same amount of threads as we now running
    38:   # and  b) we want less threads running than the total amount of nodes
    39:   # and  c) there is more than MAX_INTER_JOB_GAP since last one was started
    40:   # then we want one more thread (rationale is to fix hanging thread causing HOLB)
 => 41:   binding.pry
    42:   if @want <= size and @want < @nodes.size
    43:     @want +=1 if (Time.now.utc - @last) > MAX_INTER_JOB_GAP
    44:   end
    45: end

[1] pry(#<Oxidized::Jobs>)> print @want
1=> nil
[2] pry(#<Oxidized::Jobs>)> print @size
=> nil
```

ytti:
​Now that does not seem normal. IT should always be fixnum. Something is
off.

I think if count of nodes is 0, it currently is broken. But might be
different issue.

I think I saw jobs crash under condition where I had 1 node in router.db
and it was referring to model which did not exist. I wish I had created an
issue, as I don't remember details anymore.
​

MrRJ45:
That was a test I did with a CSV file, and two fake devices, but it did try those despite pry showing that.

I'm going to have a look through my DB to see if I can see anything wrong.

[...]

PR #180, opened 2015-09-12
Last Stats Persistance

MrRJ45:
As mentioned in Issue #177 I have re-written the storage for persisting node last stats. This should hopefully close Issue #39 if this code is approved :)

This has been working fine for a few days, and I've had no issues. I am still learning Ruby as I go so there may be better ways of doing some of these things.

Firstly, this checks for a SQL Source config. If this exists we use the same connection details, and Sequel to connect. We use the table name of the devices table + '_stats' to store this data. If there is no SQL config, we fall back to storing in memory, in a class instance (the best way I could figure of doing this).

Node::Stats has some new functions, the main two are fetch and store. These do the work needed for retrieving or persisting the data.

On the node class, "last" and "last=" call these fetch and store functions. I've tried to implement this with minimal changes to the code.

One positive effect seems that when storing in a database that the DateTime is still UTC, but it now includes the servers timezone, eg. +0100 for the UK at the moment. This may help with Issue ytti/oxidized-web#16

I've also added the attribute "force" to a node, so we now set that if we want a node to run next. This is because we dont really want to updating the database to nil to force a refresh.

Comments, and thoughts? :)

### DIFF
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -111,7 +111,7 @@ module Oxidized
         @stats.store job
       end
     end
-    
+
     def reset
       @user = @msg = @from = nil
       @retry = 0
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -19,7 +19,8 @@ module Oxidized
       @auth           = resolve_auth opt
       @prompt         = resolve_prompt opt
       @vars           = opt[:vars]
-      @stats          = Stats.new
+      @stats          = Stats.new @name, @group
+      @force          = false
       @retry          = 0
 
       # model instance needs to access node instance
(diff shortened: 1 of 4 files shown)


ytti:
I guess the original code covered more. My intention is to move most of runtime data to sqlite. Like node queue, so as node size increases, we don't increase memory use, just disk use. And when we stop and restart, we'll continue where we left of, instead of start over.

But I don't seem to be able to get time/motivation to do this refactoring...

MrRJ45:
What I wanted to do is get at least the last stats to be persistent, from there it will at least quickly get back to where it was if the service crashed.

As I am new to ruby I also would like other people to test it, as I have had no problems myself, but that doesnt mean it will work as expected :)

If this works fine then I am happy to work slowly on refactoring other bits so they also are persisted, such as the job queue, etc.

MrRJ45:
I've re-read Node::Stats and realised I can store all of the stats in the db, so I have rewritten my changes to do that.

Now all stats are stored in db, and last fetched from here.

Should we close the pull request and I can update on the issue about persistance already open as I change each bit to use a db? :)

PR #187, opened 2015-09-14
Add Arista and Cisco Nexus syslog detection

cfiske:


ytti:
I guess this alone won't do much, as you need to handle them also. Although perhaps with luck it's IOS compatible output, and you just need to add them to the conditions.

Regardless, even like this, it does not hurt so I'll accept it.

PR #190, opened 2015-09-23
Fix Jobs Problem

MrRJ45:
Slight change to fix Jobs Problem I reported in Issue #177

### DIFF
--- a/lib/oxidized/worker.rb
+++ b/lib/oxidized/worker.rb
@@ -16,12 +16,11 @@ module Oxidized
       while @jobs.size < @jobs.want
         Log.debug "Jobs #{@jobs.size}, Want: #{@jobs.want}"
         # ask for next node in queue non destructive way
-        nextnode = @nodes.first
-        unless nextnode.last.nil?
-          break if nextnode.last.end + CFG.interval > Time.now.utc
+        node = @nodes.get
+        unless node.last.nil?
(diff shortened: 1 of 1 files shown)


MrRJ45:
Just to confirm, I've seen the same issue on a clean install of 0.8.1 with none of the SQL code in place.

ytti:
I don't think this works correctly.

When you call #get you rotate the queue. If the interval has not passed, you're next time looking at the next box, which probably has newer timestamp.

We should always check if the oldest box needs to be fetched or not, if not, then we shouldn't rotate, in order to keep oldest on 'top'

MrRJ45:
Reading nodes.rb it looks like returns the top node and moves it to the end of the array?

For some reason its not trying all devices for us and the only way I've found to fix this is to change the way the jobs are picked.

ytti:
The point with #first is that it's non-destructive, it does not change the array. So we 'peek' if next node in array, should be  fetched (has interval passed). If interval has not passed, we don't want to rotate the array, as then we'll get even 'fresher' node on top of the array. So we want to keep the 'freshest' on top of the array.

ytti:
Should I close this, or? Because this is not working. I can explain offline if needed.

MrRJ45:
Feel free to close this :)

I think this was only needed with my persistent storage code

PR #193, opened 2015-09-25
only send username when username is set at all

skoef:
Some of my switches don’t expect any username on telnet. This change
allows me to leave the username field effectively empty for a switch.


ytti:
'would it be simpler to say 'if not X and X.lenght > 0', instead of testing explicitly about not being nil?

skoef:
Actually it should be 'if X and X.length > 0'

PR #194, opened 2015-09-25
made IronWare compatible with older Foundry/Brocade switches

skoef:
Oxidized model for Foundry/Brocade switch models that run IronWare but
require carriage return instead of newline on enable command and
require different filters for temperature/uptime lines


ytti:
Are we sure we need another model? It's probably ok sending \r\n on newer switches too? Also, if different filtering is needed, perhaps we should just detect the ironware type/version and use that as condition?

Something like

cmd 'show version' do |cfg|
  @legacy_model = true if cfg.match(/something something/)
  ..
 end

cmd 'something else' do |cfg|
  @legacy_model ? legacy_something_else(cfg) : something_else(cfg)
end

def something_else cfg
  #clean cfg on new models
end

def legacy_something_else cfg
  # clean cfg on legacy models
end

skoef:
Good suggestion! However, I'm not able to test the changes to the non-legacy version of IronWare, as I don't have one laying around here. I will update my PR soon!

ytti:
I don't think this would break the 'non-legacy' version. But hopefully @nopedial @dotwaffle @fhibler @FlorianDoublet or @nickhilliard can test

Issue #196, opened 2015-09-26
poor error handling if debug log is not writable


nickhilliard:

nickhilliard:
The error handling for debug log opening is messed up.  Instead of returning fail up the stack, it should either die or else warn on STDERR that it cannot open the debug log file for writing.  At the moment, it simply returns fail up the stack which is trapped by "Oxidized::Script" as "unable to connect".  This is completely misleading about the root cause of the problem. 

```
# chown rancid /tmp/oxidized.log-*
# ls -la /tmp/oxidized.log-*
-rw-r--r--  1 rancid  wheel   843 26 Sep 13:17 /tmp/oxidized.log-ssh
-rw-r--r--  1 rancid  wheel  1600 26 Sep 13:13 /tmp/oxidized.log-telnet
# su - oxidized
$ oxs switchname 'show version'
unable to connect
/usr/local/lib/ruby/gems/2.1/gems/oxidized-script-0.1.1/lib/oxidized/script/script.rb:108:in `connect': unable to connect (Oxidized::Script::NoConnection)
	from /usr/local/lib/ruby/gems/2.1/gems/oxidized-script-0.1.1/lib/oxidized/script/script.rb:87:in `initialize'
	from /usr/local/lib/ruby/gems/2.1/gems/oxidized-script-0.1.1/lib/oxidized/script/cli.rb:69:in `new'
	from /usr/local/lib/ruby/gems/2.1/gems/oxidized-script-0.1.1/lib/oxidized/script/cli.rb:69:in `connect'
	from /usr/local/lib/ruby/gems/2.1/gems/oxidized-script-0.1.1/lib/oxidized/script/cli.rb:11:in `run'
	from /usr/local/lib/ruby/gems/2.1/gems/oxidized-script-0.1.1/bin/oxs:5:in `<top (required)>'
	from /usr/local/bin/oxs:23:in `load'
	from /usr/local/bin/oxs:23:in `<main>'
$ ^D
# chown oxidized  /tmp/oxidized.log-*
# su - oxidized
$ ls -la /tmp/oxidized.log-*
-rw-r--r--  1 oxidized  wheel   843 Sep 26 13:17 /tmp/oxidized.log-ssh
-rw-r--r--  1 oxidized  wheel  1600 Sep 26 13:13 /tmp/oxidized.log-telnet
$ oxs switchname 'show version'
[...]

PR #206, opened 2015-09-29
use different path for debug logging

skoef:
When I try to run oxidized as a different user, I get permission errors
for oxidized tries to write debug logs (e.g. true-telnet) in current
working directory which in my case is not writable for the
non-privileged user. I think this can be improved by this change.

1) Why would you use CFG.input.debug?.to_s (which in this case will
always be ‘true’) as a prefix at all
2) let’s use the crash directory for the debug output, as we know this
is available and choose a more descriptive name

### DIFF
--- a/lib/oxidized/input/ftp.rb
+++ b/lib/oxidized/input/ftp.rb
@@ -18,7 +18,7 @@ module Oxidized
     def connect node
       @node       = node
       @node.model.cfg['ftp'].each { |cb| instance_exec(&cb) }
-      @log = File.open(CFG.input.debug?.to_s + '-ftp', 'w') if CFG.input.debug?
+      @log = File.open(Oxidized::Config::Crash + "-#{@node.ip}-ftp", 'w') if CFG.input.debug?
       @ftp = Net::FTP.new @node.ip, @node.auth[:username], @node.auth[:password]
       connected?
     end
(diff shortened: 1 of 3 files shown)


ytti:
So CFG is instance of Asetus object. Asetus will always return something for any method. If configuration option does not exist, it'll create it.
Question mark, is not idiomatic use of ruby method naming, it's not returning boolean value. It'll return nil if config exists or config value if it exists. #to_s just guarantees it's string, which may be useless.

Basically I like your change, but it means we no longer can decide the name or location for the file. Maybe that is not a problem.

@nickhilliard has been thinking about this lately, I'd like to hear if he thinks this makes situation better, i think it does.

nickhilliard:
it's better, yes, but it still doesn't solve the problem that there are too many log files.  oxidized shotguns debug log files all over the place and this makes it troublesome to handle via traditional log rotation / analysis mechanisms.  It would be helpful to implement a mechanism to coalesce these log files into a single channel which could be diverted to e.g. syslog / file / stderr, etc.  Also, see #196, which is probably relevant to this patch.

skoef:
Perhaps removing the `#{@node.ip}` from the filename could be a workaround until syslogging or whatever is implemented?

ytti:
I like the @node.ip, I think it's useful.

ytti:
I'll merge this, I'll figure out docs later (I won't).

PR #220, opened 2015-10-20
add support for Dell networking OS, dnos

erefre:
Force10 OS changed model name from FTOS to DNOS.


ytti:
You're basing this pull request on 'feature-hook' and it includes lot of existing commits from the newer version. Would it be  possible to send another pull request where only the changes are seen as commits which you've made? You probably want this in 'master', right? 

I should have deleted the 'feature-hook' branch, as it's been merged.

erefre:
It should have been put in the master branch yes, my bad and i will submit a new pull request.

PR #221, opened 2015-10-20
add support for Dell networking OS, dnos

erefre:
Dell Force10 has changed model name from FTOS to DNOS


ytti:
Could you add README change to this, where it's added to the list? You don't need to close this PR, just update the READM and hit PR again, and it'll add that commit here.

PR #233, opened 2015-11-03
Some minor tweaks for Oxidized

marnovdm:
We're migrating to Oxidized at work (from Rancid) and I've made some minor changes to the code. Perhaps they can be useful for others as well.

- It seemed like Oxidized was hanging when we started it, after some troubleshooting we figured it was resolving DNS names (some of the DNS names were not available in the dev environment where we were testing). This patch ads optional debug output when resolving the DNS names.

- ~~Log crashes on hostname instead of on IP (in my opinion more easy to correlate, we manage >1000 devices)~~

- Changes Nagios plugin to prefer CRITICAL state if any critical hosts were found. It used to prefer PENDING (warning) even though there could be both, in which case CRITICAL should win.

I'm not very familiar with Ruby so please let me know if I could improve anything.

### DIFF
--- a/extra/nagios_check_failing_nodes.rb
+++ b/extra/nagios_check_failing_nodes.rb
@@ -23,12 +23,12 @@ json.each do |node|
   end
 end
 
-if pending
-  puts '[WARN] Pending backup: ' + pending_nodes.join(',')
-  exit 1
-elsif critical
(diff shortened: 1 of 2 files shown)


ytti:
looks good, thanks.

PR #237, opened 2015-11-11
Updated Readme to match model names

danbaugher:
Added a couple missing models. Fixed a typo in one of the names for the os. Restructured the list to separate models and brands. Updated the names to match the model names.


ytti:
Great work!

PR #238, opened 2015-11-12
Added upstart job for ubuntu RVM installs

danbaugher:
Upstart config file for ubuntu with an RVM install version of ruby.


ytti:
Thanks!

PR #241, opened 2015-11-23
added support for Ciena SOAS 6.x device type

jgroom33:


ytti:
Thanks great work!

jgroom33:
Thanks.  What is the process for updating the gem?  (I'm new to ruby)
Should I create a new pull request with a modification to the changelog and gemspec file?

ytti:
We don't update it after every change. If you want build new gem just for you. You can type 'rake build' and gem should appear in gems/ directory, which you can then install with 'gem install XYZ' on same or another machine.

jgroom33:
I finally figured this out.  oxidized-web requires oxidized. so I had to rename the created gem to oxidized I also had to bump the version up in the gemspec file to insure the correct gem was being installed.

Thanks for the guidance

PR #247, opened 2015-12-17
Change method of Detecting Stackable Switches

### LINKED ISSUE #195: Powerconnect configuration pull updates every polling period without configuration changes
Everytime Oxidized pulls running configuration from the Dell PowerConnect switches if will obtain system information believe through show system. This includes System Up Time which changes every polled period i.e.
! System Up Time: 629 days, 03h:58m:34s
1hour later
! System Up Time: 629 days, 04h:58m:44s

Any chance the def clean cfg could be updated to strip that updating field?

MrRJ45:
Changing the way it is determined if this is a stackable switch. The word "Unit" confirms if it is stackable.
Using the model number works, but can be unreliable and extra work adding every switch model in the regex.

Fixes #195

### DIFF
--- a/lib/oxidized/model/powerconnect.rb
+++ b/lib/oxidized/model/powerconnect.rb
@@ -14,6 +14,9 @@ class PowerConnect < Oxidized::Model
   end
 
   cmd 'show version' do |cfg|
+    if (@stackable.nil?)
+      @stackable = true if cfg.match /(U|u)nit\s/
+    end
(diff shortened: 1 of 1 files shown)


ytti:
This seems broken.

You're using skip_blocks as Integer, but it's never compared against.
You're using skip_block as Boolean, but it's never true.

MrRJ45:
Oops, missed part of the code! Will fix now!

PR #248, opened 2015-12-17
Identify Stackable Switches

### LINKED ISSUE #195: Powerconnect configuration pull updates every polling period without configuration changes

MrRJ45:
Change how we identify stackable switches using the word "Unit" to identify this. This is more reliable than identifying the Model number.

Fixed #195

### DIFF
--- a/lib/oxidized/model/powerconnect.rb
+++ b/lib/oxidized/model/powerconnect.rb
@@ -14,6 +14,9 @@ class PowerConnect < Oxidized::Model
   end
 
   cmd 'show version' do |cfg|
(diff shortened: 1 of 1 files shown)


ytti:
This is not how I would have done it, but If It Works(tm).

PR #251, opened 2015-12-27
expose git repo name and commit oid in hook mechanism and env-vars

nickhilliard:
this patch exposes the git repo name and the commit ID of the relevant commit in the hook mechanism.  The information is subsequently presented as environment variables.


nickhilliard:
The second commit on this patch adds a script to tap into the oxidized hook mechanism which pulls out the relevant commit from the oxidized repo.  This can then be emailed to admins for review.

PR #266, opened 2016-01-17
Added support for using basic auth in http source

laf:
This adds support for passing a username / password over to a http source that support basic auth.


laf:
Thanks for the quick merge :)

PR #274, opened 2016-01-21
option to use ssh proxy command

ElvinEfendi:
WIP

### DIFF
--- a/lib/oxidized/input/ssh.rb
+++ b/lib/oxidized/input/ssh.rb
@@ -1,5 +1,6 @@
 module Oxidized
   require 'net/ssh'
+  require 'net/ssh/proxy/command'
   require 'timeout'
   require 'oxidized/input/cli'
   class SSH < Input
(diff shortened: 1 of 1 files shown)


ElvinEfendi:
@ytti this is not ready to be merged. Can you revert it? WIP - work in progress. I should have been more in the description :)

ytti:
What will it break? If proxy is not set, shouldn't do anything? And proxy won't be set?

ElvinEfendi:
I do not expect it to break anything. But I was planning to update README accordingly and write some specs as well.

ElvinEfendi:
It would be great if you require specs for every feature/patch - at least for the new ones.

ytti:
That is good suggestion. The reason why I don't really have test, was that I pretty much immediately thought the design is not good, and was just waiting to become motivated to redesign the core.
I like the model stuff, and easy extensibility. But I don't like how there is so much state in memory run time, and specially how the amount of state scales with amount of devices, instead of amount of devices being collected. So my idea has been to redesign the core to keep state of only running devices, and keep rest of ephemeral data in SQLite. And write tests for those.
But as I've since changed jobs, and current employer is not running Oxidized, I'm finding it hard to motivate to do it.

But I think I'll do as you suggest, and start requiring specs for new features.

PR #286, opened 2016-01-26
Strip CPU Fan Speed

mikebryant:
On a 6000 series `show inventory` contains a line like `M3mk1 CPU Fan Speed       6490 RPMs`
- this should be stripped


ytti:
I would have maybe gone with RPMs? But matter of taste.

PR #296, opened 2016-01-29
Add F5 BIG-IP model

mikebryant:


ytti:
Is BIG-IP the operating system name?

mikebryant:
This is the output of `tmsh show sys version`
```
Sys::Version
Main Package
  Product     BIG-IP
  Version     11.6.0
  Build       5.123.429
  Edition     Engineering Hotfix HF5
  Date        Tue Aug 11 13:51:32 PDT 2015
```

I'm not entirely sure whether you'd call the operating system `BIG-IP` or `TMOS` - https://support.f5.com/kb/en-us/products/big-ip_ltm/manuals/product/tmos_management_guide_10_1/tmos_intro.html#1023662

What do you think?

ytti:
To me TMOS looks like the OS. Why I think naming things as OS is important, is that often vendors keep using same product name when they introduce new OS.

I'd prefer TMOS if that's ok with you.

mikebryant:
Sure, I don't mind. Updated.

ytti:
Thanks a bunch for this, always great to have new models, especially as common kit as this.

mikebryant:
No problem.

Just working through all the devices we have here, it's nice to be able to give back :)

PR #298, opened 2016-01-31
Update fortios.rb

andreas1o:
Fixed this also but had to do a awkard workaround.. as I see it catch 22.

In a vdom environment we have to do 
config global 
then 
config system console 
set output standard 
end 
end

Right now we can't get the 
get system status
to be good since it per default does more.

So I had to just add some commands that will fail on a none vdom enabled router 
but works on a vdom enabled router. Maybe the trick would be to create a class for fortgatevdom ?

I'll try to pull this to the repository... and comments are welcome.

### DIFF
--- a/lib/oxidized/model/fortios.rb
+++ b/lib/oxidized/model/fortios.rb
@@ -38,6 +38,14 @@ class FortiOS < Oxidized::Model
   end
 
   cfg :telnet, :ssh do
+    post_login 'config global'
+    post_login 'config system console'
+    post_login 'set output standard'
+    post_login 'end'
+    post_login 'end'
+    pre_logout 'config system console'
+    pre_logout 'set output more'
+    pre_logout 'end'
     pre_logout "exit\n"
   end
 


ytti:
@aakso comments?

ElvinEfendi:
@andreas1o why do you not set the output to standard by default? This change will be ineffective for the users such as me that uses read only user to ssh to the devices using oxidized. Moreover I think `oxidized` should not alter any configuration - its job is to fetch the configuration and do post processing.

aakso:
@ytti @andreas1o in our organization we have opted to globally disable pager on our FortiOS devices. Not beautiful but better than having Oxidized changing the configuration.

Maybe we could handle the pager in the Model the expect-way if it is enabled?

ytti:
I'm also worried about precedent of changing configs, it means user has to be RW, changes the security posture. And perhaps makes it more risky (what if telnet session breaks in middle of inputting benign config command, could that turn into a malicious one?).

So yeah, either mandate that globally disable pager or implement pager handling in the model.

Sorry @andreas1o I don't think we're gonna accept that. You can always of course internally use your version, even subclass your .config/oxidized/model/X from the distribution one and rewrite minimal parts of the distribution, to reduce your maintenance requirements.

andreas1o:
Understandable,comments I think we will move towards a output standard instead. 
Since I maily  use CLI to config fortigate I wanted the more output as stanard.

ytti:
Implementing 'more' prompt pager handling isn't hard, we do that in some models. So if you want to keep pager on, we will support that.

andreas1o:
Ok if that could be done I'm happy =) Implement more pager support for fortios.

ytti:
Example how to do it: https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/ios.rb#L7

I can't do it myself, as I can't test it, due to no access to fortios devices.

Issue #300, opened 2016-02-01
Ubiquiti EdgeSwitch Model


doogieconsulting:
Hi,

It would be great to have a Ubiquiti EdgeSwitch model defined within Oxidized. I have absolutely zero experience in Ruby (mainly work in SQL / R for BI) and i have several dozen of these scattered around. I have hobbled together a very basic rb file that works in a limited way but a proper definition would be great.

Thanks!

ytti:
Doesn't it run EdgeOS? Isn't this correct: https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/edgeos.rb

If not, I'm happy to walk you through how to create new model yourself or if you prefer not to do it yourself, give me access to unsupported device and I'll create model for it.

ytti:
Do you know if the operating system has some name? EdgeSwitch sounds like name of the product, not name of the operating system?

Frankly it does not matter how limited it is, if it's sufficient to you, the user of the switch, it's good enough to me. If some other user needs more, they're welcome to contribute.

I have no problem accepting this as pull request, I'd like the operating system name figured out and the indents fixed (that might be just the paste) and it's good to go.

doogieconsulting:
It actually is just EdgeSwitch. I also concur it is confusing to have the name of the OS represent the name of the product. It is just a custom linux package. Here is a sample of the config header
![2016-02-01_16-09-43](https://cloud.githubusercontent.com/assets/17012039/12731299/41f02eaa-c8fe-11e5-9a91-9c0f599ebd26.png)

Ubiquiti also is no help, simply listing the firmware as EdgeSwitch firmware.

My paste did not work out so well, sorry about that. i will send through a pull request to you.

ytti:
I'm still not convinced the OS is called EdgeSwitch. Sometimes it's a huge chore to find out OS name.

doogieconsulting:
The only other terminology i have ever found is EdgeMAX EdgeSwitch. The problem with the EdgeMax designation (my original naming) is that the edgeos routers are also EdgeMax products although UBNT calls these EdgeMAX EdgeRouter. Here are the most recent release pages for both products

EdgeMax EdgeSwitch
http://community.ubnt.com/t5/EdgeMAX-Updates-Blog/EdgeMAX-EdgeSwitch-software-release-v1-3-0/ba-p/1426736

EdgeMax EdgeRouter
[...]

PR #301, opened 2016-02-02
EdgeSwitch Model

doogieconsulting:
Add model file to suit Ubiquiti EdgeSwitch. Assumes enable password is disabled via SSH.


doogieconsulting:
#300

PR #302, opened 2016-02-02
Add OpenGear model

mikebryant:
```
~$ cat /etc/version
OpenGear/IM42xx Version 3.16.2 50b5a369 --  Mon Nov 9 12:01:41 EST 2015
```

The physical model is an IM4248 - so I'm using OpenGear as the OS name, for lack of a better choice.

This is the product: http://opengear.com/products/specs/im4200-specs


ytti:
Yeah can't find any name either.

PR #307, opened 2016-02-06
MRV MasterOS: Ignore config saved/generation dates, match the prompt

natm:
- Fixes an issue where the prompt wasn't getting match
- Ignores lines containing "! Configuration saved on" and "! Configuration generated on" timestamps

### DIFF
--- a/lib/oxidized/model/masteros.rb
+++ b/lib/oxidized/model/masteros.rb
@@ -2,35 +2,38 @@ class MasterOS < Oxidized::Model
 
   # MRV MasterOS model #
 
-comment '!' 
+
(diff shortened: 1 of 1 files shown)


ytti:
@kwibbly comments?

The problem @natm is seeing is that prompt appears like this:
```
You are a SUPER user! ^[[?1034h<boxname># 
```

Is this something you've experienced? Looks like ANSI codes and lack of line feed.

natm:
Discussed with @nopedial, will close this and reopen with just the configuration date item. Should tackle the ANSI codes separately.

PR #312, opened 2016-02-13
Sort the config for AirOS

murrant:
Usually the config is sorted, but sometimes modifying the config moves things around.
Using sort makes for nicer diffs.


ytti:
ping @willglynn 

There isn't any line order specific context in the config? I.e. are we sure sorting is safe?

willglynn:
Sounds good to me. Only reason I didn't sort it initially is because it usually stays sorted. No harm in making it explicit.

murrant:
The configs are all key value pairs.  Changing the line order does not affect the meaning.

PR #313, opened 2016-02-15
add http output type

jgroom33:


ytti:
Thanks!

1. why the 'p [:status]' is that debugging? Is that seen anywhere?
2. why no https support?
3. if 'store' returns false, then we don't declare config as being updated, it's minor thing, and you probably can't know in the http store if or not config has been updated, so probably non-issue
4. any chance you could write a test for it? We don't really have much of them, but that's on me. It would be nice to start adding them for new features - https://github.com/ytti/oxidized/tree/master/spec - for you the benefit is, that new version won't break important feature to you.

jgroom33:
1.  'p [:status]' is for debugging.  It can be removed
2.  I don't need https, so I didn't test that.  I left it as a comment in case someone else wants to test that.
3.  I don't know what store is   :)
4.  Writing a test is above my skill level with Ruby

ytti:
```store```is the method you implemented to store the configs. But fair enough, I'll merge.

jgroom33:
Should I have used load instead of store?

ytti:
No no, it's fine. I just mentioned that depending if #store returns true or false value, we determine if config has changed or not. If you return true, config has been changed. But obviously this output model does not know if config has changed or not, so it does not really matter.

PR #316, opened 2016-02-17
Added support for "show virtual-chassis" output on Junos EX and QFX

benbacardi:


ytti:
Funny thing is, that the existing bit of code about model, was just first commit show-case, I fully expected people wanting more chassis/model based information stored. The existing mx960 isn't even relevant to me, just something that popped on my head at that time.

ciroiriarte:
I'm getting:

# error: syntax error, expecting <command>: virtual-chassis

In a qfx10002-36q, with JunOS 20.2R3-S3.6.

Can it be related?

PR #317, opened 2016-02-17
Don't run 'show virtual-chassis' on EX2200s that don't support it

benbacardi:


benbacardi:
Sorry, I only noticed this after you'd already merged the previous PR! You're quick...!

benbacardi:
Actually, thinking about it - this may be version related. Let me do some more digging, I'll close this for now.

PR #321, opened 2016-02-18
Set terinal to disable the 'more' prompt

roedie:
Creating a backup on our fortigate fails because it waits for 'more'. This disables that setting and unsets it afterwards.


roedie:
Heh, noticed this was also done with pull #298 which was rejected. Closing this one as well...

ytti:
We should add pager handling, it's very simple. Mucking with configs may be problematic.

PR #323, opened 2016-02-18
Add support for Ericsson Redback

roedie:

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
    * PowerConnect
    * AOSW
  * Ericsson/Redback
-   * SmartEdge
+   * IPOS (former SEOS)
  * Extreme Networks
    * XOS
  * F5
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
  * DELL
    * PowerConnect
    * AOSW
+ * Ericsson/Redback
+   * SmartEdge
  * Extreme Networks
    * XOS
  * F5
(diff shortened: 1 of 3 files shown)


ytti:
You want all those metadata without comment marks in front of them? The config probably won't be loadable to the device?

I'm ok with it, just wanted to confirm.

roedie:
I've modified it. Now it comments the metadata.

ytti:
Oh can you also add the new model on README.md (for  the other PR too).

Thanks a bunch, will merge after.

roedie:
I've updated the README.md

ytti:
Is the operating system really named redback? Wasn't that the company? I know sometimes we just can't find OS name, so it might be ok.

roedie:
I had to search around a bit since I normally do not do anything with this equipment but this template works with Redback Smart Edge (SE) and the OS is called SEOS. But it also works with Ericsson Smart Services Router (SSR) where the OS is called IPOS.

Not sure what a good name would be in this case.

ytti:
Hah. Your call, I'd say pick SEOS or IPOS, perhaps the term which is used more typically/more modern equiptment etc. I.e. term which you think will last longer.

roedie:
Renamed to IPOS since that's the OS on the new line of products.

yesbox:
Is there a consensus for what should be included in a model script?

"show hardware detail" used in this script will always trigger a change when checked as it includes values like fan RPM and voltage measurements. I would prefer changes only or primarily when the config changes. "show hardware" only lists the installed hardware with serials, which may be more appropriate in this case.

ytti:
I don't think we've reached consensus and probably won't until the community grows further.

Most of the models are very ad-hoc, and I still have stuff in junos.rb and ios.rb I added in initial commit to show case what is possible. Like arbitrary cherry-picking 'lets check fabric but only from MX960' it has no real-life purpose.

However my take here is, let's follow rancid's lead, and add software version, hardware inventory etc as comments. 
But if people disagree and say it should be config only, I'm ok with that too.

Certainly we should normalise all changing details, like RPM etc, so that they don't change between commits, if that happens, then that is clearly a bug.

PR #324, opened 2016-02-18
Add support for Motorola wireless controllers

roedie:
Only tested with Motorola WM3400 but should work with others.


ytti:
Motorola also may have prompt after exiting? The same I said about extreme applies here.

roedie:
I've updated the README.md and made the naming correct.

PR #326, opened 2016-02-18
multiple remotes for groups in GithubRepo hook

danilopopeye:
While was possible to create a repository per group, I couldn't find a way to configure a specific `remote` for each group.
~~So, this was the best way I found to configure it:~~
Thanks to @ElvinEfendi this is the config now:

## with groups and multiple repositories

``` yaml
---
 output:
    default: git
    git:
      user: Oxidized
      email: oxidized@example.com
    repo:
      routers: /var/lib/oxidized/routers.git
      firewalls: /var/lib/oxidized/firewalls.git
hooks:
  push_to_remote:
    remote_repo:
      routers: git@git.intranet:oxidized/routers.git
      firewalls: git@git.intranet:oxidized/firewalls.git
```

### 
``` yaml
---
 output:
    default: git
    git:
      user: Oxidized
      email: oxidized@example.com
    repo: /var/lib/oxidized/repository.git
hooks:
  push_to_remote:
    remote_repo: git@git.intranet:oxidized/repository.git
```

Note also that you can still use the `single_repo` with groups.

### DIFF
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -171,14 +171,12 @@ module Oxidized
     end
 
     def resolve_repo
-      git = Oxidized.config.output.git
-      cfg_repo = git.repo
+      remote_repo = Oxidized.config.output.git.repo
 
-      if group && !git.single_repo?
-        basedir = File.dirname(cfg_repo)
-        File.join(basedir, "#{group}.git")
+      if Oxidized.config.output.git.single_repo? || @group.nil? || remote_repo.is_a?(String)
+        remote_repo
       else
-        cfg_repo
+        remote_repo[@group]
       end
     end
 
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -24,7 +24,7 @@ module Oxidized
       @vars           = opt[:vars]
       @stats          = Stats.new
       @retry          = 0
-      @repo           = Oxidized.config.output.git.repo
+      @repo           = resolve_repo
 
       # model instance needs to access node instance
       @model.node = self
@@ -168,6 +168,18 @@ module Oxidized
         Oxidized.mgr.add_model model or raise ModelNotFound, "#{model} not found for node #{ip}"
       end
       Oxidized.mgr.model[model].new
+    end
+
+    def resolve_repo
+      git = Oxidized.config.output.git
+      cfg_repo = git.repo
+
+      if group && !git.single_repo?
+        basedir = File.dirname(cfg_repo)
+        File.join(basedir, "#{group}.git")
+      else
+        cfg_repo
+      end
     end
 
   end
--- a/README.md
+++ b/README.md
@@ -535,14 +535,15 @@ This hook configures the repository `remote` and _push_ the code when the specif
   * `username`: username for repository auth.
   * `password`: password for repository auth.
 
-When using groups repositories, the remotes should be passed in the `groups` config for each group.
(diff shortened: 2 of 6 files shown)


ytti:
No problem with this PR. But I assume you don't want merge yet because of the WIP tag?

danilopopeye:
yeah, there's still some work I'll like to do. Should push this afternoon :wink:

ElvinEfendi:
`githubrepo` is a hook, I think putting it's configs under `groups` is a mix of concerns. What about following:

```
hooks:
  github_repo_hook:
    type: githubrepo
    events: [post_store]
    # remote_repo: git@gitlab.lab:oxidized/first.git this stays same when we want to have single repo
    remote_repo:
      key1: git@gitlab.lab:oxidized/group1.git
      key2: git@gitlab.lab:oxidized/group2.git
```

This way all you need to do is to add a private method `remote_repo` to `GithubRepo` that checks if `cfg.remote_repo` is string then return it otherwise if it is `hash` return `cfg.remote_repo[node.group]`.

The same logic applies to `resolve_repo` method.

danilopopeye:
@ElvinEfendi that is a great idea! I'll make the changes :clap:

danilopopeye:
@ElvinEfendi I check `single_repo` for the case the user has explicitly chosen to use only one repo even though they also use groups. It was something we considered using.

ElvinEfendi:
> @ElvinEfendi I check single_repo for the case the user has explicitly chosen to use only one repo even though they also use groups. It was something we considered using.

But if it's not a Hash then it is expected to be single repo by convention, no?

danilopopeye:
> But if it's not a Hash then it is expected to be single repo by convention, no?

Not necessarily. If I have groups, the default behaviour its to create more than one repository. But I can force to create only one with the `single_repo` flag. Or maybe I misunderstood something. :confused: 

Any thoughts on this @ytti ?

ElvinEfendi:
I think there's a bit of misunderstanding :) What I was suggestion is, if we want to have single repo then we will have a config something like following:
```
hooks:
  github_repo_hook:
    type: githubrepo
    events: [post_store]
    remote_repo: git@gitlab.lab:oxidized/first.git
```
This has no coupling with groups config. You can still have any number of groups or no group at all.

When we want to have multiple repositories per config, then we will have something like following:
```
hooks:
  github_repo_hook:
    type: githubrepo
    events: [post_store]
    remote_repo:
      key1: git@gitlab.lab:oxidized/group1.git
      key2: git@gitlab.lab:oxidized/group2.git
```
This however assumes that `key1` and `key2` are group names.

danilopopeye:
Yeah, makes sense and is simpler. :+1: A year without coding is finally taking is toll. :disappointed:

lib/oxidized/node.rb
```diff
@@ -170,5 +170,17 @@ def resolve_model opt
       Oxidized.mgr.model[model].new
     end
 
+    def resolve_repo
```
ElvinEfendi: Could you apply the same logic here as you did for `remote_repo`?

ElvinEfendi:
:+1: for the specs!

lib/oxidized/node.rb
danilopopeye: yeah, sure. I forgot about this one!

danilopopeye:
@ytti / @ElvinEfendi anything more you guys think we should change here?

ytti:
I'm fine with it.

ElvinEfendi:
:+1:

PR #327, opened 2016-02-18
Display JUNOS configuration in configuration mode too

natm:


mikebryant:
Can we make this optional? (controlled by a var or something?) Having two forms of the same config seems redundant.

ytti:
Right. I'm big proponent of '|display set' as it's easily greppable unlike normal output. But I do agree that having it same models make little sense. 

Another company who wanted to do this, ended up using 'output types' functionality:
https://github.com/ytti/oxidized#output-types

So you'd copy ios.rb to ~/.config/model/ directory and add there:

cmd 'show config | display set' do |config|
  config.type = 'set'
  config
end


Which would then create another file with this output. 

Problem with this approach is, that you need your own models. The pain of this may be slightly reduced, if you subclass existing model or reopen it. something like

~/.config/model/ios.rb:
```
require 'oxidized/model/ios'
class IOS
  cmd 'show conf | display set' do |poop|
    poop.type = 'set-configs'
    poop
  end
end
```

I've not tested this. So bit of spit-balling. But it would avoid keeping your model up-to-date with distribution model, adding just locally relevant extension.
However like rancid does, I think we should add lot more metadata, inventory, model type etc as comments.

fhibler:
+1 for the possibility of creating (maybe even both?) outputs!

ytti:
There is possibility today,  as seen above, that's all implemented. But you will need your own model file. I don't have easy answer how to satisfy custom requirements without custom model/config.

funzoneq:
I would be happy to settle on just redundant storage. Works for me as a minimal viable product.

ytti:
I don't understand what you mean by redundant storage. Do you mean both in same file?

If so, then just remove the config.type/poop.type from my examples, then it'll go to same file, if you set the type, it'll go to different type.

ElvinEfendi:
I think we should keep the models in their simplest possible version. And do all the customizations in `~/.config/model/` as @ytti suggested.

funzoneq:
I can confirm that this works:

file: ~/.config/oxidized/model/junos.rb
```
require 'oxidized/model/junos'

class JunOS
  cmd 'show conf | display set'
end
```

ytti:
Thanks for confirming. 

I'm still very confused who would want them in same file, considering that file cannot then be used to restore config (without small grep), and that file cannot easily be ingested by most tooling build (at least all my ghetto parsers would be confused and would need small change to ingest it).
So I'd personally definitely use ``foox.type = 'set'```or comparable, so that I can collect it on separate file, dir or repo (depending on other config).
But of course it might be just my lack of imagination on seeing when they might be good on same file, and of course we will fully support this that this option exits.

ytti:
Closing this. The Recommended way to do this, is to superclass existing model.

cat > ~/.config/oxidized/mode/junos.rb
require "oxidized/model/junos.rb"
class JunOS
  cmd 'show command | display set' do |cfg|
    cfg.type = 'set'
    cfg
  end
end
^D

longzhang-barchart:
Hi @ytti , I'm trying to do this with the recommended way, that is, to superclass existing model. But how do I get the 'set' result to output to a file (I'm using 'File' as the 'output'). I am able to use the below to send 'set' output to the same file. Thank you.

> require 'oxidized/model/junos'
> 
> class JunOS
>   cmd 'show conf | display set'
> end

PR #328, opened 2016-02-19
Add support to specify cipher and kex for ssh connections

### LINKED ISSUE #319: SSH problem to certain devices
While ssh-ing to some network devices I get the error:

Net::SSH::Disconnect with msg "disconnected: Protocol error: expected packet type 30, got 34 (2)"

It seems to be related to the handshake packet size being too long(?). We've fixed the problem by adding the following to line 35 of lib/oxidized/input/ssh.rb:

:encryption => ["blowfish-cbc", "3des-cbc"],
:kex => "diffie-hellman-group1-sha1")

This doesn't look like the best solution to me, but I'm not sure what is. Do you have a solution?

roedie:
Some network equipment exists that does not want to play nice with ssh.
This adds the vars ssh_encryption and ssh_kex which can be used to tell oxidized to use certain encryption and key exchange parameters so it can connect to these devices.

Fixes #319.

### DIFF
--- a/lib/oxidized/input/ssh.rb
+++ b/lib/oxidized/input/ssh.rb
@@ -26,12 +26,18 @@ module Oxidized
       if proxy_host = vars(:proxy)
         proxy =  Net::SSH::Proxy::Command.new("ssh #{proxy_host} nc %h %p")
       end
-      @ssh = Net::SSH.start(@node.ip, @node.auth[:username], :port => port.to_i,
-                            :password => @node.auth[:password], :timeout => Oxidized.config.timeout,
-                            :paranoid => secure,
-                            :auth_methods => %w(none publickey password keyboard-interactive),
-                            :number_of_password_prompts => 0,
-                            :proxy => proxy)
+      ssh_opts = {
+        :port => port.to_i,
+        :password => @node.auth[:password], :timeout => Oxidized.config.timeout,
+        :paranoid => secure,
+        :auth_methods => %w(none publickey password keyboard-interactive),
+        :number_of_password_prompts => 0,
+        :proxy => proxy
+      }
+      ssh_opts[:encryption] = vars(:ssh_encryption) if vars(:ssh_encryption)
+      ssh_opts[:kex] = vars(:ssh_kex) if vars(:ssh_kex)
+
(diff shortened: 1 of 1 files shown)


ytti:
Thanks!

ytti:
Actually come to think of it, I'm not sure we can make this an array. Because we don't have way to specify in node sources what type the variables all node sources (csv, sql, http) will just get strings.

So perhaps this should also expect vars(:ssh_kex) and vars(:ssh_encryption) to be a string, not array? So perhaps

config:
```
vars:
  ssh_kex: "kex1, kex2, kex3"
  ssh_encryption: "encryption1, encryption"
```

ssh.rb
```
ssh_opts[:kex] = vars(:ssh_kex).split(/,\s*/) if vars(:ssh_kex)
ssh_opts[:encyption] = vars(:ssh_encryption).split(/,\s*/) if vars(:ssh_encryption)
```

Then it should be possible to also node's source to specify it.

nebriv:
Hi Roedle,

Can you share a sanitized version of your config? I can't seem to get this to work with:
```
vars:
  remove_secret: true
  ssh_kex: "diffie-hellman-group1-sha1, diffie-hellman-group14-sha1"
```

Thanks!

roedie:
Hi @nebriv,

The way I use it is the following:

config:
source:
  default: csv
  csv:
    file: "/home/oxidized/.config/oxidized/router.db"
    delimiter: !ruby/regexp /:/
    map:
      name: 0
      model: 1
      group: 2
      input: 3
      username: 4
      password: 5
    vars_map:
      enable: 6
      ssh_proxy: 7
      ssh_encryption: 8
      ssh_kex: 9
      ssh_port: 10

routerdb:
some-redback-system:ipos:some-group:ssh:some-username:some-password:::blowfish-cbc,3des-cbc:diffie-hellman-group1-sha1

One thing is, if I use the ssh_kex stuff I also need to set all other options like protocol, username and password even if they are the same as the defaults. It didn't bother me enough to look at this yet.

ytti:
One thing might be that they are "" instead of false or nil. "" is true value which we will then pass and try to use. Unsure if you can actually communicate false/nil through this, worth a shot.

PR #330, opened 2016-02-19
Adds support for Mellanox Switch devices

Ham5ter:
Tested with SX60xx but it should work with other "MELLANOX-OS" Devices to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
  * Juniper
    * JunOS
    * ScreenOS (Netscreen)
+ * Mellanox
+   * MLNX-OS
  * Mikrotik
    * RouterOS
  * MRV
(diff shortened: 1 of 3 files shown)


ytti:
Can you update README.md as well about supported OS.

Is the OS name 'MLNX' we prefer to use OS name as class/filename. But sometimes no OS name is published, just wanted to confirm.

ytti:
ping

Ham5ter:
The OS Name is MLNX-OS
I will update this soon.

Thank you for the Reminder.

ytti:
Constants In Ruby cannot include '-' character.

You could name it MlnxOs or MLNXOS or MLNX_OS

ytti:
pong

Ham5ter:
Made the requested changes.
Also I wanted to say thank you for this great Project :+1:

ytti:
Many thanks!

PR #331, opened 2016-02-19
config root env var

danilopopeye:
this provides a way to isolate the oxidized without touching the user home directory, useful for init.d scripts and others.

```
OXIDIZED_HOME=/etc/oxidized

$ tree -L 1 /etc/oxidized
/etc/oxidized/
├── config
├── log-router-ssh
├── log-router-telnet
├── pid
├── router.db
└── repository.git
```

There is also a `init.d` script template for CentOS and friends. :smiley:


danilopopeye:
@ytti you think it should have something in the README about this?

ytti:
I think it's fair to mention, in README.md, I can imagine this being general interest and docker folks probably will like it.

PR #332, opened 2016-02-20
Configurable PID file

Exhar:
Allows a user to configure the PID file location without setting ENVs et cetera. Useful for init scripts, for example.

Furthermore, it is best practice to for PID files to be stored in /var/run; and setting the home dir env to /var/run doesn't seem the best solution to comply

### DIFF
--- a/lib/oxidized/config.rb
+++ b/lib/oxidized/config.rb
@@ -30,7 +30,7 @@ module Oxidized
       asetus.default.rest          = '127.0.0.1:8888' # or false to disable
       asetus.default.vars          = {}             # could be 'enable'=>'enablePW'
       asetus.default.groups        = {}             # group level configuration
-      asetus.default.pid           = '/var/run/oxidized.pid' # location of PID file
+      asetus.default.pid           = '~/.config/oxidized/oxidized.pid' # location of PID file
 
       asetus.default.input.default    = 'ssh, telnet'
       asetus.default.input.debug      = false # or String for session log file
--- a/lib/oxidized/config.rb
+++ b/lib/oxidized/config.rb
@@ -30,6 +30,7 @@ module Oxidized
         asetus.default.rest          = '127.0.0.1:8888' # or false to disable
       asetus.default.vars          = {}             # could be 'enable'=>'enablePW'
       asetus.default.groups        = {}             # group level configuration
+      asetus.default.pid           = '/var/run/oxidized.pid' # location of PID file
 
       asetus.default.input.default    = 'ssh, telnet'
       asetus.default.input.debug      = false # or String for session log file
--- a/README.md
+++ b/README.md
@@ -446,6 +446,7 @@ vars:
   enable: S3cr3tx
 groups: {}
 rest: 127.0.0.1:8888
+pid: /var/run/oxidized.pid
 input:
   default: ssh, telnet
   debug: false
--- a/README.md
+++ b/README.md
@@ -506,7 +506,7 @@ vars:
   enable: S3cr3tx
 groups: {}
 rest: 127.0.0.1:8888
-pid: /var/run/oxidized.pid
+pid: ~/.config/oxidized/oxidized.pid
 input:
   default: ssh, telnet
   debug: false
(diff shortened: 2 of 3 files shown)


lib/oxidized/config.rb
```diff
@@ -30,6 +30,7 @@ def self.load(cmd_opts={})
         asetus.default.rest          = '127.0.0.1:8888' # or false to disable
       asetus.default.vars          = {}             # could be 'enable'=>'enablePW'
       asetus.default.groups        = {}             # group level configuration
+      asetus.default.pid           = '/var/run/oxidized.pid' # location of PID file
```
ElvinEfendi: I think default path should be `File.join(Oxidized::Config::Root, 'pid')`. This lets us run `oxidized` with as minimum permissions as possible. It should not be necessary for the user that runs `oxidized` to have write permission to `/var/run/oxidized.pid`

ElvinEfendi:
> Furthermore, it is best practice to for PID files to be stored in /var/run; and setting the home dir env to /var/run doesn't seem the best solution to comply

Could you explain why it's better to store the `pid` file in `/var/run`? Any article or stackoverflow link would also work.

I like the idea of having everything in `oxidized`'s root directory because this makes things simpler particularly in terms of user permission handling.

Exhar:
http://stackoverflow.com/questions/5173636/must-my-pidfile-be-located-in-var-run

Although the PID could be stored in the homedir (which is as easy as changing the default configured location to '~/.config/oxidized/'), I can image situations where multiple users will try to start Oxidized whilst an instance is running. If this user cannot access a 'public' dir to check if the PID file exists, Oxidized will crash instead of the (already build in) message that Oxidized is running.

It'll crash because the REST port is already in use; this will be the case when users are using the same config, for example because /etc/oxidized/config exists

ElvinEfendi:
I agree that for the daemons that is constrained to run only one instance at the given time, `/var/run` is a better place. But `oxidized` is not a system level application. It should be possible to run multiple instances of it in the same machine. As to it being crashed because of the port is already in use, this is a bug imho. Instead of changing `pid` path, we should let users to configure the `port`.

danilopopeye:
I think putting the `pid` file in `/var/run` makes sense when oxidized is running on some os daemon like init, runit, upstart etc. So I guess the default value be in oxidized root is ok. :+1: 

On account of the port problem, makes more sense to let the user configure it. Can you provide another PR for this? Or let me know and I'll hack one next week if I manage to get some time at work. :smile:

roedie:
Having the pidfile and port configurable isn't a bad thing. Oxidized even look in /etc/oxidized/config for a configuration during start. I can see no other reason than running it as a system daemon, and then it makes sense to have the pidfile in /var/run

The default behavior shouldn't be changed, but it should definitely be configurable.

jgroom33:
Would this change resolve the following issue:
When running as a docker container, I map the config volume to the host:
docker run --name "vrp-oxidized" -d -v /etc/oxidized/vrp:/root/.config/oxidized -p 8101:8888/tcp -t vfnz/oxidized:latest

The pid file is created in that directory, so is stored on the host.  When the docker container is stopped it cannot restart until the pid file is deleted on the host.

ytti:
@Exhar can you change it so that by default it's same dir as currently?

Exhar:
@ytti yes, will update the PR tomorrow when I have more time
@jgroom33 could be, not sure if this is really the fix; docker (or oxidized, actually) should clean up the PID upon shutdown..

Exhar:
Updated the request; but am not sure if this went as expected ;)

Not sure if you can pick only my changes or not. If not, Ill try to figure out how to solve this... (sigh :))

ytti:
You probably want File.expand_path as ~/ won't be something we can write or read.

ElvinEfendi:
To keep what we have currently as default you just need to do:
```
asetus.default.pid = File.join(Oxidized::Config::Root, 'pid')
```

ytti:
+1

ytti:
@Exhar any update?

ytti:
@Exhar pong?

danilopopeye:
@ytti this can be closed after #435 was merged, right?

ytti:
Yes, thanks.

Issue #320, opened 2016-02-18
Timing issue with Extreme x440


roedie:
When using telnet to backup an Extreme 440 the poll hangs if the config is not saved. It seems the 'n' to take care of the save prompt is sent to fast.

Using SSH all is fine.

ytti:
The 'n' is send with 'pre_logout' without block, this translates to ```cmd 'n'```, which further means we're waiting for a prompt.

In this instance we just want to puke 'n' and not try to wait for a prompt, because one may not exist, if I understood correctly. 

So it should be something like
```
pre_logout do
  send 'n'
end
```

@tobiasa1

tobiasa1:
Hi,

I can try it out tomorrow when I'm at office. I have many different Extreme to verify on.

ytti:
I lieeeeed. ```pre_logout "exit"``` does not wait for prompt. So original should have worked.

roedie:
I do see when using SSH the 'n' is being sent to the X440, but it is done with 1.5msec or so. So when using telnet the 'Save configuration' prompt isn't there yet, the X440 ignores the 'n' and waits for the user to press the 'n'. At least, that's the theory...

Debug using telnet:
```
D, [2016-02-19T12:15:12.639568 #2205] DEBUG -- : Telnet: disable clipaging @x.x.x.x
D, [2016-02-19T12:15:12.937902 #2205] DEBUG -- : Telnet: show version @x.x.x.x
D, [2016-02-19T12:15:13.111712 #2205] DEBUG -- : Telnet: show diagnostics @x.x.x.x
D, [2016-02-19T12:15:14.137252 #2205] DEBUG -- : Telnet: show licenses @x.x.x.x
D, [2016-02-19T12:15:14.677934 #2205] DEBUG -- : Telnet: show switch @x.x.x.x
D, [2016-02-19T12:15:14.844234 #2205] DEBUG -- : Telnet: show configuration @x.x.x.x
D, [2016-02-19T12:15:22.346915 #2205] DEBUG -- : Telnet: exit @x.x.x.x
W, [2016-02-19T12:15:42.814666 #2205]  WARN -- : x.x.x.x raised Timeout::Error with msg "timed out while waiting for more data"
W, [2016-02-19T12:15:43.126306 #2205]  WARN -- : x.x.x.x status no_connection, retry attempt 1
```
Last lines of log:
```
#
# Module vsm configuration.
#
* SW04.7 # exit
[...]

PR #333, opened 2016-02-21
Fix telnet 'wait for prompt' bug

roedie:
This fixes the telnet 'hang' reported in issue #320

### DIFF
--- a/lib/oxidized/model/ipos.rb
+++ b/lib/oxidized/model/ipos.rb
@@ -27,8 +27,10 @@ class IPOS < Oxidized::Model
 
   cfg :telnet, :ssh do
     post_login 'terminal length 0'
-    pre_logout 'exit'
-    pre_logout 'n'
+    pre_logout do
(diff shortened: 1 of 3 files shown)


roedie:
I'm actually not sure if the models where just a
```
pre_logout 'exit'
```
is done should also be modified. They do work, but in my understanding oxidized should wait for a prompt. The device just closes the connection on exit, so it shouldn't be a problem, only cosmetic.

ytti:
Telnet will have match statement 'nil == line' I'm not sure actually when that will be true, it probably hast to be true on EOF, since the models don't fail.

This has not been my original intention, I've misunderstood Telnet behaviour, my intention was, that when ```nil```is offered as match statement for ```cmd```method, match is not expected. This works for ssh, but not for telnet. 
I would need to make significant changes to Net::Telnet to change this behaviour. I'm ready to do it, but if we can live with it, all the better.

PR #336, opened 2016-02-23
Add support for ssh tunnel

roedie:
Newer versions of ssh (>=5.4) support the -W option which makes the
use of netcat obsolete. The -W flag is useful is you want to hop
through a router/switch since they rarely have netcat support.

To use, the ssh_tunnel var needs to be set:
ie:
vars:
  ssh_encryption: "blowfish-cbc,3des-cbc"
  ssh_kex: "diffie-hellman-group1-sha1"
  ssh_tunnel: yes

Some distro's (like RHEL6) have the -W flag backported to their ssh
implementation.

### DIFF
--- a/lib/oxidized/input/ssh.rb
+++ b/lib/oxidized/input/ssh.rb
@@ -23,8 +23,8 @@ module Oxidized
       secure = Oxidized.config.input.ssh.secure
       @log = File.open(Oxidized::Config::Log + "-#{@node.ip}-ssh", 'w') if Oxidized.config.input.debug?
       port = vars(:ssh_port) || 22
-      if proxy_host = vars(:proxy)
-        proxy =  Net::SSH::Proxy::Command.new("ssh #{proxy_host} nc %h %p")
+      if proxy_host = vars(:ssh_proxy)
+        proxy =  Net::SSH::Proxy::Command.new("ssh #{proxy_host} -W %h:%p")
       end
       ssh_opts = {
         :port => port.to_i,


ytti:
So you want both proxy to be set and ssh_tunnel to be set, for this to work?

Why are you setting proxy_sshio when you're not using it ever?

@ElvinEfendi would ssh -W fix your use case as well? O do we need both?

If we need both, I'd prefer something like this:

```
      if proxy_host = vars(:ssh_proxy_nc)
        proxy =  "ssh #{proxy_host} nc %h %p"
     elsif proxy_host = vars(:ssh_proxy_w)
        proxy = "ssh #{proxy_host} -W %h %p"
     end
    
     proxy = Net::SSH::Proxy::Command.new(proxy) if proxy
````

But it seems to me like there probably isn't use case for ```nc```.

**Comment by roedie** [comment]
Hmmm, just did a small investigation which distro's do not support the -W flag, and these are quite rare actually. I thought there would be more... RHEL/CENTOS 5 is one of them.

Every distro newer than 2010 will probably have ssh 5.4 or newer. So, just replacing netcat might even be a better option.

**Comment by ytti** [comment]
Yeah I've been using -W for long time (as to avoid using Agent forwarding). I should've suggested it for the original commit. But we'll see.

File: lib/oxidized/input/ssh.rb [hunk 1]
```diff
@@ -24,7 +24,11 @@ def connect node
       @log = File.open(Oxidized::Config::Log + "-#{@node.ip}-ssh", 'w') if Oxidized.config.input.debug?
       port = vars(:ssh_port) || 22
       if proxy_host = vars(:proxy)
-        proxy =  Net::SSH::Proxy::Command.new("ssh #{proxy_host} nc %h %p")
+        if proxy_sshio = vars(:ssh_tunnel)
+          proxy =  Net::SSH::Proxy::Command.new("ssh #{proxy_host} -W %h:%p")
```
danilopopeye: I know is some :nail_care: but what do you think of not duplication the whole line?

``` ruby
options = if proxy_sshio = vars(:ssh_tunnel)
            "-W %h:%p"
          else
            "nc %h %p"
          end

proxy =  Net::SSH::Proxy::Command.new("ssh #{proxy_host} #{options}")
```

ElvinEfendi:
> @ElvinEfendi would ssh -W fix your use case as well? O do we need both?

I'm not well aware of `-W` flag, but I'm surely fine having a minimum version constraint on ssh. If you say that it is equivalent to `nc` then let's just use `-W`

roedie:
In that case... Here is the updated PR. The commit/line changed ratio is a bit off ;-)

Do you want a new 'clean' PR?

ytti:
@ElvinEfendi can we change the var name to ssh_proxy?

ElvinEfendi:
> Do you want a new 'clean' PR?

@roedie you can squash the commits into one. Ideally this should always happen before the merge for all PRs.

ElvinEfendi:
> @ElvinEfendi can we change the var name to ssh_proxy?

that would probably make it clearer. :+1:

roedie:
There you go, one commit.

ytti:
merged already mistakenly, you need to also change the spec, as we changed the variable

roedie:
Will create a PR for it.

PR #339, opened 2016-02-25
Update ssh proxy

roedie:


ytti:
Well it's a test :) so knowing if it fails or not, is kinda easy 'rake test' to run the tests.

I fixed it now. Thanks.

PR #345, opened 2016-02-25
Edgeswitch remove commands from the config

murrant:
Last one :)


ytti:
I'm ok with it. Personally I'd divide it into multiple lines for clarity/easier extensibility.

murrant:
Ah, I had it that way for a little bit.

PR #348, opened 2016-02-25
create the groups inside the file directory configured

danilopopeye:
captura_de_tela_022516_043032_pm

---

the `File#dirname` was causing the groups directories to be created alongside the directory specified in the `file.directory` config instead of inside of it.

### DIFF
--- a/lib/oxidized/output/file.rb
+++ b/lib/oxidized/output/file.rb
@@ -19,11 +19,11 @@ class OxidizedFile < Output
   def store node, outputs, opt={}
     file = @cfg.directory
     if opt[:group]
-      file = File.join File.dirname(file), opt[:group]
+      file = File.join file, opt[:group]
     end
     FileUtils.mkdir_p file
     file = File.join file, node
-    open(file, 'w') { |fh| fh.write outputs.to_cfg }
+    File.open(file, 'w') { |fh| fh.write outputs.to_cfg }
     @commitref = file
   end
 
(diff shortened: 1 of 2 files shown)


ytti:
The reason why it was implemented like this was that you can decide name for the 'groupless' nodes and 'grouped' nodes. Groupless nodes is the top-level.

so if we configure ```/foo/configs/main``` groupless go there, but grouped go ../group.

How would it look like with the patch?

danilopopeye:
Oh, I see. This way it will behave something like this:

``` yaml
output:
  file:
    directory: "/foo/config"
```

```
/foo
└── config
    ├── groupless1
    ├── group
    │   └── grouped
    └── groupless2
```

My motivation was for the main directory, in my case `/etc/oxidized` would only contain the `config`, `router.db` and a `configurations` directory with all configs inside, both grouped and groupless.

ytti:
Isn't it cleaner to have own directory for groupless? For your behaviour, can't you configure
'/foo/config/groupless'

Then you'll just have 'config' in main-level, and have groupless and grouped neatly in own directories?

ytti:
Oh I think there is some bug in oxidized-web related  to this behaviour, I recall in some situations the webUI will try to read wrong directory.

danilopopeye:
> Isn't it cleaner to have own directory for groupless?

Yeah, makes totally sense. Having have a little more time to think about it, I can achieve the same result just putting another directory in my example. Something like `/etc/oxidized/configs/default`. No need for this changes to be merged :+1: 

---

> Oh I think there is some bug in oxidized-web related to this behaviour, I recall in some situations the webUI will try to read wrong directory.

Yeah, I'll try to find it and send a pull request to oxidized-web later.

PR #355, opened 2016-03-03
add DataCom support

danilopopeye:
captura_de_tela_030316_031318_pm
http://www.datacom.ind.br/en/products/product?familia-dm3000


ytti:
it seems the kit has ssh support according to PDF. I guess you don't run SSH, but would it be much effort to add ssh support to the model?

PR #361, opened 2016-03-09
Procurve filter control sequences and run `show system information`

### LINKED ISSUE #356: Procurve prompt
The following code in lib/oxidized/model/procurve.rb could be improved.

    # FIXME: this is way too unsafe
    prompt /.*?(\w+# ).*/m

Unfortunately, I could not find a way to force the type of terminal to 'none' in stead of the default 'vt100' for a single session. Setting it to 'none' permanently would work, but that will make managing the switch (directly) very hard. So we're stuck with control characters.

The following line does not filter out all possible control characters, but does work for me on HP Procurve 2510/2610/2900/2910/3800/5400:

    prompt /(\e\[\??\d+(;\d+)*[A-Za-z]?)+([\w.-]+[#>] )(\e\[\??\d+(;\d+)*[A-Za-z]?)+$/

It does not start matching from the beginning of the line, because the output actually includes the previous input (which seems to be split by control characters every 10 characters).

The control characters do not end up in the saved config files, so the best way might be to save the output to a temporary file, and parse that file (or use some shortcut that achieves the same).

Ultra2D:
Fixes #356.

Noticed that `show system-information` works on 2510 and 2610, but you need to use `show system information` on 2900/2910/3800/5400.

The carriage return (`\r`) only appears on 2510 and 2610, and only as the first character of a line -> added `^` to existing `cfg.gsub`.

### DIFF
--- a/lib/oxidized/model/procurve.rb
+++ b/lib/oxidized/model/procurve.rb
@@ -1,32 +1,38 @@
 class Procurve < Oxidized::Model
 
-  # FIXME: this is way too unsafe
-  prompt /.*?(\w+# ).*/m
+  # some models start lines with \r 
+  # previous command is repeated followed by "\eE", which sometimes ends up on last line
+  prompt /^\r?([\w -]+\eE)?([\w-]+# )$/
 
   comment  '! '
 
+  # replace all used vt100 control sequences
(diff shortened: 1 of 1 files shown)


Ultra2D:
While I'm at it, might as well remove some secrets.

PR #385, opened 2016-04-08
Fix for issue 359

orealez:
Hi, here's the other fixes for this issue.


orealez:
Actually, this issue seems to have now picked up both commits, one for issue 359 and one for 382.

dschwarz2011:
I'm still having this same issue with my gtm. Each time oxidized runs, there is a rehash of the encrypted password (in this case snmpv3) and it generates a diff. I can see that the updates that the tmos.rb file in the above pull request has the same configuration as my file as well. 

In the /etc/config, f5 group is tied to the tmos model:

model_map:
  f5: tmos

Any ideas?

PR #389, opened 2016-04-11
Add 'nomore' to NOS model for sh run command

PANZERBARON:
With my version of NOS, I have not 'skip-page-display' like ironware models.


ytti:
@hvt will this be compatible with your NOS gear?

TibshoOT:
Oh, I append a new commit to this PR but its not the same scope >_< my bad.

ytti:
No comment from @hvt I assume this commit safe, I can't test myself.

PR #420, opened 2016-05-13
Update acos.rb

broxio:
Modified pre_logout and change \n to \r

### DIFF
--- a/lib/oxidized/model/acos.rb
+++ b/lib/oxidized/model/acos.rb
@@ -58,14 +58,12 @@ class ACOS < Oxidized::Model
   end
 
   cfg :telnet, :ssh do
+    # preferred way to handle additional passwords
     post_login do
-      if vars :enable
-        send "enable\r\n"
-        send vars(:enable) + "\r\n"
-      else
-        send "enable\r\n"
-        send "\r\n"
-      end
+      pw = vars(:enable)
+      pw ||= ""
+      send "enable\r\n"
+      cmd pw
     end
     post_login 'terminal length 0'
     post_login 'terminal width 0'
--- a/lib/oxidized/model/acos.rb
+++ b/lib/oxidized/model/acos.rb
@@ -58,16 +58,18 @@ class ACOS < Oxidized::Model
   end
 
   cfg :telnet, :ssh do
-    # preferred way to handle additional passwords
-    if vars :enable
-      post_login do
+    post_login do
+      if vars :enable
         send "enable\r\n"
         send vars(:enable) + "\r\n"
+      else
+        send "enable\r\n"
+        send "\r\n"
       end
     end
-    post_login "en\r\n"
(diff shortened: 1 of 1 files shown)


ytti:
Why \r instead of \r\n?

Also you're now enabling twice, by sending ```en``` in post_login. Why enable again?

broxio:
@ytti my bad I had wrong assumption on enable logic , my device doesn't need password to switch to privilege mode so only send en command is sufficient, regard Newline \r\n should be good in all case.

ytti:
How will this work on people who have enable? Will it prompt for PW second time?

Does ACOS _always_ require enable? Regardless if or not enable PW is set? If it always needs enable, we probably want something like this:

```
if vars :enable
  post_login do
    send "enable\r\n"
    cmd vars(:enable)
  end
else
  cmd "enable"
end
```

(cmd expects to see prompt, send does not)

broxio:
yes ACOS requires **enable** command to switch to privileged mode and only prompt password once, when I'm in privileged mode (#) I'm not able to send enable second time , can refer to debug output below 

SSH-Debug-Output

```
$ ssh oxidized_admin@10.0.x.y
Last login: Mon May 16 12:05:38 2016 from 10.0.x.x

ACOS system is ready now.

[type ? for help]

A10-Standby>en
Password:
A10-Standby#en
                ^

A10-Standby#
```
Thank you for your emphasize , I didn't realize what different between send and cmd ,
even no enable PW is set, prompt Password: always pop out. so I slightly modify script from your suggestion.

```
    post_login do
      if vars :enable
        send "enable\r\n"
        send vars(:enable) + "\r\n"
      else
        send "enable\r\n"
        send "\r\n"
      end
    end
```
with above code I do have test and verify on my environment both have and not have enable password , working like expected.

ytti:
if PW prompt _always_ happens, We can do this:

```
post_login do
  pw = vars(:enable)
  pw ||= ""
  send "enable\r\n"
  cmd pw
end
```

```||=``` sets pw, if it's not currently set, that is, if no enable PW, we send empty string. I think we need the ```cmd```, because we need to 'consume' the prompt always. If we don't consume it, we might run into desync, some future command might return immediately, because it saw unconsumed previous prompt. We've had this problem in some models.

broxio:
I had test and verify with your suggestion and  it worked on every single unit in my environment, can't definitely a agree more on with ``` cmd ``` to consume prompt always. 

Thank you , made my day.

ytti:
Do you wan to update your pull request, so we can get the change in?

broxio:
@ytti Thank you for your instruction and PR was updated.

ytti:
If you could, maybe rebase it so that it's single commit, now it's lot of history we don't need.

This is not a big deal, if you don't want to rebase it into single commit, that's fine.

PR #421, opened 2016-05-17
fix the git group path config

danilopopeye:
* make sure to expand the paths of all repositories
 * get the right group repository path


danilopopeye:
@ytti can push a new version to rubygems? :smiley:

PR #422, opened 2016-05-18
Dot's in hostnames are frowned upon, but allowed

funzoneq:
```
AMS01.LI17.RM.SW01(config)# hostname "AMS01.LI17.RM.SW01"
Note: To be compliant with RFC 1123, the hostname must contain only letters,
      numbers and hyphens, and must not start or end with a hyphen.
```

### DIFF
--- a/lib/oxidized/model/procurve.rb
+++ b/lib/oxidized/model/procurve.rb
@@ -2,7 +2,7 @@ class Procurve < Oxidized::Model
 
   # some models start lines with \r 
   # previous command is repeated followed by "\eE", which sometimes ends up on last line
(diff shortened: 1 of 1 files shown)


ytti:
If you keep the - as last in group, then you don't need to escape it.

funzoneq:
fixed @ytti

ytti:
Do you need to escape . in character group? I don't think so.

funzoneq:
@ytti it's now officially a one character pull request 👍

ytti:
With three commits :) does not matter, usually nice to rebase multiple typo-fixes/refactors into single commit for better history. Merging.

Issue #425, opened 2016-05-20
output to both git and file


wiad:
It would be useful for me to be able to store the collected config files both as plain file and in git. This does not seem to be supported, or am i missing something?

The file storage is useful to be able to easily provide the config files as input to other applications and scripts. For example, we have a script for parsing and checking for ACL rules in our switch configs. We could of course remodel this to get the switch configuration via git instead, but it's more convenient to just get it from a filepath. 

We really want the git output too, for version control and backup, so it would be nice if we could have both.

ytti:
It's not supported, but would be quite easy to add. Essentially what we'd need to do to support is, split output by ',\*' into an array, essentially in ```https://github.com/ytti/oxidized/blob/master/lib/oxidized/node.rb#L157``` copy the logic from above ```resolve_input``` which supports multiple inputs.

Then in ```https://github.com/ytti/oxidized/blob/master/lib/oxidized/worker.rb#L42``` do something like
```
node.output.new.each do |output|
   # existing code verbatim
end
```

I think that's all that's needed. But don't have time or motivation to do it or test it, not any time soon either.

In the mean time, if you're not ready to write ruby code, why not just have crontab pull your git repo every  5min, to have plain files?

wiad:
I'll give it a go and add a PR if I end up with something useful. Thanks for the pointers!

davama:
@waid I was able to do exactly what you are trying but using the hooks section. Hope it's useful
https://github.com/ytti/oxidized/issues/396

wiad:
@davama Thanks for the tip!

wiad:
@davama If you have the time it would be very helpful if you could test the patch for multiple outputs (ytti/oxidized#426) that I submitted.

davama:
@wiad yes i was looking to try it since this will probably be a more elegant solution than the hooks method.
will probably have an answer by next week though. It's read-only-friday :-)
ill have more time next week
Thanks
Dave

davama:
I thought i had replied.
Well either way it seems there was a git bug with v0.14.2. Now running v0.14.3.
I made the changes to the node{s}.rb and worker.rb files.
and attempted to use multiple outputs but failed. Last default takes precedence:
```
output:
   default: file
   file:
     directory: /home/config/repository/.
   default: git
   git:
    user: config
    email: config@nwk.jwm2.net
    repo: /home/config/repository/.git
```
The git in this instance takes precedence. If i comment the git default line i only get file as my output but no git. 
[...]

PR #426, opened 2016-05-20
Added support for multiple outputs

wiad:
ytti/oxidized#425

Pretty much implemented as per the pointers provided in the issue, but with some minor modifications to make it work. I also had to modify lib/oxidized/nodes.rb to make api calls work. Since you wouldnt want the api to, for exampel, 'fetch' from multiple outputs I just made it so it uses the first output in the array. So if you use multiple outputs, the first one you state in the 'defaults:' line will be used by the api.

### DIFF
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -155,11 +155,13 @@ module Oxidized
     end
 
     def resolve_output opt
-      output = (opt[:output] or Oxidized.config.output.default)
-      if not Oxidized.mgr.output[output]
-        Oxidized.mgr.add_output output or raise MethodNotFound, "#{output} not found for node #{ip}"
+      outputs = (opt[:output] or Oxidized.config.output.default)
+      outputs.split(/\s*,\s*/).map do |output|
+        if not Oxidized.mgr.output[output]
+          Oxidized.mgr.add_output output or raise MethodNotFound, "#{output} not found for node #{ip}"
+        end
+        Oxidized.mgr.output[output]
       end
-      Oxidized.mgr.output[output]
     end
 
     def resolve_model opt
@@ -182,3 +184,4 @@ module Oxidized
 
   end
 end
+
--- a/lib/oxidized/nodes.rb
+++ b/lib/oxidized/nodes.rb
@@ -58,7 +58,8 @@ module Oxidized
     def fetch node, group
       with_lock do
         i = find_node_index node
-        output = self[i].output.new
+        # Use first output type
+        output = self[i].output[0].new
         raise Oxidized::NotSupported unless output.respond_to? :fetch
         output.fetch node, group
       end
@@ -155,7 +156,8 @@ module Oxidized
     def version node, group
       with_lock do
         i = find_node_index node
-        output = self[i].output.new
+        # Use first output type
+        output = self[i].output[0].new
         raise Oxidized::NotSupported unless output.respond_to? :fetch
         output.version node, group
       end
@@ -164,7 +166,8 @@ module Oxidized
     def get_version node, group, oid
       with_lock do
         i = find_node_index node
-        output = self[i].output.new
+        # Use first output type
+        output = self[i].output[0].new
         raise Oxidized::NotSupported unless output.respond_to? :fetch
         output.get_version node, group, oid
       end
@@ -173,7 +176,8 @@ module Oxidized
     def get_diff node, group, oid1, oid2
       with_lock do
         i = find_node_index node
-        output = self[i].output.new
+        # Use first output type
+        output = self[i].output[0].new
         raise Oxidized::NotSupported unless output.respond_to? :fetch
         output.get_diff node, group, oid1, oid2
       end
@@ -181,3 +185,4 @@ module Oxidized
 
   end
 end
+
--- a/lib/oxidized/worker.rb
+++ b/lib/oxidized/worker.rb
@@ -39,13 +39,16 @@ module Oxidized
         msg = "update #{node.name}"
         msg += " from #{node.from}" if node.from
         msg += " with message '#{node.msg}'" if node.msg
-        output = node.output.new
-        if output.store node.name, job.config,
+
+        node.output.each do |output| 
+          output = output.new
+          if output.store node.name, job.config,
                               :msg => msg, :user => node.user, :group => node.group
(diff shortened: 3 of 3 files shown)


ytti:
Thanks, looks good, I want someone else to ACK it too. I recall there are some talk about this feature in issues earlier, so you might want to ping them, to get review.

ytti:
I think I want @ElvinEfendi to chime in too. 

I think it might be good. But to be honest I don't understand git+file output, but maybe git+http or file+http are reasonable combinations.

And two people separately have wanted this. I'm leaning towards merge, just want another opinion.

ElvinEfendi:
I'd really like to keep the core as simple as possible. This sounds like a great feature to be implemented as a hook. That being said I'll review the code in case you want to merge.

@ytti

lib/oxidized/node.rb
```diff
@@ -155,11 +155,13 @@ def resolve_input opt
     end
 
     def resolve_output opt
-      output = (opt[:output] or Oxidized.config.output.default)
-      if not Oxidized.mgr.output[output]
-        Oxidized.mgr.add_output output or raise MethodNotFound, "#{output} not found for node #{ip}"
+      outputs = (opt[:output] or Oxidized.config.output.default)
```
ElvinEfendi: Would be nice to have consistent naming. You've renamed `output` to `outputs` in some places but left it as it is in another places.

lib/oxidized/node.rb
```diff
@@ -182,3 +184,4 @@ def resolve_repo
 
   end
 end
+
```
ElvinEfendi: this is unnecessary

ytti:
If this should be done via hooks, then maybe also the primary output should be hook instead, as to avoid special case.

lib/oxidized/node.rb
ElvinEfendi: by using `||` instead of `or` you can get rid of `()`.

lib/oxidized/node.rb
```diff
@@ -155,11 +155,13 @@ def resolve_input opt
     end
 
     def resolve_output opt
-      output = (opt[:output] or Oxidized.config.output.default)
-      if not Oxidized.mgr.output[output]
-        Oxidized.mgr.add_output output or raise MethodNotFound, "#{output} not found for node #{ip}"
+      outputs = (opt[:output] or Oxidized.config.output.default)
+      outputs.split(/\s*,\s*/).map do |output|
```
ElvinEfendi: Needs to be tested.

ElvinEfendi:
some nitpicks but the code looks good otherwise

ElvinEfendi:
> If this should be done via hooks, then maybe also the primary output should be hook instead, as to avoid special case.

Having more consistent way of configuring outputs would probably better. However my assumption! is that the case where a user wants multiple outputs is already a special case, so why not use a special hook.

danilopopeye:
haven't tested myself, but the code looks :+1:

lib/oxidized/node.rb
wiad: Re: consistent naming: well, this part is really copy/paste from 'resolve_input'. So if I change this it will not be consistent with the rest of the code. And I wanted to keep the changes to a minimum.

lib/oxidized/node.rb
danilopopeye: :+1:

lib/oxidized/nodes.rb
```diff
@@ -58,7 +58,8 @@ def show node
     def fetch node, group
       with_lock do
         i = find_node_index node
-        output = self[i].output.new
+        # Use first output type
+        output = self[i].output[0].new
```
danilopopeye: shouldn't this get the _default output_ here instead of just the first?

wiad: Not sure what you mean - you specify the outputs you want as a comma-separated list, the same way as with inputs:
```
output:
  default: git, file
```
So which is the default output? My simple solution was to use the first one for fetching.

danilopopeye: hum, now I _really_ read the other issue and everything made sense, hahahaha!

idk, I didn't like the string splitting when YAML already have arrays:

``` yml
output:
  defaults:
    - git
    - file
```

If I configure the `file` first, then what happens with the _versions_ button in the `oxidized-web` ? :thought_balloon:

ytti: If we do that, then we need to check if defaults is string or array, which seems dubious practice. Also I think users will find it more confusing than comma separated.

danilopopeye: > If we do that, then we need to check if defaults is string or array, which seems dubious practice.

i know both travis and [gitlab-ci](https://gitlab.com/gitlab-org/gitlab-ce/blob/1339fda1cd3325c0186b5f1b53444e7319ad3cb6/lib/ci/gitlab_ci_yaml_processor.rb#L103-109) use it.

I like it more to use YAML features :neckbeard:, but since its already used in other places, maybe its the best approach here.

ytti: ACK. We'd need to leave such change to version 1 to change it all around and not accept string. But I think our average user actually prefers the string, and more competent people can adapt to what ever.

rgarrigue:
Hi 

Will this be merged someday ?

Best regards,

ytti:
probably not, there is merge conflict and likely would be refactored anyhow. I'm not gonna work on it until core rewrite, which may or may not come.

davama:
> I'm not gonna work on it until core rewrite, which may or may not come.

I too am hoping for core rewrite for another request... https://github.com/ytti/oxidized/pull/654#issuecomment-271000463

hopefully it does :D

laf:
This seems to be a popular requested feature.

I realise we have different opinions on how this should be approached but would it not make sense to merge this as is (once the conflicts are fixed) and work to a 1.0 release where hooks deal with the outputs?

@wiad would you be able to rebase this?

wiad:
I'm havent upgraded oxidized since I made this patch since I didnt want to break its functionality. So I'm not at all familiar with the current code base plus I'm pretty swamped right now so a new patch could take a while. So if anyone else want to fix this they are more than welcome.

mortzu:
any news on this?

PR #436, opened 2016-05-25
Add support for watchguard (FirewareOS)

Alexandre-io:


ytti:
Thanks!

Is there no better name for the OS than 'watchguard', we try to shy away naming the models by the vendor name, because then it's confusing when vendor has products with different OS.

Alexandre-io:
Renamed into FirewareOS :)

ytti:
Many thanks!

Issue #433, opened 2016-05-25
disable interval?


wiad:
Is it possible to disable interval completely and only rely on manual fetching via rest api?

ytti:
No, but that sounds like something that would be useful. I would definitely accept PR implementing that.

ytti:
Fixed by 7da4179dd3e9

DonovanEsterhuizen:
How do i disable the interval option in the configuration file, currently I have the following configured:

interval: 3600

what do I need to do to disable it?

wiad:
Should be by setting:
`interval: 0`

jseifeddine:
currently `interval: 0` disables the interval but it still runs a fetch every time oxidized is started. 

This is somewhat annoying issue, 

In our scenario, fortios devices rekey secrets each and every time you run `show running` which means oxidized will always think there is device changes
for this reason, we have set `interval: 0` and rely on watching syslog events for device config to blame the user who made changes

this works well, however, sometimes we need to restart the oxidized server for whatever reason and in that event oxidized will fetch and commit new backups with only the secrets having been changed


It would be nice if the `interval: 0` would disable automatic fetching entirely - or if there was another config option that implicitly sets no automatic fetching and only if requested through `next` api 


thanks

PR #448, opened 2016-05-27
Added feature to disable polling based on interval

wiad:
Set 'interval' to '0' to disable polling based on interval. Useful if you only want to poll based on external api requests. 

Solves ytti/oxidized#433


ytti:
Thanks. I think this is good. However we have another case where we want to retain ```last```value after doing ```#next``` if we do retain it, then this won't work. But I guess we'll cross that bridge when we get there.

I think this is good feature and will merge it..

PR #458, opened 2016-06-02
Debug statements

ElvinEfendi:
The PR adds detailed debug statements that covers the complete flow from initializing ssh until fetching a config. It also makes sure tests run successfully.

@ytti


ElvinEfendi:
Main motivation to add all those debug statements is that currently I'm seeing a weird bug in my config outputs. I have following model
```
class EOS < Oxidized::Model
  prompt(/^.+[#>]\s?$/)

  comment '! '

  cmd :all do |cfg|
    cfg.each_line.to_a[1..-2].join
  end

  cmd :secret do |cfg|
    cfg.gsub!(/^(snmp-server community)\s(.*)(\sro)/, '\\1 <removed>\\3')
    cfg.gsub!(/(secret 5 )(\S+)/, '\\1<removed>')
    cfg.gsub!(/(password 7 )(\S+)/, '\\1<removed>')
    cfg
  end

  cmd 'show inventory | no-more' do |cfg|
    comment cfg
  end

  cmd 'show running-config' do |cfg|
    cfg
  end

  cfg :ssh do
    if vars :enable
      post_login do
        send "enable\n"
        # Interpret enable: true as meaning we won't be prompted for a password
        unless vars(:enable).is_a? TrueClass
          expect(/[pP]assword:\s?$/)
          send(vars(:enable) + "\n")
        end
        expect(/^.+[#>]\s?$/)
      end
    end
    pre_logout 'exit'
  end
end
```

Intermittently I see the output of `terminal length 0` command instead of the output of `show inventory | no-more`. And the output of later gets prepended to the output of `show running-config`.
Here is an example output:
```
! Pagination disabled.
System information
  Model                    Description                                         
  ------------------------ ----------------------------------------------------
....
According the model there should not be `! Pagination disabled.` and `System information` should be commented.
```

danilopopeye:
:clap: I was missing something like this! :clap:

lib/oxidized/core.rb
```diff
@@ -31,6 +31,7 @@ def initialize args
     private
 
     def run
+      Oxidized.logger.debug "lib/oxidized/core.rb: Starting the worker..."
```
danilopopeye: why include the path of the class in the message?

ytti: path of the file, but yeah, I think it's useful  to people who want to fix something but are not intimately aware with the codebase.

ytti:
looks good thanks.

ytti:
@ElvinEfendi I think your problem is that the model is not consuming prompt, so it gets desynchronised, it's detecting wrong prompt as end for command.

```send``` is not waiting for a prompt.

The model also seems to not send ```terminal length 0``` unless enable is true, that seems odd, right?

What should e fixed:

  1. if enable is set, and enable password is required, send ```password``` with ```cmd```
  2. if enable is set, and enable password is not required, send ```enable``` with ```cmd```
  3. move ```terminal length 0``` outside of enable scope (not related to your problem)

Generally we should use always ```cmd``` if produced output is ```prompt``` terminated, to ensure we're eating the prompt.

Or if EOS supports ```exec```mode we don't have to rely on screenscraping at all.

**Comment by ElvinEfendi** [comment]
> The model also seems to not send terminal length 0 unless enable is true, that seems odd, right?

hrm, I do not see that. `post_login 'terminal length 0'` is outside of `unless` statement.

> I think your problem is that the model is not consuming prompt, so it gets desynchronised, it's detecting wrong prompt as end for command.

In the config I have `enable: true`. So in this case first Oxidized will send `enable` and because of enable is true it will directly go to `expect(/^.+[#>]\s?$/)`. After this it will execute the next post_login command which is `terminal length 0`. As post_login uses `cmd` it will expect the prompt. So I do not understand how would the desynchronization happen.

I understand that I can get rid of several lines in my model and just have something like:
```
  cfg :ssh do
    pre_logout 'exit'
  end
```
Before doing this, I just want to be sure why the issue I described above happening.

@ytti

ytti:
It's not outside the if vars :enable?


-- 
  ++ytti

Issue #466, opened 2016-06-09
syslog.rb + Rest API with URI


ldep30:
Hi,

I've already managed to setup Oxidized to trigger config retrieval via the Syslog extra module, which works fine with a standard configuration :

```
rest: 127.0.0.1:8888
syslogd:
  port: 1514
```

However, I'm now facing issues when using a relative URI in the rest parameter of the Oxidized config file :

```
rest: 127.0.0.1:8888/oxidized
syslogd:
  port: 1514
```
Configuration retrieval is not triggered in this case.

I need to specify the /oxidized path for reverse-proxy and other management constraints.

By looking at [https://github.com/ytti/oxidized/blob/master/extra/rest_client.rb#L5](https://github.com/ytti/oxidized/blob/master/extra/rest_client.rb#L5), it seems that HOST, PORT and URI paths are hard-coded when doing the API call, ignoring the extra URI part specified in the config file.

The Web GUI works fine as well as management functions (config reloading, ...).

Am I missing something ?

Thanks !

ytti:
Yeah  it's not supporting config. You could add support, do something like ```URI.join CFG.rest, 'node/next',opt[:name].to_s```

ldep30:
Thanks for your help.
I'll first try to manually specify the /oxidized path in rest_client.rb, like that :

`@web.put '/oxidized/node/next/' + opt[:name].to_s, data`

ytti:
I'm absolutely confident you're able to make it read oxidized config. I'd happily accept pull request on the change.

ldep30:
I've created the pull request : #467

PR #467, opened 2016-06-10
Update rest_client.rb to support 'rest' parameter

ldep30:
Pull request to fix [https://github.com/ytti/oxidized/issues/466](https://github.com/ytti/oxidized/issues/466)

### DIFF
--- a/extra/rest_client.rb
+++ b/extra/rest_client.rb
@@ -6,7 +6,7 @@ module Oxidized
     require 'asetus'
 
     class Config
-      Root      = File.join ENV['HOME'], '.config', 'oxidized'
+      Root      = Root = ENV['OXIDIZED_HOME'] || File.join(ENV['HOME'], '.config', 'oxidized')
     end
 
     CFGS = Asetus.new :name=>'oxidized', :load=>false, :key_to_s=>true
@@ -21,7 +21,7 @@ module Oxidized
     end
 
     restcfg = CFG.rest
-    if ! restcfg.match(/^http:\/\//)
+    unless restcfg.match(/^http:\/\//)
       restcfg.insert(0, 'http://')
     end
 
--- a/extra/rest_client.rb
+++ b/extra/rest_client.rb
@@ -16,11 +16,9 @@ module Oxidized
       CFGS.load
     rescue => error
       raise InvalidConfig, "Error loading config: #{error.message}"
-    ensure
-      CFG = CFGS.cfg
     end
 
-    restcfg = CFG.rest
+    restcfg = CFGS.cfg.rest
     unless restcfg.match(/^http:\/\//)
       restcfg.insert(0, 'http://')
     end
--- a/extra/rest_client.rb
+++ b/extra/rest_client.rb
@@ -2,8 +2,32 @@ module Oxidized
   class RestClient
     require 'net/http'
     require 'json'
-    HOST = 'localhost'
-    PORT = 8888
+    require 'uri'
+    require 'asetus'
+
+    class Config
+      Root      = File.join ENV['HOME'], '.config', 'oxidized'
+    end
+
+    CFGS = Asetus.new :name=>'oxidized', :load=>false, :key_to_s=>true
+    CFGS.default.rest = '127.0.0.1:8888'
+
+    begin
+      CFGS.load
+    rescue => error
+      raise InvalidConfig, "Error loading config: #{error.message}"
+    ensure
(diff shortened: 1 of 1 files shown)


ytti:
You're creating new config, shouldn't you load the existing config and use what ever url there is in config?

ldep30:
Is there an existing Ruby "Oxidized config" object I could use ?

I've tried to use something like :

`restcfg = Oxidized.config.rest`

but it leads to an error : 

```in `<class:RestClient>': undefined method `config' for Oxidized:Module (NoMethodError)```

Could you point me to the right direction ? Thanks

**Comment by ytti** [comment]
I believe ```CFG``` should be usable. I.e. ```CFG.rest```.

https://github.com/ytti/oxidized/edit/master/extra/syslog.rb#L46

**Comment by ldep30** [comment]
I had already tried that, an error is generated too :

```in `<class:RestClient>': uninitialized constant Oxidized::RestClient::CFGS (NameError)```

This error is triggered when the config code is just reduced to 
```
CFG = CFGS.cfg
restcfg = CFG.rest
```
Same error with just 
```
restcfg = CFG.rest
```

error : ```in `<class:RestClient>': uninitialized constant Oxidized::RestClient::CFG (NameError)```

I don't know Ruby much but I think the object CFGS is not passed to RestClient when called from Syslog.

**Comment by ytti** [comment]
Capitals are const/global, so if it had been loaded before, it should be visible.

So ```CFG.rest```or ```::Oxidized::CFG.rest``` should be visible, provided ```syslog.rb``` is ran first, before it is referred to in rest_client. 
If ```CFG``` is referred in rest_client.rb definition (loaded during ```require ´rest_client'````) then it won't work, as it's not defined yet.

So use the ```CFG``` inside methods and it should work.

Or, maybe even better, just redefine it, like in ```syslog.rb``` so that ```rest_client.rb``` can work correctly, when when ```syslog.rb``` has not been loaded.

File: extra/rest_client.rb [hunk 1]
```diff
@@ -2,8 +2,32 @@ module Oxidized
   class RestClient
     require 'net/http'
     require 'json'
-    HOST = 'localhost'
-    PORT = 8888
+    require 'uri'
+    require 'asetus'
+
+    class Config
+      Root      = File.join ENV['HOME'], '.config', 'oxidized'
```
danilopopeye:
``` ruby
Root = ENV['OXIDIZED_HOME'] || File.join(ENV['HOME'], '.config', 'oxidized')
```

extra/rest_client.rb
```diff
@@ -2,8 +2,32 @@ module Oxidized
...
+    begin
+      CFGS.load
+    rescue => error
+      raise InvalidConfig, "Error loading config: #{error.message}"
+    ensure
+      CFG = CFGS.cfg
+    end
+
+    restcfg = CFG.rest
+    if ! restcfg.match(/^http:\/\//)
```
danilopopeye: `unless` isn't better?

extra/rest_client.rb
```diff
@@ -2,8 +2,32 @@ module Oxidized
...
+    end
+
+    CFGS = Asetus.new :name=>'oxidized', :load=>false, :key_to_s=>true
+    CFGS.default.rest = '127.0.0.1:8888'
+
+    begin
+      CFGS.load
+    rescue => error
+      raise InvalidConfig, "Error loading config: #{error.message}"
+    ensure
```
danilopopeye: I don't get why we need the `ensure` here, @ldep30. care to explain pls?

ldep30:
Hi danilopopeye,

Thanks for the tips. I've added another commit with your 2 suggestions above.

About the "ensure" statement : basically, this bit of code is just a copy/paste of extra/syslog.rb to load the config. I'm a total Ruby beginner so code refactoring is probably necessary :)

I've also tried ytti suggestions about not reloading the whole config and using the CFG object but that was a fail.

Please feel free to submit your modifications here and I'll edit the code.

extra/rest_client.rb
danilopopeye: following [your comment](https://github.com/ytti/oxidized/pull/467#issuecomment-226261488), you don't need this `ensure`.

``` ruby
    begin
      CFGS.load
    rescue => error
      raise InvalidConfig, "Error loading config: #{error.message}"
    end

    restcfg = CFGS.cfg.rest
```

ldep30: Thanks. Changes are committed.

PR #475, opened 2016-06-20
Add support for Coriant 8600 and Coriant 8800.

udhos:


ytti:
two questions

 1. does it really prompt username and password in ssh? I.e. it uses ssh noauth?
 2. there isn't any operating system specific name? We try to name models based on OS, not based on vendor or product family? Perhaps operating system + version?

udhos:
> 1. does it really prompt username and password in ssh? I.e. it uses ssh noauth?

Sorry, I only tested TELNET. SSH part was only a mistake of mine.

> 2. there isn't any operating system specific name? We try to name models based on OS, not based on vendor or product family? Perhaps operating system + version?

8800 calls the OS "TMOS" (Tellabs Multi-service OS)

I could not find what the 8600 family calls its operating system. However it is completely unrelated to 8800 TMOS.

udhos:
I have renamed coriant8800 to corianttmos.

Also I have restricted the login/password prompts for telnet only.

Unfortunately, I could not make SSH work for 8800/TMOS: it hangs forever after:
8840_LAB# enable config terminal length 0

I can't currently test SSH for 8600.

danilopopeye:
@udhos could you add them to the README also?

danilopopeye:
Seems :ok_hand: by me! @ytti ?

ytti:
I don't like the class names, I talked with tellabs engineer and he told even internally they don't have any operating system names, because it's just two different teams in different locations, so no overlap/confusion is possible. And well, it's always the shade of the barn I guess.

Thank you for your work, merged.

Issue #338, opened 2016-02-24
ASA All Contexts via Admin Context Login


jthunt:
Anyone have a good working model to save the admin, system, and all other contexts on a multi context ASA?

Thanks!

etern4ever:
Well..  I was bored and hacked something together.  Not knowing Ruby that well, I really doubt I did this correctly.  It's a bad hack but it works for me.  Someone better at Ruby coding please fix this and add it to the models available in the repo?

Created a new model using the existing ASA and changed the Model Name.
Removed the show running section and replaced it with the following.

` cmd 'changeto system' do |cfg|
   cmd 'show running-config' do |systemcfg|
      allcfg = "\n\n" + systemcfg + "\n\n"
      contexts = systemcfg.scan(/^context (\S+)$/)
      contexts.each do |context|
         cmd "changeto context " + context.join(" ") do
           cmd 'show running-config' do |cfgcontext|
              allcfg = allcfg + "\n\n" + cfgcontext
           end
         end
      end
      cfg = allcfg
   end
   cfg
 end`

yostie:
I didn't notice this issue and posted a request for multi-context ASA support.  Not sure how bad your code is because I don't know Ruby, but the only thing I would suggest is don't do a "show running-config", instead do a "more disk0:/context_name".  This way you get some of the normally starred out sections of the config.  I may take your code and see how well it works for me.

ytti:
I am happo accept pull request. I don't personally have ASAs, and don't have knowledge on them. So it might be prudent to try to ping the authors (i.e. @ytti) in ticket, to see if they want to chime in.

yostie:
The code above by @etern4ever does work.  The only issue is that it dumps all contexts into a single file which is not desirable.  When it goes to obtain the current context config, it would be nice to write it to disk and then move on to the next one to do the same.  Making the changes to go to the system context and do a "more disk0:/context_name" would be trivial.  Unfortunately I don't know Ruby to have it write to disk although I can't imagine that would be difficult to add even with no knowledge of Ruby.  

I may try to find in what .rb file has the code for writing to disk and see what the config object looks like at that point.  If it was presented as a list of config objects, that would be easy to iterate over and write each one to disk.

ytti:
Is #388 duplicate of this?

panzerfly:
From my point of view, yes, #388 is a duplicate.

[...]

PR #479, opened 2016-06-23
Cisco ASA - Added support for multiple contexts

panzerfly:
#338

In order to solve the Cisco ASA issue with multiple contexts I've added some changes to automatically detect if the ASA is running in single or multiple context mode. Based on that if it's running in multiple context mode it will also save the config for each context. Everything will be saved in the same file and contexts configs are preceded by a description line.

I've tested this against 4 Cisco ASA firewalls without problems.

### DIFF
--- a/lib/oxidized/model/asa.rb
+++ b/lib/oxidized/model/asa.rb
@@ -16,6 +16,13 @@ class ASA < Oxidized::Model
     cfg
   end
 
+  # check for multiple contexts
+  cmd 'show mode' do |cfg|
+    if (@context.nil?)
+      @context = true if cfg.include? 'multiple'
+    end
+  end
+
   cmd 'show version' do |cfg|
     # avoid commits due to uptime / ixo-router01 up 2 mins 28 secs / ixo-router01 up 1 days 2 hours
     cfg = cfg.each_line.select { |line| not line.match /(\s+up\s+\d+\s+)|(.*days.*)/ }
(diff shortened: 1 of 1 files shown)


lib/oxidized/model/asa.rb
```diff
@@ -27,27 +34,53 @@ class ASA < Oxidized::Model
...
+            cmd ("more" + profile) do |xml|
+              cfg << (comment xml)
+            end
+        end
+        # if DAP is enabled, also backup dap.xml
+        if cfg.rindex(/dynamic-access-policy-record\s(?!DfltAccessPolicy)/)
+            cfg << (comment "disk0:/dap.xml\n")
+            cmd "more disk0:/dap.xml" do |xml|
+              cfg << (comment xml)
+              puts xml
```
danilopopeye: can you remove this?

lib/oxidized/model/asa.rb
```diff
@@ -27,27 +34,53 @@ class ASA < Oxidized::Model
...
+  post do
+    if (@context == true)
+      # Multiple context mode
+      cmd 'changeto system' do |cfg|
+        cmd 'show running-config' do |systemcfg|
+          allcfg = "\n\n" + systemcfg + "\n\n"
+          contexts = systemcfg.scan(/^context (\S+)$/)
+          files = systemcfg.scan(/config-url (\S+)$/)
+          i = 0
+          contexts.each do |cont|
```
danilopopeye: you can use [#each_with_index](http://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-each_with_index) here:

``` ruby
contexts.each_with_index do |cont, i|
  # ...
end
```

lib/oxidized/model/asa.rb
```diff
@@ -16,6 +16,13 @@ class ASA < Oxidized::Model
     cfg
   end
 
+  # check for multiple contexts
+  cmd 'show mode' do |cfg|
+    if (@context.nil?)
```
danilopopeye: this variable is only defined here, can't we do something more simple?

``` ruby
@context = cfg.include?('multiple')
```

lib/oxidized/model/asa.rb
```diff
@@ -27,27 +34,53 @@ class ASA < Oxidized::Model
...
-	      cfg << (comment xml)
-	    end
-    end
-    # if DAP is enabled, also backup dap.xml
-    if cfg.rindex(/dynamic-access-policy-record\s(?!DfltAccessPolicy)/)
-   	  cfg << (comment "disk0:/dap.xml\n")
-      cmd "more disk0:/dap.xml" do |xml|
-        cfg << (comment xml)
+  post do
+    if (@context == true)
```
danilopopeye: this seems like too much code inside a single method. maybe we could think about refactoring this to something like:

``` ruby
post do
  if @context
    multiple_contexts
  else
    single_context
  end
end

private

def multiple_contexts
  # Multiple context mode
  cmd 'changeto system' do |cfg|
    cmd 'show running-config' do |systemcfg|
    allcfg = "\n\n" + systemcfg + "\n\n"
    contexts = systemcfg.scan(/^context (\S+)$/)
    files = systemcfg.scan(/config-url (\S+)$/)
    # ...
end
```

what do you guys think? /cc @ytti + @panzerfly

ytti: I think refactoring to smaller methods would add clarity and maintainability.

yostie:
My comment doesn't have anything to do with how the code is written but rather how you attempt to pull the information.  If it was me, I would pull all configs from the system context as that is the context you can access the flash from where it stores all the configs.  If you pull all the filenames from URL section of the individual context then do a "more disk0:/<CONTEXT_FILE_NAME" then you will get not just the config but also any hidden passwords such as VPN pre-shared keys.  Doing a "changeto context" and then doing a "sh run" will not produce these keys.  They will be starred out.

Another recommendation would be to put each context's config into its own individual file rather than 1 huge file with all the configs.  I don't know Ruby so if you are already doing some of this above, then my apologies.  A quick glance doesnt look like you are doing either.

panzerfly:
@danilopopeye i've modified the code with your suggestions (thanks for suggested improvements - i'm not a programmer). I've retested the code and works.

@yostie actually the configs are saved directly from disk using "more disk0:/file" :) For putting the context config in different files I think this depends on what output is selected. Personally I use git as output and not files, that's why I put everything in the same config.

danilopopeye:
@panzerfly + @yostie I can't comment on the way to get the files since I don't have such device to test. But the way Oxidized works today you can only have **one** file per device.

oyvellef:
Tested on 4 ASA's with contexts, works fine, and the solution with all configs in one file is good enough for my usage. Thank you for creating this @panzerfly ! 

I also think it would be a better solution to have one file per context + one file for the "system context" as the contexts are separate firewall devices from an operations point of view, but I do not have the skills to provide a suggested change to the code.

arko95:
Any news about pushing it ?

marnovdm:
Tested this on 36 Cisco ASA's at several of our customers, most single context but some multiple context which we couldn't backup before using Oxidized. Can confirm the patch by @panzerfly works well. We now have config backups for the multiple context ASA's and the single context devices still work as well. In my opinion this could be merged into Oxidized.

ytti:
There is still ```puts``` there, I don't think that's intended.

Also we must send the enable password with ```cmd```, so that expect to see prompt.

After those are fixed, I'll merge this, trusting @marnovdm testing.

marnovdm:
@ytti thanks for quick response. I'll test the changes you suggested and will report back tomorrow.

lib/oxidized/model/asa.rb
marnovdm: I've tested the model without the 'puts xml' line and it still works so I agree it doesn't seem required.

Ruben-Bosch:
noticed it is not in 0.20.0, can this be merged

ytti:
There is still the ```send``` and ```puts``` issue in the PR, the PR has to be updated.

marnovdm:
As I still can't get a reply from u/panzerfly I've opened https://github.com/ytti/oxidized/pull/859 rebased on the 0.20.0 codebase. Still testing but I don't see a reason why this wouldn't work.

laf:
@ytti This can be closed now as #859 is merged.

mr-deamon:
@panzerfly can you tell what your files variable contains? for me, the files[i].join(" ") on the following line breaks as there is nothing to join:
https://github.com/panzerfly/oxidized/blame/a6383ea36bcc5d1cf999f5348d4e5058d301f7d5/lib/oxidized/model/asa.rb#L88

PR #480, opened 2016-06-24
Add support for Brocade 6910 (br6910)

cardboardpig:
The Brocade 6910 switch has a significantly different CLI to Ironware/Fastiron and requires different treatment to pull config.

Major differences are:
1. Not possible to disable paging prior to show running-config. Using expect and looking for "--- [Space] Next page, [Enter] Next line, [A] All, Others to exit ---" then sending "a".
2. No show flash command. Closest is "dir" which shows OS bin files plus config filenames.
3. No separate elevated command prompt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
    * Ironware
    * NOS (Network Operating System)
    * Vyatta
+   * 6910
  * Ciena
    * SOAS
  * Cisco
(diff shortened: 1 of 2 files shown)


ytti:
Isn't there better discriminator for the name than model name? Like OS name, OS version? README.md should also be updated to include new boxes.

cardboardpig:
The OS is a once off for this particular model of switch and doesn't have a name like FastIron/Netiron/IronWare. The software image filename is es023004.bix, but any users of the device will know br6910.

Apologies for forgetting the readme update, i'll do that now.

PR #488, opened 2016-07-05
Added support for ssh_no_exec config file variable

nickhilliard:


danilopopeye:
wouldn't make more sense to use this `ssh_no_exec` flag in the methods definition?
The way it's implemented we have to remember to use this in every model we touch.

ytti:
Perhaps. I like to think 'vars' as mostly model/customer arbitrary tags. Original reason was 'enable', it was such a model specific concept, I didn't want to create own configuration item for it.

But this is not really model specific, so perhaps it should be
```
ssh:
  no_exec: true
````

And in https://github.com/ytti/oxidized/blob/master/lib/oxidized/input/ssh.rb#L114
```    
def exec state=nil
  if Oxidized.config.input.ssh.no_exec?
    @exec=nil
  else
   state == nil ? @exec : (@exec=state)
  end
end
```

PR #491, opened 2016-07-08
strip trailing lines in junos model if ssh_no_exec enabled

nickhilliard:
logic was the wrong way around, oops

### DIFF
--- a/lib/oxidized/model/junos.rb
+++ b/lib/oxidized/model/junos.rb
@@ -7,8 +7,7 @@ class JunOS < Oxidized::Model
   end
 
   cmd :all do |cfg|
-    # we don't need screen-scraping in ssh due to exec
-    cfg = cfg.lines.to_a[1..-2].join if telnet || !vars(:ssh_no_exec)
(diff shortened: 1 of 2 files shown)


ytti:
Is it even needed though? I mean 'telnet' shouldn't return true, if we're on ssh, so that line shouldn't be ran on SSH, exec or no exec?

nickhilliard:
it's needed due to ssh_no_exec being set in oxidized-script.

nickhilliard:
the logic is:

- if you're logging in on telnet, the if-expression evaluates to true, so the lines are stripped
- if you're logging in via ssh with exec channels enabled, ssh_no_exec is set to false, so the if-expression evaluates to false, so the lines aren't stripped
- if you're logging in via ssh with exec channels disabled, ssh_no_exec is set to true, so the if-expression evaluates to true, so the lines are stripped

ytti:
Right.  I guess we could also change the 'telnet' to 'screenscrape', and check it there.

Heck, perhaps we should provide 'screenscrape' method in model.rb, which is true if telnet or ssh with ssh_no_exec.
Then it could be just 'if screenscrape'

PR #499, opened 2016-07-15
saos: support ssh and different disable-pager command

cfra:
I can retrieve configurations from a Ciena 3911 running saos-06-05-01-0051 with this when I set `saos_shell_set_more_off`.

Looking around, it seems like the command to disable the pager differs based on version and/or model, so I am not sure whether there is a better solution than to use the variable.

### DIFF
--- a/lib/oxidized/model/saos.rb
+++ b/lib/oxidized/model/saos.rb
@@ -17,7 +17,9 @@ class SAOS < Oxidized::Model
     username /login:/
     password /assword:/
   end
-  cfg :telnet do
+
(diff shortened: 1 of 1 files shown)


ytti:
Do we yet know if there is single command supported by both models?

I'm not sure if the condition is needed. It does not break anything giving two commands and having one of them fail on one of the boxes.

cfra:
No update from original author whether all versions of SAOS support one command.
New version which issues both commands to disable paging and ignores failure of one command.
Works fine for my Ciena 3911.

Issue #497, opened 2016-07-13
Hook Remote git error "wrong argument type nil "


totosh:
I'm trying to setup remote repo git but i get this error :
```
E, [2016-07-13T11:40:43.201248 #12005] ERROR -- : Hook push_to_remote_repo (#<GithubRepo:0x0000000221cda0>) failed (#<TypeError: wrong argument type nil (expected String)>) for event :post_store
D, [2016-07-13T11:40:43.201307 #12005] DEBUG -- : lib/oxidized/worker.rb: Jobs 0, Want: 1
```

`$ oxidized -v` 
0.15.0

I get this error with 2.1.2 and 2.1.8 Ruby version.

Below my config :

```
pid:
"/var/run/oxidized/oxidized.pid"
log: /var/log/oxidized/log
use_syslog: false
username: #####
password: #####
model: cisco
interval: 10800
debug: true
threads: 30
timeout: 20
retries: 3
prompt: !ruby/regexp /^([\w.@-]+[#>]\s?)$/
rest: #####
vars: {}
groups: 
  net-cisco-backup:
    username: #####
    password: #####
input:
  default: ssh
  debug: false
  ssh:
    secure: false
output:
  default: git
  git: 
    user: #####
    email: #####
    repo: "/home/oxidized/git-repos/default.git" 
source:
  default: sql
  sql:
    adapter: sqlite
    database: "/etc/oxidized/oxidized.db"
    table: devices
    map:
      name: address
      model: model
      group: groupname
      username: username
      password: password
      enable: enable
model_map:
  cisco: ios
  juniper: junos
hooks:
  push_to_remote_repo:
    type: githubrepo
    remote_repo:
      net-cisco-backup: "git@gitlab.domain.lan:10022/oxidized/net-cisco-backup.git"
    events: [post_store]
    publickey: /home/oxidized/.ssh/id_rsa.pub
    privatekey: /home/oxidized/.ssh/id_rsa
```

Local repo work fine.

Totosh.

ytti:
Do you have crash file from ```~/.config/oxidized```  I'd like to see the traceback.

totosh:
There is no crash file with this error.
I only see this error in the log file.

ytti:
Is  your node group 'net-cisco-backup'? If not, it won't work. It must match to your group.

[...]

Issue #473, opened 2016-06-17
Hook push to remote : Issue (Unsupported URL protocol)


limace255:
Hi all,

We try to use oxidized, with following concernes :

backup networks equipments
push configs to internal git server (gitlab)
When launching oxidized (in debug mode), we have this in the logs :

```
I, [2016-06-17T11:53:52.466907 #12943] INFO -- : GithubRepo: Pushing local repository(/root/.config/oxidized/oxidized.git/)...
I, [2016-06-17T11:53:52.467765 #12943] INFO -- : GithubRepo: to remote: routers:"git@git_server:PATH/test.git"
E, [2016-06-17T11:53:52.469240 #12943] ERROR -- : Hook push_to_remote (#GithubRepo:0x0000000212f3e8) failed (#) for event :post_store
```

Our configuration file :
``` yaml
hooks:
push_to_remote:
type: githubrepo
events: [post_store]
remote_repo:
routers:"git@git_server:PATH/test.git"
publickey: "/root/.ssh/id_rsa.pub"
privatekey: "/root/.ssh/id_rsa"
```
Did we make something wrong ?

Thanks in advance !

limace255:
Hi all,

We try to use oxidized, with following concernes :+1: 

- backup networks equipments
- push configs to internal git server (gitlab)

When launching oxidized (in debug mode), we have this in the logs :+1: 

I, [2016-06-17T11:53:52.466907 #12943]  INFO -- : GithubRepo: Pushing local repository(/root/.config/oxidized/oxidized.git/)...
I, [2016-06-17T11:53:52.467765 #12943]  INFO -- : GithubRepo: to remote: routers:"git@git_server:PATH/test.git"
E, [2016-06-17T11:53:52.469240 #12943] ERROR -- : Hook push_to_remote (#<GithubRepo:0x0000000212f3e8>) failed (#<Rugged::NetworkError: Unsupported URL protocol>) for event :post_store

Our configuration file :+1: 

hooks:
  push_to_remote:
    type: githubrepo
    events: [post_store]
    remote_repo:
        routers:"git@git_server:PATH/test.git"
    publickey: "/root/.ssh/id_rsa.pub"
    privatekey: "/root/.ssh/id_rsa"

Did we make something wrong ?

Thanks in advance !

laf:
If it helps, here's our config. We use groups but only utilise a single gitlab repo to store them:

```
hooks:
  push_to_remote:
    type: githubrepo
    events: [post_store]
[...]

Issue #334, opened 2016-02-22
OX_REPO_NAME has wrong value


roedie:
During some testing I dumped env to a file using a hook and noticed OX_REPO_NAME was wrong. In my config I configured:
```
repo: "/home/oxidized/git/default.git"
```
for hosts that are not in a group. But, all of my devices are in a group, the default.git repo does not even exist, and I'm pretty sure this device's config is written in /home/oxidized/git/mydevices.git. Still, I get the following output:
```
OX_NODE_FROM=
OX_EVENT=post_store
OX_NODE_MSG=
OX_NODE_GROUP=mygroup
OX_NODE_NAME=mydevice
OX_JOB_STATUS=success
PWD=/
OX_REPO_COMMITREF=b9df0bc7b62023566606fd0024caf17527a7f383
OX_REPO_NAME=/home/oxidized/git/default.git
OX_JOB_TIME=50.121631829
```
I would expect OX_REPO_NAME to have /home/oxidized/git/mydevices.git as value.

danilopopeye:
I am pretty sure this will be fixed in #326

roedie:
Ah, big modification I see. I'll wait and test when it's merged.

danilopopeye:
@roedie did you managed to check if #326 has fixed?

roedie:
I checked it but now the OX_REPO_NAME is empty:
```
OX_NODE_FROM=
OX_EVENT=post_store
OX_NODE_MSG=
OX_NODE_GROUP=just-an-average-group
OX_NODE_NAME=just-a-casual-node
OX_JOB_STATUS=success
PWD=/
OX_REPO_COMMITREF=bbc3e35081b95347d5b8afd486d25136c7f9657d
OX_REPO_NAME=
OX_JOB_TIME=5.708300124
```
I don't have a big issue with it anymore since I don't think I will need that variable anymore, but it still might be a bug.

ytti:
@danilopopeye any thoughts?

danilopopeye:
@ytti I'll try to have a look at this this week. :+1: Never really used the exec hook myself.

danilopopeye:
@ytti can you assign this to me? I keep forgetting to fix it :disappointed:

danilopopeye:
[...]

PR #503, opened 2016-07-21
fix repository path resolution

danilopopeye:
There was missing the option for when using groups and defining the repository as a `String` (like, for `default` group).

This should (hopefully) fix some issues, like #473, #497, #334 :smile:

### DIFF
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -176,8 +176,12 @@ module Oxidized
     def resolve_repo
       remote_repo = Oxidized.config.output.git.repo
 
-      if Oxidized.config.output.git.single_repo? || @group.nil? || remote_repo.is_a?(String)
-        remote_repo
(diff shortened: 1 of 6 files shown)


ytti:
Thanks. Seems like non-trivial amount of work. Bit worried about removal of locking, but can't justify the need for it either.

danilopopeye:
> Bit worried about removal of locking, but can't justify the need for it either.

I haven't removed it, still here :point_right: [lib/oxidized/nodes.rb#L169-L176](https://github.com/danilopopeye/oxidized/blob/de2a9d18a5847439b40adfc0cc4c75e6cbf6262d/lib/oxidized/nodes.rb#L169-L176) :wink:

ytti:
Aah my bad. I don't really know if we actually even need it, I'm not sure. I guess better safe than sorry.

PR #505, opened 2016-07-21
create a `logs` dir for all the debug logs

danilopopeye:
since we have several hundreds of devices, the `oxidized` directory is bloated with log files.
now the logs are kept in a `oxidized/logs` direcotry :smiley:

### DIFF
--- a/lib/oxidized.rb
+++ b/lib/oxidized.rb
@@ -1,3 +1,5 @@
+require 'fileutils'
+
 module Oxidized
   class OxidizedError < StandardError; end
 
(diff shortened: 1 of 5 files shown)


shailigum:
Would it be possible for logs to be to created based on hostname instead of IP?  This would help users who use NAT and have multiple devices behind a single IP.

danilopopeye:
> Would it be possible for logs to be to created based on hostname instead of IP?

could you create a issue so we can discuss it?

ytti:
There may not exist name at all (i.e. it may be IP). So then it would have to be IP:port if uniqueness is needed. But it also could be configurable, which I'm happy to accept PR.

shailigum:
#511

PR #510, opened 2016-07-26
removed current bw stats from screenos 'get system'

mediumo:
removed current bandwith from get system, because it can create different config versions when there is traffic on the interfaces.
current bw is just stats.


ytti:
the alignment isn't correct, granted this isn't python and it does not functionally matter, but lot of non-rubyist read and make models and they might get confused and assume there is some functional reason.

PR #517, opened 2016-08-01
Added support for the DM2104 and DM4100

caiorasec:
I was unable to receive the prompt in any of the versions , but with this change now I'm getting .

### DIFF
--- a/lib/oxidized/model/datacom.rb
+++ b/lib/oxidized/model/datacom.rb
@@ -24,7 +24,12 @@ class DataCom < Oxidized::Model
     cfg.cut_head
   end
 
-  cfg :telnet, :ssh do
+  cfg :ssh do
+    password /^Password:\s$/
+    pre_logout 'exit'
+  end
+
+  cfg :telnet do
     username /login:\s$/
     password /^Password:\s$/
     pre_logout 'exit'


ytti:
Can you do 'ssh -vvvv box' and put output somewhere.

It looks like the box is using SSH for passing login but then will do noauth and fork some password querier, seems odd, so I just want to confirm.

This may also break existing kit, if some other kit does no ssh user auth either, but expects ssh user from application, not from ssh protocol.

@danilopopeye may want to review

caiorasec:
Prior to the change I received the following message:
**raised Oxidized::PromptUndetect with msg "ROUTER# not matching configured prompt (?-mix:^([\w.@-]+[#>]\s?)$)"**

ytti:
And you actually still need the 'password' stanza in the :ssh?

caiorasec:
You're right, I removed the password and also ran perfectly , I had to just keep **pre_logout 'exit'**

caiorasec:
```
  cfg :ssh do
    pre_logout 'exit'
  end

  cfg :telnet do
    username /login:\s$/
    password /^Password:\s$/
    pre_logout 'exit'
  end
```

ytti:
Ok. So now I need @danilopopeye to check if the devices model was already made for works with this.

Because there is two ways to do SSH auth
  1. the right way, using SSH protocol
  2. the wrong way, doing SSH noauth making auth always succeed, then run external program to do auth as payload

We have bunch of devices where same vendor, same operating system, some kit do it the right way, some kit do it the wrong way. And we'll probably need some way to handle both, perhaps:

```
cfg :ssh do
  if vars_ssh_noauth
    username /blaah/
    password /bliih/
  end
  pre_logout 'exit'
end
```

but we'll see

danilopopeye:
@ytti the devices I have here all use telnet, so the changes shouldn't affect them.

PR #519, opened 2016-08-03
Added Nokia.

ghankins:


ghankins:
Going to add some more commands.

PR #520, opened 2016-08-04
Updates for Nokia SR OS.

ghankins:


ytti:
Without the comments, the config isn't loadable, right? I wonder if that's in best interest of the users? @aakso any comments?

ghankins:
Will add the comments again.

PR #523, opened 2016-08-04
Added support for pfSense

stokbaek:
The prompt match is a mess due to colour/text format matching. If there exists a more simple way to match formatted text then please change it to be less messy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
    * Opengear
  * Palo Alto
    * PANOS
+ * pfSense
  * Quanta
    * Quanta / VxWorks 6.6 (1.1.0.8)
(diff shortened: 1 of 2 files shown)


ytti:
Does it support ssh exec mode?

```
cfg :ssh do
  exec true
end
```

This means we're not doing screen-scraping for prompt for ssh.

stokbaek:
pfSense is just a normal BSD machine really, so don't think it supports exec mode.

ytti:
Normal OpenSSH does support exec.

This is exec: ```ssh server "command"``` it does not start shell channel at all, it runs exec channel with command as argument. If SSH implementation supports that, we can use it, which means we never need to do any screen-scraping for prompt.

stokbaek:
pfSense does support this. I will create a new pull request with this instead.

PR #526, opened 2016-08-05
Fix prompt regexp

teunvink:
The prompt regexp shouldn't check for literal parenthesis, they're used to mark a regexp group.

### DIFF
--- a/lib/oxidized/model/edgeswitch.rb
+++ b/lib/oxidized/model/edgeswitch.rb
@@ -4,7 +4,7 @@ class EdgeSwitch < Oxidized::Model
 
   comment '!'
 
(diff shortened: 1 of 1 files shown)


ytti:
So the device does not have literal parenthesis ?

Why are we capturing byte in the prompt?

Why 'tr', is that guaranteed to be there?

teunvink:
Hmm, I have no idea how that 'tr' got there, it shouldn't be there. The parenthesis aren't literal, that's what I wanted to fix. Here's an example of a valid prompt:

```
sw11.bit-2a#
```

So this fix is wrong (due to the 'tr'). I'll submit a new pull request.

PR #536, opened 2016-08-22
A10 and ASA fixes

funzoneq:
- Remove secrets from the A10 config
- Fix the regex for tacacs secret on ASA

- The A10 that I have doesn't have the key 'all-partitions' for show running-config. Maybe I should add two commands?

### DIFF
--- a/lib/oxidized/model/acos.rb
+++ b/lib/oxidized/model/acos.rb
@@ -37,6 +37,14 @@ class ACOS < Oxidized::Model
      cfg
   end
 
+  cmd 'show partition-config all' do |cfg|
+     cfg.gsub! /(Current configuration).*/, '\\1 <removed>'
+     cfg.gsub! /(Configuration last updated at).*/, '\\1 <removed>'
+     cfg.gsub! /(Configuration last saved at).*/, '\\1 <removed>'
+     cfg.gsub! /(Configuration last synchronized at).*/, '\\1 <removed>'
+     cfg
+  end  
+
   cmd 'show running-config all-partitions' do |cfg|
      cfg.gsub! /(Current configuration).*/, '\\1 <removed>'
      cfg.gsub! /(Configuration last updated at).*/, '\\1 <removed>'
--- a/lib/oxidized/model/acos.rb
+++ b/lib/oxidized/model/acos.rb
@@ -6,6 +6,13 @@ class ACOS < Oxidized::Model
   ##ACOS prompt changes depending on the state of the device
   prompt /^([-\w.\/:?\[\]\(\)]+[#>]\s?)$/
 
+  cmd :secret do |cfg|
+    cfg.gsub!(/community read encrypted (\S+)/, 'community read encrypted <hidden>') # snmp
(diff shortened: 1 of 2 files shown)


ytti:
I'm not familiar with this platform, but wouldn't omitting 'all-partitions' remove support for virtual firewalls config?

lib/oxidized/model/acos.rb
```diff
@@ -22,11 +29,12 @@ class ACOS < Oxidized::Model
     comment cfg
   end
 
-  cmd 'show running-config all-partitions' do |cfg|
+  cmd 'show running-config' do |cfg|
      cfg.gsub! /(Current configuration).*/, '\\1 <removed>'
      cfg.gsub! /(Configuration last updated at).*/, '\\1 <removed>'
      cfg.gsub! /(Configuration last saved at).*/, '\\1 <removed>'
      cfg.gsub! /(Configuration last synchronized at).*/, '\\1 <removed>'
+     cfg
```
danilopopeye: you don't need this `cfg` here, since `#gsub!` already return the mutated string, but not harm in keep it 😉

funzoneq:
@ytti That's my feeling as well. But without a partition (which is de stock config), it doesn't show any config at all (because the all-partitions stanza doesn't exist according to the CLI). So either we just do 'show running-config' or we do both 'show runnning-config all-partitions' in addition to 'show running-config'.

danilopopeye:
@funzoneq can you confirm that running both commands in both cases (with and without partitions) nothing breaks?

funzoneq:
Running without partitions:
"""!Current config commit point for partition 0 is 0 & config mode is classical-mode
! COMMAND: show running-config all-partitions
                                                               ^
% Unrecognized command.Invalid input detected at '^' marker."""

ytti:
I guess we should assume people run newer software. So does this mean in newer releases, we won't get config from all partitions?

funzoneq:
@ytti from the manual: "This command is used to view the running-config in the partition where the command is issued. To view the running-config for a different partition, use the show partitionconfig command."

ytti:
And 'show partition-config all' does not show the 'main' partition? Do we need 3 commands? Or should two suffice?

 * show running-config
 * show partition-config all

funzoneq:
@ytti `show partition-config all` does include the config of `show running-config`. So I removed it.

ytti:
Isn't the 'show running-config all-partitions' redundant/useless now too?

funzoneq:
@ytti not if you are running 2.x firmware. So if you want to keep backward compatibility, then we need it.

ytti:
The config you propose works in both firmwares, correctly? I recalled you said in 2.x it won't show the main instance config? So it's not exactly working in 2.x, right?

I'm ok by dropping support for 2.x. I would generally opt into simplicity + modern release, but I'm happy to consider hacks for backward compatibility when there is good justification why running modern release isn't an option.

funzoneq:
@ytti this config will work on both versions.

`show running-config all-partitions` will show all config for 2.x, but an error for 4.x
`show partition-config all` will show all config for 4.x, but an error in 2.x

ytti:
Ok, thanks, merging.

PR #537, opened 2016-08-22
This PR is to fix the cumulus

nertwork:
Right now cumulus prompt regex is broken out of the box, I've put in this fix, as well as added in more cumulus specific 'show commands'


penguin02007:
Cumulus only look at `/etc/network/interfaces` and files in `/etc/network/interfaces.d`. `interface*` is probably not necessary.

https://github.com/ytti/oxidized/commit/393128772849543bcbd71a11ba6d7d356474ab5c#diff-7168cb80842a0d21717293348734a40dL25

Issue #546, opened 2016-08-28
Alvarion support


MajesticFalcon:
Support for getting backups from Alvarion equipment

ytti:
Can you put sanitised version of device online, so that it is reachable via 91.198.120.1. Email authentication details to saku@ytti.fi

MajesticFalcon:
I was hoping I could take care of this one. It will require a new input 'tftp' which uses the same commands as your 'ftp'. I already have a script that backs them up, I am just trying to figure out how to put it into your software

ytti:
Cool, let me know if you need help. Yeah by looking at the zyxel model which uses ftp, I think you can add tftp input method and model which uses it.

PR #548, opened 2016-08-30
Add Alvarion support

MajesticFalcon:
#546

### DIFF
--- a/lib/oxidized/input/tftp.rb
+++ b/lib/oxidized/input/tftp.rb
@@ -2,27 +2,19 @@ module Oxidized
   require 'timeout'
   require 'stringio'
   require_relative 'cli'
-
+  
+  begin
+    require 'net/tftp'
+  rescue LoadError
+    raise OxidizedError, 'net/tftp not found: sudo gem install net-tftp'
+  end
+  
   class TFTP < Input
-    RescueFail = {
-      :debug => [
-        #Net::SSH::Disconnect,
-      ],
-      :warn => [
-        #RuntimeError,
-        #Net::SSH::AuthenticationFailed,
-      ],
-    }
     
     include Input::CLI
     
     # TFTP utilizes UDP, there is not a connection. We simply specify an IP and send/receive data.
     def connect node
-      begin
-        require 'net/tftp'
-      rescue LoadError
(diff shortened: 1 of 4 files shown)


ytti:
you're requiring timeout, but you're not using it

for style can you move the require block for net/tftp after the other requires, not inside connect.


Then we're good to go. You could do git rebase to squas the commits 7 commits into single coherent commit, but that's no biggy, i'll merger regardless once the requires are done. Thanks!

lib/oxidized/input/tftp.rb
```diff
@@ -0,0 +1,61 @@
+module Oxidized
+  require 'timeout'
+  require 'stringio'
+  require_relative 'cli'
+
+  class TFTP < Input
+    RescueFail = {
```
danilopopeye: Do we need this constant if nothing is selected?

lib/oxidized/input/tftp.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    def output
+      ""
+    end
+
+    private
+
+    def disconnect
+      # TFTP uses UDP, there is no connection to close
+    #rescue Errno::ECONNRESET, IOError
```
danilopopeye: Can we remove this?

lib/oxidized/model/alvarion
```diff
@@ -0,0 +1,13 @@
+class Alvarion < Oxidized::Model
+
+  # Used in Alvarion wisp equipment
+
+  # Run this command as an instance of Model so we can access node
+  pre do
```
danilopopeye: wow, this seems super hacky 😝

lib/oxidized/input/tftp.rb
MajesticFalcon: I was not comfortable removing commands that I wasn't sure their intent. I will remove them

danilopopeye: You could also remove the 'ensure' here, better than a empty method

lib/oxidized/model/alvarion
ytti: maybe we should have more descriptive alias for pre or post, which would have same behaviour for now, just use-cases where argument themselves need to see instance context.

PR #558, opened 2016-09-07
Added APC model to oxidized

davromaniak:
Hello.

I'm switching from Rancid to Oxidized, and I wrote a quick APC model, which gets the configuration file through a FTP connection.

Thanks.


ytti:
Great work.

APC does lot of kits, I doubt they universally support this method for config fetch? Try to find some operating system name, we prefer to use operating systems, not vendors.

If you can't find operating system name, then perhaps include product family to the APC name, so next APC model won't problematic.

ytti:
Oh also, please update README.md.

davromaniak:
Hi.

I only use one kind for APC kit (AP7922), maybe I will call this module "apc7922" instead of the too generic "apc", because I only know it works with AP7922 PDUs.

For the README.md, it's on the next commit, sorry for forgetting it.

Thanks.

ytti:
I think we should do bit more research, to know if there is OS name, or at least if there is specific family which all use it. Would be bit confusing to call it apc7922 if lot of other APCs work the same.

I know it's PITA, but it's one off work, and will benefit future users. It might be that information simply is not available, but then I'm inclined to arbitrarily rename the model later when more information does become available, which will mean things will break for you during upgrade.

davromaniak:
Hi.

I agree it's a bit too violent to rename it after only one model of their devices.

The only thing I'm sure now, is the FTP is implemented by the APC Web/SNMP Management Card, so it would work on any APC device with the management card having FTP enabled.

I will ask some colleagues who use others APC PDUs to test the FTP access and tell me if there's the config.ini file in the FTP root folder.

Thanks.

matejv:
I have some APC InRow RC cooling units, that have the same type of management interface. From what I gather, APC calls this APC Operating System or AOS:

http://www.apc.com/us/en/faqs/FA156047/

So I'd suggest naming this module apc-aos, apc_aos or something similar (not sure what characters are allowed in Oxidized module names).

ytti:
APC_AOS or APCAOS seem fine to me.

We can't use minus, because that would be thought as method call ```APC.-(AOS)```

ytti:
Can you rename the class to APC_AOS.

davromaniak:
Hello.

Sorry for not answering in a long time, I was busy on other subjects.

There's also another issue on this plugin, by default, the config file generated contains the timestamp of the generation on a line like "; Configuration file, generated on 09/28/2016 at 14:22:17 by Administrator".

Well, it doesn't need to be kept in the resulting file being commited in oxidized.

I'm a complete newcomer in ruby, but do you know where I have to add some code lines to do something like "grep -v 'Configuration file, generated on'" on the resulting file ?

I will rename the plugin and do the necessary in the README file tomorrow.

Thanks.

ytti:
Pretty much look at any model in model directory. Most of them deal with issue exactly like this. You'll want something like this ```var.gsub! /match/, 'replace'``

davromaniak:
I just commited/push the needed changes.

It's not the most beautiful ruby code ever seen, but it works.

Thanks.

ytti:
There is merge conflict on README.md, can you update README.md on the newest version of README.md. I'll then merge the change, thanks!

davromaniak:
Hello.

I just downloaded the latest version of the README.md file and re-applied my modification.

Is it the way it has to be done ?

Thanks.

davromaniak:
Well now, it's cleaner and clearer as I updated the APC entry using the new way of listing vendors and modules.

ytti:
It still won't merge, you need to solve the conflicts.

You could probably do it with 'rebase', squashing the unnecessary changes in the process.

Or you could just reclone and copy changes on top of the existing files.

davromaniak:
Ok.

I'm deleting my fork and redoing all from scratch.

Thanks.

davromaniak:
I've created a new pull request (#578) from an up to date repository.

I close this one.

Thanks.

PR #588, opened 2016-10-13
Add key authentication to ssh input

denvera:


ytti:
We already look OpenSSH config for keys, so if OpenSSH knows which key to use, it'll already work.

However I am happy to accept explicit key definition, but I don't think it should be node specific, but rather variable vars(:ssh_key) or equivalent.

Thanks!

denvera:
Thanks for the feedback!  I've made a change to use vars(:ssh_keys) instead.

PR #596, opened 2016-10-27
add support for the HPE BladeSystem

flokli:


ytti:
no way to disable pager?

flokli:
Unfortunately I didn't find any by way for it by looking through the help. Probably there are undocumented commands, but I don't know them.
It's not pretty, but everything pager-related gets filtered out :-)

ytti:
fair enough, thanks

PR #600, opened 2016-10-31
add Ruby.gitignore from github.com/github/gitignore

flokli:


ytti:
Pretty sure we had gitignore at one point or another with:
```
Gemfile.lock
gems
````

Now gems should be changed to pkg though. 

I'm not against that, I'm just curious if you have some particular problem you're looking to solve?

flokli:
I usually use bundler to manage ruby dependencies, and got annoyed by the unclean working directory, due to the created `vendor/` and `.bundle` directories. Then I saw there currently is no `.gitignore` at all.
Having one also makes fiddling with the source code much easier, as a `git diff` only contains the real changes :-)

ytti:
Fair enough, merging.

flokli:
Thanks!

PR #601, opened 2016-10-31
add Planet SGS

flokli:
This adds support for Planet's Layer 3 [Managed Ethernet Switches](http://www.planet.com.tw/en/product/product_list.php?id2=504).

Their syntax is pretty similar to the one from Cisco, but there are some hacks needed (like `\r` sometimes being echoed before the prompt, and a non-existing `show inventory`, which is why I decided to use a separate model for them.


ytti:
Please add it to README.md as well

flokli:
Woops, fixed :-)

ytti:
Thanks!

PR #606, opened 2016-11-04
This adds support for D-Link switches

rfdrake:
This specifically supports DES-3526/3550 series switches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
  * DELL
    * [PowerConnect](lib/oxidized/model/powerconnect.rb)
    * [AOSW](lib/oxidized/model/aosw.rb)
+ * D-Link
+   * [D-Link](lib/oxidized/model/dlink.rb)
  * Ericsson/Redback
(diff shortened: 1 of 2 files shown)


ytti:
Dlink really has no OS name?

rfdrake:
They've got several switch models but I'm not sure if the firmware has 
an actual OS name.  The firmware for the switches was labelled like this:

des3550_firmware_501b65.had

Even looking at a few of the newer switches, they don't appear to name 
the OS.

ytti:
But the newer will also work with this model? And big Dlink too? Or just 'DES' Dlinks? If just DES, it might be prudent to add that to the model name, so it does not become confusing when new Dlink switch is added.

However if all are under same OS now, I'm happy to accept this too. It's just that I won't guarantee I won't change names later, if it's named non-uniquely, but that may be better option now, than to invent bad unique name, due to lack of clear commonly understood name (which is bound to happen, when new software emerges from same vendor).

rfdrake:
It should work with the newer models and the big dlink switches. I've 
checked and the CLI is the same across their models.  If anyone tries it 
and it fails I'll try to fix it up so it works.

PR #613, opened 2016-11-10
renamed  alvarion -> alvarion.rb

rfdrake:
I believe from this:
require File.join dir, file+'.rb'

That the only way this model would work is if it has a .rb extension.  I
don't have any of these devices to test if this works.


ytti:
absolutely, thanks.

PR #614, opened 2016-11-10
add support for PLANET SG switches

flokli:
So far, only SGS switches were supported.

Now, we check the model type during the `show version` command, and only
execute the `show transceiver details` command in case an SGS switch was
detected (as its not supported on SG models).
We will also strip lines containing the current System Time and Uptime.
These only appear on SG models, but it's a good idea to strip them anyways.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,8 +140,7 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
    * [Opengear](lib/oxidized/model/opengear.rb)
  * Palo Alto
    * [PANOS](lib/oxidized/model/panos.rb)
- * PLANET
-   * [SGS](lib/oxidized/model/planetsgs.rb)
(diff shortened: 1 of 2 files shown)


ytti:
I would prefer not to rename files, unless rename is to actual unique OS name. But I guess you may be only user of the model right now, so what the heck.

flokli:
Yes, sorry for that. Thought there were more differences on these models from an oxidized point of view, which is why I named them pretty specific.
In case I'll encounter another planet model, I will try to add support to it in `planet.rb`.

Do you plan to do a release? Then I could strip the custom model implementations in `.config/oxidized` ;-)

ytti:
Yeah I guess I can do release, maybe during the weekend.

flokli:
nice, thanks!

PR #616, opened 2016-11-18
Added SG support to ciscosmb.rb model

Atroskelis:
Adapted for SG series, not sure about SF. Tested on SG200 with default config. 

Please test.

### DIFF
--- a/lib/oxidized/model/ciscosmb.rb
+++ b/lib/oxidized/model/ciscosmb.rb
@@ -36,6 +36,8 @@ class CiscoSMB < Oxidized::Model
   cfg :telnet, :ssh do
     username /^Username:/
     password /^\r?Password:$/
+    post_login 'terminal datadump' # Disable pager
+    post_login 'terminal width 0'
     post_login 'terminal len 0'
     pre_logout 'exit' #exit returns to previous priv level, no way to quit from exec(#)
     pre_logout 'exit'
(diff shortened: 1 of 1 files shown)


ytti:
I thnk we need also 'terminal datadump 0', having invalid command won't hurt, but I think the other, older SMB version won't do 'term len 0'.

ytti:
And probably no reason to remove terminal width, right?

Atroskelis:
You are right, it will run commands nonetheless  and the invalid commands wont affect it. I shall modify it accordingly.

Atroskelis:
The 'terminal width' command isnt available for me, hence it's removal. I readded it, just in case 

```
D18-5u#terminal 
  length  Terminal length
  no      negate the command history function
```

ytti:
The whole point of this exercise is, that there are two distinct SMB boxes. Another of them, does not use SSH auth at all, but always allows SSH users in, and then presents CLI auth. Another of requires SSH auth.
And this isn't only difference between them, there is also the pager disabling difference, and pretty sure the width difference.

Unfortunately I don't have access to either of these devices, so I cannot verify, but I trust that the newest modifications make it work for both.

Atroskelis:
For further reference, i tested this on Cisco Sx220 Series Switch Software, Version 1.0.0.17

PR #618, opened 2016-11-19
Added option to disable ssl verification checks for http source

laf:
As always, prefaced by I'm not a ruby guy :)

Tested by someone on gitter channel (well the basics of just adding `http.verify_mode = OpenSSL::SSL::VERIFY_NONE` directly to the http.rb was but I've turned it into a config option and update docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -506,7 +506,7 @@ source:
       X-Auth-Token: 'somerandomstring'
 ```
 
-You can also pass `no_check_certificate = true` if you want to disable ssl certificate verification:
+You can also pass `secure: false` if you want to disable ssl certificate verification:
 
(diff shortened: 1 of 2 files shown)


ytti:
Looks good. But would it be more consistent to do ```insecure: true```, like we do on SSH, when we don't verify keys?

ytti:
Oh and you probably want ```@cfg``` not ```uri```

ytti:
So perhaps ```http.verify_mode = OpenSSL::SSL::VERIFY_NONE unless @cfg.secure```

laf:
You're the boss :)

Just pushed README and code updates as above.

PR #619, opened 2016-11-21
IOSXR Admin Configuration Backup

clay584:
In order for an XR device to be successfully restored, the admin config must also be backed up.  This PR adds that.

I wanted to also leave in the line "!! IOS XR Admin Configuration 5.3.4." and "!! IOS XR Configuration 5.3.4" so you could distinguish the admin config from the running-config, but I suck at ruby.  Using the delete_at method on the array returns the value of the deleted element, not the array.  Can't figure out how to just yank that line using array methods.  Maybe you guys can help me with that.  I was also thinking to using gsub to remove it, but figured the cleaner way would be to remove it in the cmd block as you were already removing the first four lines of the config in this manner.

**My first PR, be gentle. **

### DIFF
--- a/lib/oxidized/model/iosxr.rb
+++ b/lib/oxidized/model/iosxr.rb
@@ -23,8 +23,11 @@ class IOSXR < Oxidized::Model
     comment cfg
   end
 
+  cmd 'admin show running-config' do |cfg|
+    cfg
+  end
+
   cmd 'show running-config' do |cfg|
-    cfg = cfg.each_line.to_a[3..-1].join
     cfg
   end
 
--- a/lib/oxidized/model/iosxr.rb
+++ b/lib/oxidized/model/iosxr.rb
@@ -24,10 +24,12 @@ class IOSXR < Oxidized::Model
   end
 
   cmd 'admin show running-config' do |cfg|
+    cfg = cfg.each_line.to_a[1..-1].join
     cfg
   end
 
   cmd 'show running-config' do |cfg|
+    cfg = cfg.each_line.to_a[1..-1].join
     cfg
   end
 
--- a/lib/oxidized/model/iosxr.rb
+++ b/lib/oxidized/model/iosxr.rb
@@ -29,7 +29,7 @@ class IOSXR < Oxidized::Model
   end
 
   cmd 'show running-config' do |cfg|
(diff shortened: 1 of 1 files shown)


ytti:
What do we need from admin config? Username/password? Maybe some low-level HW config stuff? Problem with this is, the config is no longer directly usable as backup, right? You cannot copy it to fresh box and have parser read it?

It is bit of a pickle how to do this. We do support multiple files per device, like:


```
cmd 'admin show running-config' do |cfg|
   cfg = cfg.each_line.to_a[1..-1].join
   cfg = Oxidized::String.new(cfg)
   cfg.type = "admin"
 end
```

This would create separate file for it.


But does everyone want multiple files for given device?

clay584:
Using the above snippet you shared in /home/oxidized/.config/oxidized/model/iosxr.rb, still does not create a second file; it just appends it to the single config file output.  For my purposes, having it in the same file is ok.

If I put the config.type = "admin at the end of the block, throws an exception of...

`undefined method `type=' for #<String:0x00000002330160> [NoMethodError]`

If I do the following, it appends the admin config to the existing single output file, not to a new file.

```
require 'oxidized/model/iosxr'

class IOSXR

  cmd 'admin show running-config' do |cfg|
    cfg.type = "admin"
    cfg = Oxidized::String.new(cfg)
    cfg = cfg.each_line.to_a[1..-1].join
  end

end
```

ytti:
So ```cmd``` gives Oxidized::String to the block, but ```each_line.to_a.join``` return a string, string does not have ```type``` method.

So in the end you need
```
cfg = Oxidized::String.new(cfg) # turn it back to Oxidized::String
cfg.type = "something" # set the type
cfg # return it from block
```

nickhilliard:
@clay584 64-bit XR changes the admin mode output format by using a tcp connection to a local process and issuing the admin commands over this connection. Here's an example of what happens when you type `admin show running`: http://p.ip.fi/CIKq

On execution, this causes per-run diffs to be generated because the timestamp will change for each run.  You can test this out easily on XRv.

laf:
No further response has been provided on this so I'm closing this PR. Feel free to re-comment @clay584 with responses to the above and we can re-open.

PR #624, opened 2016-12-01
Update file.rb

paulocoimbrati:
When a group is explicitly set, backup files are stored at oxidized base ~/.config/oxidized/GROUP not at  ~/.config/oxidized/configs/GROUP

### DIFF
--- a/lib/oxidized/output/file.rb
+++ b/lib/oxidized/output/file.rb
@@ -19,7 +19,7 @@ class OxidizedFile < Output
   def store node, outputs, opt={}
     file = File.expand_path @cfg.directory
     if opt[:group]
-      file = File.join File.dirname(file), opt[:group]
+      file = File.join file, opt[:group]
     end
     FileUtils.mkdir_p file
     file = File.join file, node


ytti:
This is by design. Consider you have ungrouped and grouped devices, you probably want them all unique directory and at same level, right? One way would be to have two configuration options for this, another way is to ask user to give directory for the ungrouped devices and use that as clue for grouped dirs.

So it's not base, it's ungrouped dir. Hope this clarifies.

paulocoimbrati:
Is there a way to put grouped and ungrouped at same directory?

ytti:
No, I don't think that's possible. Canonical example what people want to do is:

~/.config/oxidized/configs/default

then you'll see ungrouped in configs/default and grouped in configs/group. But you want everything under ~/.config/oxidized/configs/default, without directories? All the files in flat space?

paulocoimbrati:
the above example works for me. I need all files (grouped or not) inside configs. Thanks and great job you are doing here.

PR #628, opened 2016-12-04
New hook: awssns - Publish messages to AWS SNS topics

natm:
This hook can be used to publish messages to an AWS SNS topic.

We are integrating multiple systems with Oxidized by having them listen to SQS queues which are subscribed to SNS topics. When device configuration changes or a backup takes place messages are published to the specified topic.


ytti:
Looks good, thanks.

PR #633, opened 2016-12-07
exclude time from show run output

jbaker48:
Older versions of EOS output the current time when running the command 'show running-config | no-more'


ytti:
I would have normalised the data in the block by ```cfg.sub```,  I worry if that type of exclude ! is available in all versions of EOS, but let's assume it is and fix if needed.

Issue #630, opened 2016-12-05
Errant changes on AOSW


thanegill:
I'm seeing whitespace changes and hardware info passed on my Aruba Controller configs. Might be a similar issue to #468.

![screen shot 2016-12-05 at 9 24 53 am](https://cloud.githubusercontent.com/assets/1141680/20895051/b646a3a4-bacc-11e6-8cc3-a45c6da177cb.png)
![screen shot 2016-12-05 at 9 24 40 am](https://cloud.githubusercontent.com/assets/1141680/20895055/b7c037c2-bacc-11e6-89d6-17e8c5e3c4bf.png)

ytti:
Seems to be two issues. The trailing whitespace is simple, you'll just add #rstrip on the model so that right hand side white spaces are always strippeed, then it does not matter if box gives them or not.

The latter one is classic error of desync in prompt detection, you need to change this https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/aosw.rb#L64  to ```cmd vars(:enable)``` so that the prompt following enable password prompt is consumed.

PR #634, opened 2016-12-07
[AOWS] Remove trailing whitespace and enable prompt detection

thanegill:
Fixes: #630


ytti:
Looks good. Another way:

```
#with keeping empty line feeds
str.each_line.map { |line| line.sub(/[ \t]+$/, '') }.join

#with removing even empty line feeds
str.each_line.map { |line| line.rstrip }.join("\n")
```

Issue #610, opened 2016-11-08
CiscoSMB Model doesn't work on cisco SG series (still small business)


Atroskelis:
I have a couple of cisco SG-200 specifically, and the model doesn't match. Not even the prompt.

I've tried modifying the model file, and i've reached to the point where the oxs commands returns the output of most commands (like show clock) but i cannot manage to make it output the "show running-config" command, as it ends with 

Here is the diff of what i've done to the file (i wasn't sure i could make a new model file and use it)

```
[oxidized@oxidized model]$ diff ciscosmb.rb backupsmb 
6c6
<   prompt /\n?([\w.@()-]+[#>]\s?)/

---
>   prompt /^\r?([\w.@()-]+[#>]\s?)$/
37,38c37,38
<     username /Username:/
<     password /Password:/

---
>     username /^User Name:/
>     password /^\r?Password:$/
41a42
>     post_login 'terminal datadump' # Disable pager
[oxidized@oxidized model]$
```

here's the output using the oxs 

```
[oxidized@oxidized model]$ oxs -m ciscosmb -u cisco -p <pass> -v -d <redacted> 'show clock'         
D, [2016-11-08T12:40:25.614364 #8714] DEBUG -- : resolving DNS for <redacted>...
D, [2016-11-08T12:40:25.614839 #8714] DEBUG -- : node.rb: resolving node key 'input', with passed global value of 'telnet' and node value ''
D, [2016-11-08T12:40:25.614877 #8714] DEBUG -- : node.rb: returning node key 'input' with value 'telnet'
D, [2016-11-08T12:40:25.619064 #8714] DEBUG -- : node.rb: resolving node key 'output', with passed global value of 'git' and node value ''
D, [2016-11-08T12:40:25.619114 #8714] DEBUG -- : node.rb: returning node key 'output' with value 'git'
D, [2016-11-08T12:40:25.633508 #8714] DEBUG -- : node.rb: resolving node key 'model', with passed global value of '' and node value 'ciscosmb'
D, [2016-11-08T12:40:25.633572 #8714] DEBUG -- : node.rb: setting node key 'model' to value 'junos' from global
D, [2016-11-08T12:40:25.633599 #8714] DEBUG -- : node.rb: returning node key 'model' with value 'ciscosmb'
D, [2016-11-08T12:40:25.633626 #8714] DEBUG -- : lib/oxidized/node.rb: Loading model "ciscosmb"
[...]

PR #635, opened 2016-12-07
Fix Cisco SMB prompt suggested by ytti

SaaldjorMike:
Related to #610

Hope it is ok with you @ytti that I took your suggestion and added it here. The current master branch implementation of the user name prompt is broken for all my 300 series Cisco SMB switches and would prefer to get this wrapped up.


ytti:
thanks!

Issue #236, opened 2015-11-09
Comware SMB switch


ujemvi:
Hi guys!
I'm having trouble with a Comware SMB switch. It's a HP V1910-24G.
The system connects just fine to the switch, but it seems that it's not using the '_cmdline-mode on' command. All I got when I look at the config is this:

```
#          ^
#  % Unrecognized command found at '^' position.
#          ^
#  % Unrecognized command found at '^' position.
         ^
 % Unrecognized command found at '^' position.
```

I've been trying to debug the comware gem, but I'm useless with this and I don't want to quit this awesome tool just for not being able to solve a problem with one of my devices.
Could anybody throw me a hand here?

ytti:
Hopefully @erJasp @sid3windr or @nickhilliard have some clue, I don't have access to any Comware box

erJasp:
@ujemvi Have you provided the unlock password with the `comware_cmdline` variable? What happens when you connect to the switch (with the same user as configured in oxidized), run `_cmdline-mode on` and provide the unlock password?

Can you post the [debug log](https://github.com/ytti/oxidized#debugging) of the telnet/SSH session?

admlko:
I had the same problem and here's what's going on:
First, the prompt doesn't work with the default hostaname <HP 1920G Switch>
So I modified comware.rb file from:
`prompt /^\0*(<[\w.-]+>)$/`
to allow whitespaces
`prompt /^\0*(<[\w\s.-]+>)$/`

Then the problem with unrecognized command:
The problem only exists when using Telnet, SSH seems to work just fine.
With Telnet, I noticed that
`post_login 'screen-length disable'`
`post_login 'undo terminal monitor`

commands were executed without newline, and that was causing the unrecognized command error:
`<HP 1920G Switch>screen-length disableundo terminal monitor`
`                               ^`
` % Unrecognized command found at '^' position.`
`<HP 1920G Switch>display version`
Sorry for the arrow position, it should be pointing to the beginning of `disableundo`

I was able to work around it by adding extra
`post_login ''`
after both commands but hey, that's pretty damn ugly.

If someone has an idea how to fix this behavior, that would be just great!

danilopopeye:
as per [#637 comment](https://github.com/ytti/oxidized/pull/637#issuecomment-266370781) changing the `prompt` should fix this.

sau70networkadmin:
I am using ssh, but getting the same (I think) error:

#                    ^
#  % Unrecognized command found at '^' position.
[...]

PR #637, opened 2016-12-09
Uptade comware.rb

srekkas:
Added sleep command, wait for CLI enable. 5 enough, bet 10 for safe side.
Second command runs before CLI enables and you get timeout message.
It works, but i dont know or it is correct.

### DIFF
--- a/lib/oxidized/model/comware.rb
+++ b/lib/oxidized/model/comware.rb
@@ -32,6 +32,7 @@ class Comware < Oxidized::Model
         send "_cmdline-mode on\n"
         send "y\n"
         send vars(:comware_cmdline) + "\n"
+        sleep(10) # wait for CLI enable
       end
     end
 


ytti:
```   
 if vars :comware_cmdline
      post_login do
        send "_cmdline-mode on\n"
        send "y\n"
        send vars(:comware_cmdline) + "\n"
      end
    end
``` 

Do any of these commands output prompt? If so, that should be changed to ```cmd``` from ```send```, because ```cmd``` will wait for prompt, ```send``` won't, causing next ```cmd``` to potentially think end-of-output before any output appears, because there is 'dangling' prompt.

srekkas:
Send command do not output to debug screen, only cmd. I try tshark capture and see what without sleep command it try next command faster than CLI enables. With CMD do not work at all.
It works fine with 1910, but with 1920 it outputs:
_10.20.12.138 raised Oxidized::PromptUndetect with msg "unable to detect prompt: (?-mix:^\0*(<[\w.-]+>)$)"_
 Somehow it does not send "_cmdline-mode on" for 1920. 

Router.db:
_10.20.13.208:HP:up:512900
10.20.12.194:HP:up:512900
10.20.12.138:HP:up:Jinhua1920unauthorized_

config:
    _map:
      name: 0
      model: 1
      state: 2
    vars_map:
      comware_cmdline: 3
model_map:
  cisco: ios
  juniper: junos
  HP: comware_

srekkas:
I found similar issue: https://github.com/ytti/oxidized/issues/236
It works after i chaged from link:
_So I modified comware.rb file from:
prompt /^\0*(<[\w.-]+>)$/
to allow whitespaces
prompt /^\0*(<[\w\s.-]+>)$/_

ytti:
I didn't catch quite from your respond, of any of the send() commands produces prompt after entering it? Anything that produces prompt, MUST consume prompt. Otherwise we get desynchronised, and we don't know when command output begins and end.

When you enter these in CLI manually, do any of the commands produce output which is terminated by prompt?

InsaneSplash:
Is there anything I could add to help provide support for the ComWare devices using the *Secret* login routine?

ytti:
What is the output from device after entering this:
```send vars(:comware_cmdline) + "\n"```, I expect it will output prompt, if so, then the model is broken.

```send``` should be ```cmd```, as ```cmd```will wait for and consume the prompt.  ```send``` does not, by default, expect anything, so if prompt does come out, it's not consumed, and it'll break the whole flow of the model, as next ```cmd``` that runs, may immediately finish, without reading any output, because it incorrectly read the old prompt as end-of-output.

Alex33150:
Oxidized 0.19.0

Hello,

None of the proposed solutions work. It remains an error to use the 

    post_login 'screen-length disable'
    post_login 'undo terminal monitor'
    pre_logout 'quit'post 

when retrieving the HP1910 configuration. I have the conf file cut by one and with a timeout an idea ????

Last question how to dissociate in the config file the use of different switch model using both the comware.rb.
HP5120 H3C and HP1910 (using the vars_map: comware_cmdline)?

Thank you

InsaneSplash:
Ive tested this using the bundled comware.rb file with no issues. The only requirement is to set the special password as a var in the configuration file.

danilopopeye:
@srekkas is this still a issue? since @InsaneSplash reported the model working, can you try with a newer version of oxidized?

srekkas:
Updated to 0.2.Then migrated to ssh, with ssh everything works OK. Thank
you.

2017-08-22 0:16 GMT+03:00 Danilo Sousa <notifications@github.com>:

laf:
@srekkas Can we close this then?

srekkas:
Yes. Thanks.

2017 rugp. 22 02:53 pop. "Neil Lathwood" <notifications@github.com> rašė:

PR #641, opened 2016-12-14
delete secret password if it is called secret

mattttv:
hello ytti!
First of all: thanks for this project :)
I encountered the problem that some router-configs do not call their password entries "password" at it is expected in your lib/oxidized/model/ios.rb. This leads to the problem that the secrets are still there due to the fact that the regex does not match.
I took the change to add two substitutions that also match this case.
Do you need anything else from me or is this PR fine for you?

Thank you in advance,
Kind Regards,
Matthias


ytti:
Thanks!

Other way to write this would've been (?:secret|password)

mattttv:
thank you for accepting the PR :)

PR #653, opened 2016-12-27
Hide remaining passwords

### LINKED ISSUE #620: remove_secret doesn't remove all password on Cisco ios
the option remove_secret: true
does not remove all encrypted passwords on Cisco ios devices from the configuration which could be deemed sensitive such as the configuration row:
"enable secret 5 $1$Ac8t$jaszMu.tsfEsdd3qa/KwBN/"

gaetronik:
Should fix #620

### DIFF
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -27,8 +27,8 @@ class IOS < Oxidized::Model
     cfg.gsub! /username (\S+) privilege (\d+) (\S+).*/, '<secret hidden>'
     cfg.gsub! /^username \S+ password \d \S+/, '<secret hidden>'
     cfg.gsub! /^username \S+ secret \d \S+/, '<secret hidden>'
-    cfg.gsub! /^enable password \d \S+/, '<secret hidden>'
-    cfg.gsub! /^enable secret \d \S+/, '<secret hidden>'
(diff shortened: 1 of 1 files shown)


ytti:
Where is the line 31 intended to match?

gaetronik:
In one of the switch I just get access to there is lines like:
```
line vty 0 4
 password <password>
```
I could prepend with \s+ in the regexp if you prefer

ytti:
How would this match to it?

```cfg.gsub! /(password|secret) \d \S+/, '\\1 <secret hidden>'```it is expecting number after the password, in your example above there is just raw password after it, so it wouldn't match.

Could this ever be ```secret```?

Wouldn't be safer to do:
```cfg.sub! /^ password \S+/, "password <secret>"```

**Comment by gaetronik** [comment]
Just checked and I have also lines like that:
```
line vty 0 4
  password 7 <password>
````
Will update the commit

gaetronik:
updated

ytti:
```cfg.gsub! /^\s+(password|secret) (\d |)\S+/, '\\1 <secret hidden>'```

Can there be ```secret```? Should you retain same indent in the generated output? Maybe:

```cfg.gsub /^(s+(?:password|secret) (?:\d )?\S+/, '\\1 <secret hidden>'```

gaetronik:
You're totally right. I updated the PR

Issue #396, opened 2016-04-15
Nested Groups?


tdurack:
Does Oxidized support nested groups? Something like:

Country/Pop/Core
Country/Pop/Agg
Country/Pop/Access

Etc.

Thanks in advance!

Tim:>

ytti:
Not right now, but this was recently brought up in recent ticket where user did this, and ended up getting configs in those directories but also in parent. 

I think this is useful and something we should support and it does not appear to be very complicated to support. But I won't have time to work on it any time soon.

davama:
I believe i tried this using a file as output. It did create a folder for each group with the configs inside the folders. Now using git output and a repo for each group is created.

nested groups would be convenient.
can you point me to the right direction so i can take a stab at it?

Thanks

ytti:
I guess output/git.rb output/file.rb etc. Split group by '/' and go from there.

davama:
Folder within folder worked. So it looks like so
using git as output
`GROUP/subgroup{a,b,c}.git  
`

or with file as output: ( these are all flat files of course)
`GROUP/subgroup{a,b,c}/device{a,b,c}
`

I was attempting to combine the two but noticed that the last "default" takes precedence. in this instance the default output is a file.
```
output:
   default: git
   git:
    user: config
    email: config@nwk.jwm2.net
    repo: /home/config/repository/.git
   default: file
   file:
     directory: /home/config/repository/.
```

I was wondering is this is possible?
```
GROUP/subgroupa/deviceflatfile
GROUP/subgroupa/.git
```

So the output would be saved on a flat file and the git commit would monitor all the flat file devices.

Hope it makes sense.

Thanks
Dave

davama:
Think i found a solution for this.

Noticed the hooks section and so i changed the config to this:

```
output:
   default: file
   file:
     directory: /home/config/repository/.
hooks:
   first_hook:
     type: exec
     events: [post_store]
     cmd: 'cd /home/config/repository/$OX_NODE_GROUP && git add --all . && git config user.name "config" && git config user.email "config@nwk.jwm2.net" && git commit --quiet -m "$OX_NODE_GROUP"'
     timeout: 10
```

So in the end i still have a visible flat device file which is monitored by git. 
Structure looks like so:

```
GROUP/subgroupa/deviceflatfile
GROUP/subgroupa/.git
```

May not be an elegant solution. 
What do you guys think? Any "oxidized like" solution? Multiple outputs?

Thanks
Dave

[...]

Issue #471, opened 2016-06-15
Force Config Pull


davama:
Is it possible on an already running oxidized process to force oxidized to pull the config on a particular device?

Thanks
Dave

laf:
You can call for the device to be backed up next in the Webui / api:

/node/next/default/myrtr.example.com is one example.

davama:
Right!!
completely forgot.
Any CML way ?

laf:
curl :)

davama:
Ok, interestingly oxidzed webui/api says "link status" is red when it's actually alive and well. DNS is fine too. 
i clicked on "update" button and got this page:
```
Sinatra doesn’t know this ditty.


Try this:
# in webapp.rb
class Oxidized::API::WebApp
  get '/node/next/CAN/FTTX/CAN-A2-152-SW-5/CAN-A2-152-SW-5' do
    "Hello World"
  end
end
```
Dont know if it's timing out or what but we do have a few thousands of devices on this oxidized process, if that makes a difference. Running curl didn't help, unless doing it wrong:
`curl -k https://<ip-addr>:9002/node/next/CAN/FTTX/CAN-A2-152-SW-5/CAN-A2-152-SW-5`
This gave me the above 'sinatra' output but in html format.

Ran oxidized in debug mode with only that switch in the DB and it's timing out, interesting...
I increase the timeout to 30 and that worked.

But still couldnt get it to force pull

laf:
We just hit:

`curl http://ip/node/next/HOSTNAME` which works for us.

Do your group names contain /'s? Running the latest version?

davama:
yes to divide by groups.
and yes

danilopopeye:
@davama can you post the `config` and the `router.db` line for this device?

davama:
DB table name is "devices"
Table fields/columns: "name,model,group"
```
echo "select * from devices where Name like '%CAN-A2-152-SW-5%';" | sqlite3 repository/region-oxidized.db
```
which outputs:
` 
CAN-A2-152-SW-5|ciscosmb|CAN/FTTX/CAN-A2-152-SW-5
`

Config: (posting only what i think may be relevant, let me know if otherwise)
```
.....
output:
   default: file
   file:
     directory: /home/config/repository/.
.....
source:
  default: sql
  sql:
    adapter: sqlite
    database: "/home/config/repository/region-oxidized.db"
    table: devices
    map:
      name: name
      model: model
      group: group
    vars_map:
      enable: enable
[...]

PR #654, opened 2017-01-03
Nested Group Support

davama:
Hello this is for requested enhancement on #396 . Having nested group using /

I'm not really sure if i did this PR correctly. I understand you would not be able to accept the changes of my fork. Just forked it to create this ticket.

Please let me know if there's any info needed from me.

Thanks!
Dave

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,13 +188,6 @@ gem install oxidized
 gem install oxidized-script oxidized-web
 ```
 
-## Building from Git
-```shell
-git clone
-rake install
-gem install pkg/*.gem
-```
-
 # Configuration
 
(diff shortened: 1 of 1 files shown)


ytti:
Unsure if I actually want to document this as supported, since if it works, it kinda works on accident.  Code doesn't really support subgroups.

davama:
>  it kinda works on accident

:D 

i guess the only issue is [this](https://github.com/ytti/oxidized/issues/471#issuecomment-226234994) cannot see on api because of the /

ytti:
I'm pretty sure there are other issues, but that's at least one issue.

davama:
I understand the concerns.

So i guess at this point this is an accidental "partially" working feature with no hopes for official support.... at this time at least :-)

ytti:
I think this would be good feature, but I don't see personally working towards it. If someone else does, grand. 
If I ever get around rewriting the core, support like this probably would come in as a side effect to that work.

laf:
@davama Proposing that we close as this is just a doc change but wouldn't be officially supported.

davama:
Agreed, closing

laf:
Thanks very much.

PR #662, opened 2017-01-10
Mask out configuration date for fiberdriver

devicenull:
Our fiberdriver devices report a timestamp during 'show running-config', which results in a config diff on every run.  This patch removes the '+! Configuration saved on 2017/01/10 14:21:20' line from the config, as well as a couple other useless status messages


ytti:
```String#gsub!``` is in-place/destructive method. IIF it modifies, it also returns the modified string. If it does not modify, it returns nil. 

So if it would ever happen that this code finds nothing to modify, the block returns nil.

I suggest you add ```cfg``` before and, so we guarantee block always returns value of ```cfg```, not return value ```String#gsub!```

devicenull:
Updated, I think that's what you were trying to say there.

ytti:
Yes.

PR #665, opened 2017-01-11
Add suggestion to set `ip` variable in CSV reader

devicenull:
We have around ~1000 nodes, and oxidized was taking 30s to do the initial config load.  After adding the 'ip' variable to our CSV, load time dropped to less than 1 second.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,7 +409,7 @@ vars:
 
 ### Source: CSV
 
-One line per device, colon seperated.
+One line per device, colon seperated. If `ip` isn't present, a DNS lookup will be done against `name`.  For large installations, setting `ip` will dramatically reduce startup time.
 
 ```
 source:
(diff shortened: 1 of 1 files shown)


ytti:
If you do this in the machine where oxidized runs:
```
% ruby -rresolv -e 's=Time.now;Resolv.new.getaddress ARGV[0];p Time.now-s' ytti.fi
0.005834167
```

Replace ```ytti.fi``` with one of your names, how long does it take?
I my example it took 0.0058s or 5.8ms, which would be 5.8seconds resolve time for 1000 nodes. 30 sounds lot, but perhaps still reasonable.

Startup should be rare, so startup time usually shouldn't matter. But still good to point in documents for sure.

devicenull:
We use some pretty slow active directory DNS servers... based on the command you posted, our DNS takes 10x longer (0.056768616)

ytti:
Thanks for the information.

davama:
This a great enhancement! we too have many devices. ~8000
Tested with sqlite and startup is also improved.

Not sure if necessary to update the README file on the SQLite source section

Thanks

PR #667, opened 2017-01-18
Update ironware.rb

devicenull:
Some Brocade devices only have two possible fan speeds.  Previously, this only handled three levels of fan speeds.  Updated regex will handle both of these cases:

    Fan 1 ok, speed (auto): 1<->[[2]]
    Fan 1 ok, speed (auto): 1<->[[2]]<->3


ytti:
perhaps ```cfg = cfg.tr('[]', '') if cfg.match(/^Fan /)```would've been cleaner and more robust

ytti:
or ```cfg = cfg.delete('[]') if cfg.match(/^Fan /)```

ytti:
aah both would require you parse the cfg line by line:
```
cmd 'blaah do |cfg|
  cfg.lines.map do |line|
    ! mangle lines here
  end.join
end
```

PR #673, opened 2017-01-24
Support lowercase U and P on telnet login prompt.

network-guy:
Cases where "username:" and "password:" are all lowercase, seen in older IOS versions when connecting via telnet.

### DIFF
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -53,8 +53,8 @@ class IOS < Oxidized::Model
   end
 
   cfg :telnet do
-    username /^[U,u]sername:/
-    password /^[P,p]assword:/
+    username /^[Uu]sername:/
+    password /^[Pp]assword:/
   end
 
   cfg :telnet, :ssh do
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -53,8 +53,8 @@ class IOS < Oxidized::Model
   end
 
   cfg :telnet do
-    username /^Username:/
-    password /^Password:/
+    username /^[U,u]sername:/
+    password /^[P,p]assword:/
   end
 
   cfg :telnet, :ssh do


ytti:
```[U,u]```means capital u, comma or lower case u. You probably intended [Uu]. I don't consider your IOS old. And I don't recall we fully explored the possibility that the prompt is in fact presented by TACACS config (Which is possible).

network-guy:
Fixed, thanks. It is possible that it's being presented by TACACS, I didn't think so, but I don't manage it. I don't see any downsides to making the U in username and P in password case-insensitive.

ytti:
If it is presented by TACACS, then clearly it can be arbitrary. What is the upside of accepting uU in case it's user defined prompt? Then it will help exactly on your case, not on anyone elses?
I.e. you can edit group or node prompt, or you can subclass the ios.rb and edit local version of it in ~/.config/oxidized.

If it is not coming from TACACS, I'm more than happy to accept this. I mean it is possible that there was some IOS version where they had this type of bug.

Would it be too much trouble to

a) review that prompt is not set anywhere in config
b) capture TACACS communication via tshark/tcpdump and review in wireshark to see if prompt is being passed.

I'm not strongly against merging this, even if it is site local, I just feel like the added complexity and the weaker guarantee (maybe someone has MOTD where there is string username:?) of correct prompt detection require that we at least understand what we are doing.

network-guy:
I have confirmed this is in fact being set by TACACS. I'm OK with not merging and using a custom model file unless there are others that feel this would be useful.

ytti:
I encourage you to change your tacacs configs, this is path to pain for any screen-scraping solution you're going to deploy, and I can see no benefits.

In case you are unable, I am happy to walk you through how to configure Oxidized in your environment for these prompts, so that change does not get ovewritten during upgrades.

Reason why merging does not seem to make much sense, we cannot know what is the next customer tacacs prompt, will it be exactly like your custom tacacs prompt? Likely not, so then we'd keep adding complexity to prompt detection, and risk matching the regexp in say 'motd' message prior to prompt.

danilopopeye:
so, we can close this @ytti ?

danilopopeye:
since both the PR author, @network-guy and @ytti seems to be 🆗 with closing

PR #674, opened 2017-01-24
Add GPG support for reading router.db

elmobp:
Add support for GPG encryption for router.db, open for feedback!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,2 @@
 source 'https://rubygems.org'
-gem 'gpgme'
 gemspec
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
 source 'https://rubygems.org'
-
+gem 'gpgme'
 gemspec
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ source:
     file: ~/.config/oxidized/router.db
     delimiter: !ruby/regexp /:/
     gpg: 'false'
+    gpg_password: 'password'
     map:
       name: 0
       model: 1
--- a/README.md
+++ b/README.md
@@ -235,12 +235,14 @@ oxidized
 
 Now tell Oxidized where it finds a list of network devices to backup configuration from. You can either use CSV or SQLite as source. To create a CSV source add the following snippet:
 
+Note: If gpg is set to anything other than false it will attempt to decrypt the file contents
 ```
 source:
   default: csv
   csv:
     file: ~/.config/oxidized/router.db
     delimiter: !ruby/regexp /:/
+    gpg: 'false'
     map:
       name: 0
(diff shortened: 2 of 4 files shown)


lib/oxidized/source/csv.rb
```diff
@@ -11,14 +11,21 @@ def setup
...
   end
 
   def load
     nodes = []
-    open(File.expand_path @cfg.file).each_line do |line|
+    if @cfg.gpg != 'false'
+      crypto = GPGME::Crypto.new :password => @cfg.gpg_password
+      file   = crypto.decrypt(File.open(@cfg.file)).to_s
+    else
+      file = open(File.expand_path @cfg.file)
```
ytti: Perhaps ```
file = File.expand_path(@cfg.file)
file = if @cfg.gpg?
  crypto = GPGME::Crypto.new password: @cfg.gpg_password
  crypto.decrypt(file)
else
  open(file)
end  
```

File: README.md [hunk 2]
```diff
@@ -235,12 +235,15 @@ oxidized
...
 Now tell Oxidized where it finds a list of network devices to backup configuration from. You can either use CSV or SQLite as source. To create a CSV source add the following snippet:
 
+Note: If gpg is set to anything other than false it will attempt to decrypt the file contents
 ```
 source:
   default: csv
   csv:
     file: ~/.config/oxidized/router.db
     delimiter: !ruby/regexp /:/
+    gpg: 'false'
```
ytti: ```gpg: false```  not string false

Gemfile
```diff
@@ -1,3 +1,3 @@
 source 'https://rubygems.org'
-
+gem 'gpgme'
```
ytti: We don't put requirements in Gemfile, they're in spec file. But I would not want to mandate installing ```gpgme```. I would put it in ```CVS#setup``` as ```require 'gpgme' if @cfg.gpg```, so  that we only load it, if user is actually using it.

elmobp:
@ytti apologies :) hows that looking?

lib/oxidized/source/csv.rb
```diff
@@ -10,15 +10,23 @@ def setup
       Oxidized.asetus.user.source.csv.file      = File.join(Config::Root, 'router.db')
       Oxidized.asetus.user.source.csv.delimiter = /:/
       Oxidized.asetus.user.source.csv.map.name  = 0
-      Oxidized.asetus.user.source.csv.map.model = 1
```
ytti: Is this line being removed intentionally? If so, can you try to elaborate why.

elmobp: Sorry! Back in again :)

ytti:
Many thanks!

Issue #669, opened 2017-01-20
Aruba Mobility Access Switch does not provide "show slots" or "show license" commands


vppencilsharpener:
In working through an Aruba S2500-24P switch I found that the "show slots" and "show license" command result in a simple error in the output file because the commands do not exists in the firmware. There errors are not really hurting anything so it may be a non-issue, but I like things clean. 

I am more than willing to fix this and submit a pull request, but I am rather new to the Oxidized system and do not know how best to handle this within the framework. 

Should this be a new model file?
Possibly AOSMAS (Aruba OS Mobility Access Switch)  to complement AOSW (Aruba OS Wireless controller)

If not how should this be structured?

--
I currently have a proof of concept installation of Oxidized that I am working with and loving. Thank you for making this. 

I hope to spend some time making the Aruba and possibly the Aruba/HPE models better.

ytti:
If we can reasonably avoid duplicate models, we should. 

If there are some commands which are only valid in other platform, it should be somewhat simple matter of just regexping the complaint out of the output, leaving no output for systems which don't support the commands.

We can also in model gather stuff like version, and store information about the platform in variable, and then run different commands for different hw models.

It's ultimately up-to-you which solution is cleanest, all new model or more complexity to existing model.

vppencilsharpener:
I though about using regex to clean the error, but worry about a legitimate error not getting written. 

I think handling different devices within the model may be the best approach as Aruba currently uses ArubaOS for different devices with different capabilities. For example the latest stable ArubaOS version for the Access Mobility Switches is 7.4.1.7. For Aruba Instant the latest stable ArubaOS version is 6.4.4.8-4.2.

Can you recommend a model to review that gathers information and runs different commands based on that information? I want to see how it has been done before so I can try to be consistent. 

From our hardware both are reported as ArubaOS
Access Mobility Switch: ArubaOS (MODEL: ArubaS2500-24P), Version 7.4.1.4
Aruba Instant: ArubaOS (MODEL: 205), Version 6.4.2.6-4.1.1.11

ytti:
If the error is like 'invalid command', then what is there to worry if you match exactly to that line and remove that line? Then if that command complains about invalid command, it's gone but it'll only complain on boxes not supporting it?

I think there are some legit models doing conditional commands, but there is still day1 contrived example in junos.rb (it's not actually that anyone needed that, i just wanted to add to the model example how to do it without having use case). I know some model uses that for real use-case, just can't remember which.

[...]

PR #675, opened 2017-01-26
Don't show error is unsupported

vppencilsharpener:
This is the proposed fix for issue #669. Let me know if there is anything that I need to fix.
This appears to be working correctly for the switch I have in house (S2400-24P). However the model is not working for the Aruba Instant AP (IAP) (virtual) controller. I hope to submit a fix for those once I have it working. The problems with that device appear to be unrelated to these changes as the prompt is incorrect.


vppencilsharpener:
Title should be "Don't show error IF unsupported"

PR #683, opened 2017-02-02
Add git build to README

davama:


ytti:
```rake install``` already installs it, ```rake build``` just builds package, for manual installation with gem (perhaps on other machine, or perhaps installation as other user).

PR #687, opened 2017-02-02
Node - Cumulus additional params

nertwork:
Just adding additional things to check for with Cumulus Quagga configs.

### DIFF
--- a/lib/oxidized/model/cumulus.rb
+++ b/lib/oxidized/model/cumulus.rb
@@ -48,6 +48,9 @@ class Cumulus < Oxidized::Model
     cfg += add_comment 'QUAGGA OSPF6'
     cfg += cmd 'cat /etc/quagga/ospf6d.conf'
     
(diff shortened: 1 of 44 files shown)


nertwork:
Also added in ios gsub for hidden secrets.

lib/oxidized/model/cumulus.rb
```diff
@@ -32,7 +32,10 @@ def add_comment comment
     
     cfg += add_comment 'IP Routes'
     cfg += cmd 'netstat -rn'
-
+    
+    cfg += add_comment 'SNMP settings'
+    cfg i= cmd 'cat /etc/snmp/snmpd.conf'
```
ytti: Sure about this?

nertwork: Oh thanks for the catch - I'll update, test, and submit a new PR.

PR #690, opened 2017-02-03
feature: support an IP Address provided with a netmask

InsaneSplash:
If an IP address is provided with a netmask, stip the netmask and keep the IP Address.

### DIFF
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -11,8 +11,7 @@ module Oxidized
     def initialize opt
       Oxidized.logger.debug 'resolving DNS for %s...' % opt[:name]
       # remove the prefix if an IP Address is provided with one as IPAddr converts it to a network address.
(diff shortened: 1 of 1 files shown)


lib/oxidized/node.rb
```diff
@@ -10,8 +10,12 @@ class Node
     alias :running? :running
     def initialize opt
       Oxidized.logger.debug 'resolving DNS for %s...' % opt[:name]
+      # remove the prefix if an IP Address is provided with one as IPAddr converts it to a network address.
+      ip_address = opt[:ip]
```
ytti: isn't ip_address completely redundant?

Shouldn't this suffice:
```ip_addr, _ = opt[:ip].to_s.split("/")```

InsaneSplash: True :) late night... let me update as you suggested.

ytti:
thanks!

Issue #676, opened 2017-01-26
Problem with custom model for older FoundryOS devices


rjorgenson:
Hi,

Our environment has some older FoundryOS devices which is very similar to the IronWolf devices. I copied the IronWolf model and removed the logout command as it's not used and everything runs fine up until the user-mode prompt after quitting the shell. It just hangs but the prompt return matches the prompt regex in the model.

`prompt /^.*(telnet|ssh)\@.+[>#]\s?/i`

I removed the $ at the end of the regex because the debug log wasn't showing any EOL characters after the prompt.

`SW-SSH@host.name>`

The run eventually times out and all the output is in the ssh debug output with the user-mode prompt as the last line without any newline characters at the end.

`end`
`SW-SSH@host.name#quit`
`SW-SSH@host.name>`

Here is the pre_logout commands in the model

`    pre_logout 'quit'`
`    pre_logout 'exit'`

It never runs the exit command and eventually times out. Does anyone have any ideas why that regex would not be matching that prompt to run the second pre_logout command?

rjorgenson:
I found #208 and tried the suggested fix of just sending one exit command and that did not produce any results =/

wk:
This appears to be resolved in #693.

PR #693, opened 2017-02-04
Restore compatibility with older versions of IronWare

### LINKED ISSUE #208: Ironware never exit.
I have set up oxidized for some ironware devices.

I can connect and get the configuration, but at the end it's stop at the ">" prompt. :

telnet@ro-rku-1#exit
telnet@ro-rku-1>

After that it retry. and stop again at the same point.

I just comment the logout command on the pre_logout because the command doesn't exist in my devices.

Anyone as an idea ? I try on ironware 5.2, 5.4 and 5.6

wendelj:
Remove \r from the "enable" command. Both \r and \n are interpreted the same on older versions, causing a blank line to be sent as the enable password.
Add a delay between the sending of the "logout" command and each "exit" command. Older devices don't seem to be able to keep up with the speed the commands are sent without a delay.
Fixes #208 and #676.

### DIFF
--- a/lib/oxidized/model/ironware.rb
+++ b/lib/oxidized/model/ironware.rb
@@ -74,16 +74,20 @@ class IronWare < Oxidized::Model
   cfg :telnet, :ssh do
     if vars :enable
       post_login do
-        send "enable\r\n"
+        send "enable\n"
         cmd vars(:enable)
       end
     end
     post_login ''
     post_login 'skip-page-display'
     post_login 'terminal length 0'
-    pre_logout 'logout'
-    pre_logout 'exit'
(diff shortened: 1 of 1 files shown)


ytti:
If there is problem with this, we probably just want to have 'send' expect for something, instead of blindly sleeping. We can expect for arbitrary regexp.

wendelj:
After your comment I went looking for a better way to handle this.
Looks like this also works:
pre_logout "logout\nexit\nexit\n"
Should I create a new PR to change it?

ytti:
Sure. I think relying on timing is dodgy.

PR #701, opened 2017-02-07
add support for hp msa storage

aschaber1:
tested with hp msa 2040 storage system

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,7 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
    * [Comware (HP A-series, H3C, 3Com)](lib/oxidized/model/comware.rb)
    * [Procurve](lib/oxidized/model/procurve.rb)
    * [BladeSystem (Onboard Administrator)](lib/oxidized/model/hpebladesystem.rb)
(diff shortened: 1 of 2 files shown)


ytti:
1. What does the ```telnet``` method do there?
 2. Why is there expect for more prompt, when pager is actually disabled?
 3. Why is telnet in cfg block, while there is no telnet username/password prompt shown?

It is perfectly fine to submit SSH only model. Only submit what you've tested and known to work. To me it seems you can remove telnet from config, telnet model and the expect pager thing.

ytti:
Also please edit README.md accordingly. 

We really prefer if the models are called by the operating system name, so if you can find operating system name, use that. Spend some time researching, we understand often there simply isn't any particular name for the operating system.

ytti:
bump

ytti:
bump

laf:
@aschaber1 Any updates to this?

aschaber1:
hello @ytti, @laf 
just rebased/pushed the cleanup.
I didn't find any other os name, so I'd stick with hpmsa.
regards

laf:
You've got two model files now, hpemsa and hpmsa, can you remove the redundant one?

PR #702, opened 2017-02-07
add support for netgear switch

aschaber1:
tested with netgear m5300 series switch
unfortunately reading config requires enable..

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,8 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
  * MRV
    * [MasterOS](lib/oxidized/model/masteros.rb)
    * [FiberDriver](lib/oxidized/model/fiberdriver.rb)
+ * Netgear
+   * [Netgear](lib/oxidized/model/netgear.rb)
  * Netonix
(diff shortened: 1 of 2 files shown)


ytti:
Same comments are your other model. The telnet method seems unused, you don't seem to specify telnet password/username prompt, so telnet wouldn't work anyhow. 

Is there no way to disable pager in netgear?

Also you need to send the enable password with ```cmd``` not ```send```, because ```cmd``` will wait for a prompt, ```send``` won't, and this will cause problems.

ytti:
bump

bbrendon:
I'd like to use this on a Prosafe m4100 . Any idea how? If I provided some details about how the login process worked, could someone create the expect script or whatever it is that oxidized uses?

aschaber1:
@ytti if this looks better, I can rebase it of you'd like..

bbrendon:
```
For this hostname:
current regex doesn't match: ^(\([\w-]+\)\s[#>])$
This does match :            ^(\([\w-+]+\)\s[#>])$
```

PR #703, opened 2017-02-07
added tplink model

mediumo:
added model for tp-link switches.

Had to built in some workarounds because of strange carriage-return, line-break an whitespace behaviour.
tested with:
Hardware Version     - TL-SG3424P 2.0
Softare Version     - 1.0.7 Build 20141219 Rel.33340(s)

### DIFF
--- a/lib/oxidized/model/aosw.rb
+++ b/lib/oxidized/model/aosw.rb
@@ -5,43 +5,50 @@ class AOSW < Oxidized::Model
   # Also Dell controllers
 
   comment  '# '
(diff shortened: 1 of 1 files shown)


ytti:
You need to use ```cmd vars(:enable)´´´ instead of ´´´send vars(:enable)```, send won't wait for prompt, cmd waits for prompt.

Also splease update README.md. 

The fortios change really should be in different PR.

mediumo:
follow up in #728 from a clean branch

PR #715, opened 2017-02-10
Next with new thread functionality (v2)

natm:
Updated PR to include only functionality to increase the job pool by 1 when /next is called.

This feature is gated by the `next_adds_job: true` config option.

Cherry picked commits from this branch https://github.com/ytti/oxidized/tree/next-with-new-thread


laf:
Nice work @natm

Worth adding to the docs?

natm:
@laf yes I will update docs via a different PR

natm:
@laf README.md updated in https://github.com/ytti/oxidized/pull/718

PR #716, opened 2017-02-10
nxos.rb: Add secret block

davama:
Took config from our local setup.
Adding since it does not have it

Thanks
dave


ytti:
thanks!

PR #718, opened 2017-02-13
Update documentation to include next_adds_job

natm:
Adding notes on when you might want to set:

```
next_adds_job: true
```

Docs as suggested by @laf


laf:
Thanks @natm and @ytti :)

PR #723, opened 2017-02-15
Updated timos.rb to ignore "Generated" and "Finished" lines

pozar:
so it doesn't look like a new config each time to crawls the box.


ytti:
You know we run ```:all``` for all commands, so you could have had it just there.

pozar:
Ok.  Will redo that.  More eloquent.

PR #725, opened 2017-02-15
Dockerfile support any branch

natm:
The existing Dockerfile installs oxidized from a published gem. This does not support building containers for development from any branch.

Proposed changes build the docker image from the git clone you currently have checked out locally.

🐳

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,9 +3,22 @@ MAINTAINER Samer Abdel-Hafez <sam@arahant.net>
 
 RUN add-apt-repository ppa:brightbox/ruby-ng && \
 	apt-get update && \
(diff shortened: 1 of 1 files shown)


ytti:
It looks like it only builds it from git. Shouldn't we support this optionally? I don't at all guarantee the git repo works, I also do WIP stuff in the main repo.

natm:
The original Dockerfile can only build from a git clone too, due to this line in the spec file... https://github.com/ytti/oxidized/blob/master/oxidized.gemspec#L17

Currently any push to the master branch results in the public `oxidized/oxidized` Docker hub image being updated. If there is a worry about pushes/merges into master breaking the build, then I'd suggest the docker hub automation is also changed to only build from a release branch.

ytti:
you're right, i'll merge this and we'll just assume that docker image is for HEAD, for better or worse, i.e. might be broken, might not.

PR #752, opened 2017-03-03
make mellanox config versioning much less noisy

electriceel1985:
When pulling Mellanox configs, the values for CPU info, System memory info, Uptime, and "Generated at" constantly change, resulting in a new configuration diff, file save, or git commit every time the configs themselves are fetched by Oxidized. This results in having to sift through tons of commits to find real network configuration changes.

Using the gsub method, these fields are removed when the configs are fetched. Versioning only occurs when a significant change is detected (OS versions, interface changes, inventory serial numbers, etc...)


ytti:
I generally don't delete the lines or replace with empty line, rather I mask the changing bits to NNN or XXX or something like that. But good enough for me.

PR #757, opened 2017-03-07
added model for OneAccess OneOS Voice Gateways/Routers

crami:
Slight modifications to gemspec for newer ssh gem and correct telnet syntax

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
    * [WISP Switch (As Netonix)](lib/oxidized/model/netonix.rb)
  * Nokia (formerly TiMetra, Alcatel, Alcatel-Lucent)
    * [SR OS (TiMOS)](lib/oxidized/model/timos.rb)
+ * OneAccess
+   * [OneOS](lib/oxidized/model/oneos.rb)
  * Opengear
    * [Opengear](lib/oxidized/model/opengear.rb)
  * Palo Alto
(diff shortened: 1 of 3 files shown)


ytti:
Why that Net::SSH version? If we're gonna move forward, we should go to 4.1 straight. We had specific reason to run 3.0.2, there was bug in newer ones, but /I Think/ it's fixed.

Please add the OneOS to README.md as well.

crami:
Had Problems with the old version (3.0.2) connecting. 3.2 was the newest which worked without code change. Will Add OneOS to the Readme. Should I do a new or additional pull request?

ytti:
You can add README.md change here. I'm not going to jump to 3.2, I don't know if it works, the issue did not affect everyone and every model. If we're gonna move forward, might as well go to 4.

crami:
Tested again with net-ssh 3.0.2 and it works now with some config settings. So we can go with that unchanged.

taro-75:
Hello,
I have a problem with router OneAccess: oneos6 when i use oneos.rb
My congif for Oxidized work fine with Cisco router
i can't connect Oneaccess Router:
raised Oxidized::PromptUndetect with msg "unable to detect prompt: (?-mix:^([\w.@()-]+#\s?)$)"

When can i find the regex on my router model plz ?

    show version
    Software version : ONEOS6-ADVIP_11N-V5.2R1E2_HA1
    Software created on : 04/03/15 17:03:22

Thank You
@taro-75

Issue #756, opened 2017-03-07
Slack: Option to only send a link to diffs into slack


makzdot:
Hi,

Great stuff oxidized. I was wondering is it possible to create an option to only post a link to slack instead entire snippets with possible secrets. (We want our secrets backupped). I get kinda nervous when posting secrets to a cloud solution like slack.

Kind regards.
Max

ytti:
No, but I'll accept PR if you want to add that support.

natm:
I will add this to my todo list, it seems like a good ask.

natm:
@makzdot as promised see https://github.com/ytti/oxidized/pull/763

makzdot:
Hi @natm, Great news thanks. I'll try it out ASAP

laf:
This has been merged in now.

PR #763, opened 2017-03-10
Slack hook - optionally disable diff snippets and post custom messages

natm:
As requested by @makzdot https://github.com/ytti/oxidized/issues/756 - in some environments posting a diff snippet to a slack channel may not be ideal, for instance if secrets are not removed.

This change includes the ability to optionally disable the diff snippet posting:

```yaml
    type: slackdiff
    diff: false
```

And post a custom message instead:

```yaml
    type: slackdiff
    diff: false
    message: "%{node} %{group} %{model} updated https://github.com/natm/home-network-configs/commit/%{commitref}"
```

I left the message customizable so users can include links to any git repo or service.

**Examples**

Default with a diff snippet:

screen shot 2017-03-10 at 18 47 13

Diff disabled, formatted message:

screen shot 2017-03-10 at 18 48 04


ytti:
You could query boolean key with cfg.key?

If cfg.diff?
  ...
end

if cfg.message?
  ...
end


Might be bit cleaner. Let's you get rid of whole:
```
 +        diffenable = true
 +        if cfg.has_key?('diff') == true
 +          if cfg.diff == false
 +            diffenable = false
 +          end
 +        end
```

makzdot:
Also the slack channel is hard coded:

Line 44:
- client.chat_postMessage(channel: "#oxidized-test", text: msg,  as_user: true)
+ client.chat_postMessage(channel: cfg.channel, text: msg,  as_user: true)

ytti:
Can you remove the diffenable too? And use ```if cfg.diff?``` instead of ```if diffenable```

ninoabbate:
@ytti sorry to ask here, but I didn't find a _complete_ info in README about the slackdiff hook: it is not clear what kind of token must be used. It would be great to have more info regarding the setup to be done on slack for it.

ytti:
bump

ytti:
bump

ytti:
bump

makzdot:
bump

laf:
Hey @natm. Any chance you can take a look at the couple of comments / changes needed above?

natm:
We don't have a use case for this internal and post diffs to our channels. If this functionality is still required then someone else will need to work on it, I'm short of time. Sorry.

laf:
@natm If you don't mind I'll keep this open for now, will see if I can spare some time on it.

laf:
Ok so the reason why this does contain code like:

```diff
 +        diffenable = true
 +        if cfg.has_key?('diff') == true
 +          if cfg.diff == false
 +            diffenable = false
 +          end
 +        end
```

Is because it then maintains backwards compatibility. If we just did `if cfg.diff == true` then old configs would fail.

I've tested this and it works so I'm going to merge in now the docs conflict is resolved.

Issue #364, opened 2016-03-21
Vyos support


sliddjur:
Hello, it would be nice to have vyos support in oxidized

ytti:
https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/vyatta.rb

Is it this?

ytti:
That may be misnamed, if the real name is vyos.

sliddjur:
E, [2016-03-21T14:42:40.066711 #11618] ERROR -- : node {:name=>"vyosrouter.example.com", :model=>"vyos", :group=>nil} raised Oxidized::ModelNotFound with message 'vyos not found for node 1.2.3.4'

Fed from librenms source.

Vyos and Vyatta are two different OSes

ytti:
Are you sure? What if you map to vyos to vyatta, that does not produce working results?

If so, if they are different, we can surely create model for it, but we'll need access to one of such device.

laf:
vyos is a fork of vyatta (I use neither) but I would have thought the vyatta model would work fine for vyos.

sliddjur:
@ytti thanks this worked
```
model_map:
  vyos: vyatta
```
@laf also edgeos model file is exactly the same as vyatta

thanks for the help

PR #765, opened 2017-03-14
Add support for VyOS

acgustafson:
VyOS is a fork of Vyatta however some applications treat the two devices differently. When Oxidized integrates with LibreNMS via the LibreNMS API, LibreNMS passes back the OS for VyOS devices as "vyos". This makes Oxidized not work properly with VyOS devices since oxidized is currently looking for the OS type "Vyatta".

This is tested on VyOS 1.17


laf:
You can just map vyos to vyatta https://github.com/ytti/oxidized/issues/364

acgustafson:
Can't believe I forgot about mapping. That would be a much easier solution. Do you think it would be worth submitting a new PR to change the default config to have that map?

laf:
Pass on that one, seems quite specific though.

PR #769, opened 2017-03-16
Ignore the host key checking

funzoneq:
@ytti this was blocking our backups. It's not pretty to do it this way, but I don't know a better way to fix it.


ytti:
We have config option for this 'secure', I think this should be only done if 'secure' is false.

ytti:
perhaps:
```
"ssh %s#{proxy_host -W %h:%p" % [secure ? "" : "-o blaablaa "]
```

or
```
proxy_command =  "ssh  "
proxy_command += "-o blaablaa " unless secure
proxy_command += "#{proxy_host} -W %h:%p
```

funzoneq:
I opted for the more readable last option.

Issue #772, opened 2017-03-16
Support for Cisco NGA3240


udhos:
Is there an existing model with support for Cisco NGA3240 ?

ytti:
I don't think so. Should be easy to produce, copy existing model to ~/.config/oxidized/model/X.rb
modify it until it works.

Helpful is to have session/input debugging turned on:
```
input:
  debug: true
```

Good luck.

PR #774, opened 2017-03-16
Support for Cisco NGA.

udhos:
Fixes https://github.com/ytti/oxidized/issues/772


ytti:
Are you sure IOS model would not have worked? Why do you comment the config out, makes the file  unusable directly as config?

udhos:
The IOS model issues 'show run' while NGA accepts only 'show configuration'.
Sorry, commented out the output by accident.

PR #777, opened 2017-03-21
Remove PreLogout Command

MajesticFalcon:


MajesticFalcon:
The logout command causes issues with older origin foundry ironware.

ytti:
What kinda issues? And does the new version work for all cases?

Perhaps ask the person who added the logout as well.

PR #779, opened 2017-03-22
Edit init script

MajesticFalcon:
Making it more obvious to users what needs to happen

### DIFF
--- a/extra/oxidized.init
+++ b/extra/oxidized.init
@@ -15,11 +15,11 @@
 set -e
 
 PATH=/sbin:/bin:/usr/sbin:/usr/bin
-DAEMON=PATH_TO_OXIDIZED_BIN
+DAEMON=/usr/local/bin/oxidized
 NAME="oxidized"
 DESC="Oxidized - Network Device Configuration Backup Tool"
 ARGS=""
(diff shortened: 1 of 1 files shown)


ytti:
From casual view, earlier to me is obvious what are local and what I need to change. The new looks like maybe they magically appear from somewhere.

I don't mind, just odd.  Perhaps we should assume user to be oxidized and assume binary to be in path and require configuration when that is not true?

MajesticFalcon:
That makes more sense. For some reason, my eyes didn't notice 'sts'. I assumed it would have been defaulted to oxidized.

ytti:
I would have just assume oxidized binary is in path and just use 'oxidized'. If it's not in path, have someone change it?

MajesticFalcon:
Yeah, I tried that first. For some reason, I get the following error. 'start-stop daemon unable to stat '//oxidized'' When I run start-stop-daemon manually with -exec oxidized, it does the same thing

ytti:
Maybe $(which oxidized)

MajesticFalcon:
$(which oxidized) starts the server

ytti:
I meant DAEMON=$(which oxidized)

MajesticFalcon:
That worked for me.

PR #783, opened 2017-03-24
Adding support for more "show version" commands

rfdrake:
This is a port of some of the rancid "show version" parsing.  I tried to
simplify it from the way rancid does it, so it doesn't attempt to guess
the device type, but it does provide IOS version and the type of
processor.


rfdrake:
Here is an example of the output from the changes:

```
! Cisco IOS Software, 2400 Software (C2430-I6K9O3S-M), Version 12.4(15)T8, RELEASE SOFTWARE (fc3)
!
! Image: slot0:c2430-i6k9o3s-mz.124-15.T8.bin
! Chassis type: IAD2430
! Memory: main 119808K/11264K
! Processor ID: FHK1208F09V
! CPU: R527x , R527x CPU at 225MHz, Implementation 40, Rev 3.1
!
```

PR #785, opened 2017-03-28
iosxe support

chifu1234:
copy of iosxr
tested and worked with
Cisco IOS XE Software, Version 03.16.04a.S - Extended Support Release
Cisco IOS Software, ASR1000 Software (X86_64_LINUX_IOSD-UNIVERSALK9-M), Version 15.5(3)S4a, RELEASE SOFTWARE (fc1)

### DIFF
--- a/lib/oxidized/model/iosxe.rb
+++ b/lib/oxidized/model/iosxe.rb
@@ -1,0 +1,47 @@
+class IOSXE < Oxidized::Model
+
+  # IOS XE model #
+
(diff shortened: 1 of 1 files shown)


ytti:
Why do we need IOS-XE model? The CLI is IOS CLI, just it's running on top of linux as a process, instead of embedding an OS too. 
From Oxidized POV, I don't understand it.

chifu1234:
Sorry yes  you are right! I can handle the probelm with:
model_map:
  iosxe: ios

PR #798, opened 2017-04-05
Update slackdiff.rb

sspans:
Add proxy option to slackdiff hook


ytti:
```cfg.proxy?``` return value or false. But this works too and may be more readable .

PR #801, opened 2017-04-06
Improve IOS secret handling

drbild:
- Fixed IOS model to redact only secrets
- Added IPSEC keys to IOS secret config


ytti:
Thanks!

PR #807, opened 2017-04-12
Add new output GitCrypt

clement-parisot:
The GIT backend (recommended) will initialize an empty GIT repository in the specified path and create a new commit on every configuration change. The GIT-Crypt backend will also initialize a GIT repository but every configuration push to it will be encrypted on the fly by using ```git-crypt``` tool.

This uses the gem git and system git-crypt interfaces. Have a look at [GIT-Crypt](https://www.agwa.name/projects/git-crypt/) documentation to know how to install it.
Additionally to user and email informations, you have to provide the users ID that can be a key ID, a full fingerprint, an email address, or anything else that uniquely identifies a public key to GPG (see "HOW TO SPECIFY A USER ID" in the gpg man page).

For a single repositories for all devices:

``` yaml
output:
  default: gitcrypt
  gitcrypt:
    user: Oxidized
    email: o@example.com
    repo: "/var/lib/oxidized/devices"
    users:
      - "0x0123456789ABCDEF"
      - "<user@example.com>"
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -33,6 +33,7 @@ PLATFORMS
 
 DEPENDENCIES
   bundler (~> 1.10)
+  git (~> 1)
   minitest (~> 5.8)
   mocha (~> 1.1)
   oxidized!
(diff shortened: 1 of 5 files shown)


ytti:
Only really care about the node.rb change. I would have left the ```else;return;end``` branch out. And allow it just to return implied nil at the end of the method.

But as it looks like it's not going to break anything and you seem to need it, merged.

I am worried about the maintainability, I won't expect people who make changes to ```git``` to make changes to this. We might need to create middleware so that we have only single common ```git``` output and crypto is just matter of config.

PR #814, opened 2017-04-17
Adding support for internal CA certs

mikenowak:
I have this specific use case for oxidized docker image.

We got a private github instance that uses a cert signed by our internal CA.

Of course the githubrepo hook fails, since the internal CA isn't trusted.

The simplest solution I could think of is this:
1) Utilise `update-ca-certs` that scans scans `/usr/local/share/ca-certificates/`
2) New environment variable `UPDATE_CA_CERTIFICATE` (similar to `CONFIG_RELOAD_INTERVAL`) to control whether update-ca-certificate should be executed.

Of course this means that the internal CA certs need to end up in the container somehow, but something among the lines of should work.

```
docker run -v /etc/oxidized:/root/.config/oxidized -v /path/to/MY-CA.crt:/usr/local/share/ca-certificates/MY-CA.crt -p 8888:8888/tcp -e UPDATE_CA_CERTIFICATES=true -t oxidized/oxidized:latest
```

I've also updated the README.md to reflect this

Let me know what you think.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN apt-get -y autoremove
 
 ADD extra/oxidized.runit /etc/service/oxidized/run
 ADD extra/auto-reload-config.runit /etc/service/auto-reload-config/run
+ADD extra/update-ca-certificates.runit /etc/service/update-ca-certificates/run
 
 VOLUME ["/root/.config/oxidized"]
 EXPOSE 8888/tcp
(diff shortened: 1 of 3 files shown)


ytti:
Educate me, why does the update certs need to sleep for infinity?

mikenowak:
I assumed (and haven't really tested so happy to stand corrected) that the lack of sleep (this sounds horrible) would cause the service to restart every time the update-ca-certificates command completes. So it would be a never ending loop of a sort.

This assumption is based on what you got in auto-reload-config.runit

```
if [ -z "$CONFIG_RELOAD_INTERVAL" ]; then
    # Just stop and do nothing
    sleep infinity
fi
```

ytti:
Right, but auto-reload-config is desirable to restart if it for some reason quits, right? Is this desirable to restart? I mean it's one-off, right? So maybe service isn't the best place anyhow?

But I'll let someone change this, if it bothers them.

PR #822, opened 2017-04-21
Create siklu.rb to add support for Siklu EtherHaul radios

bdg-robert:
I've tested this file on a Siklu EH-1200FX radio, however that's the only model I have, and it pulls the configuration without error.

I'm looking for feedback with grabbing the running-configuration versus the startup-configuration, though. The only difference in the output is that with the startup-configuration command you get a line at the beginning that tells you what version created the configuration file. I feel like this might be important to have, however I'd also like to know if changes were made to the running config that weren't saved.

Example of the line in startup-configuration output:
`
####====####  Generated by ver. 7.0.0.16739, convert ver. 12`

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,8 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
  * [pfSense](lib/oxidized/model/pfsense.rb)
  * Quanta
    * [Quanta / VxWorks 6.6 (1.1.0.8)](lib/oxidized/model/quantaos.rb)
+ * Siklu
+   * [EtherHaul](lib/oxidized/model/siklu.rb)
  * Supermicro
    * [Supermicro](lib/oxidized/model/supermicro.rb)
  * Trango Systems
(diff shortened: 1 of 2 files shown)


ytti:
Generally people want to take running-config, but your probably gonna need to wait quite while for feedback, I don't think this is very common device. So you're mostly going to have to make the model do what /you/ need it to do. 

Please also update README.md to include the new device and link to the model.

bdg-robert:
Would it be possible to list the startup-configuration first, extract only the line with the creating version, then grabbing the running-configuration for the rest?

ytti:
Yes that is possible. You gotta do your home work. Essentially what you're asking 'how do I extract one line from a string'.

bdg-robert:
Feels like this should cut it, but it isn't:

```
  cmd 'copy startup-configuration display' do |cfg|
    cfg.each_line.take_while { |line| line.match(/\#+\=+\#+.+/) }
  end

  cmd 'copy running-configuration display' do |cfg|
    cfg.each_line.to_a[1..-2].join
  end

  cfg :ssh do
    pre_logout 'exit'
  end

```

Any hints?

bdg-robert:
I played around with trying to match the version string with a regex until my eyes crossed, then I just went with cutting the top lines of the output instead. I feel like the better way to do it is with a regex, but I'm afraid I don't have the requisite ruby knowledge to do so.

PR #824, opened 2017-04-30
Ssh extraction

MajesticFalcon:


ytti:
Thanks! You've put lot of work into this, but I'm not currently in position to test it. Perhaps 
@ElvinEfendi or @danilopopeye

I've not read the extracted library code at all, but in the ssh.rb client using the library, at least ```#send``` is broken,  and ```#cmd``` appears no longer to support ```expect``` but rely on ```exec```working, which is true for minority of our models.
I feel like there is still duplication and role-overwrap, like ```exec``` and ```disconnect```

This was just short 5min review, sorry.

MajesticFalcon:
Thanks for your feedback. I'll work on it.

On May 7, 2017, at 5:54 AM, ytti <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks! You've put lot of work into this, but I'm not currently in position to test it. Perhaps
@ElvinEfendi<https://github.com/ElvinEfendi> or @danilopopeye<https://github.com/danilopopeye>

I've not read the extracted library code at all, but in the ssh.rb client using the library, at least #send is broken, and #cmd appears no longer to support expect but rely on execworking, which is true for minority of our models.
I feel like there is still duplication and role-overwrap, like exec and disconnect

This was just short 5min review, sorry.

ElvinEfendi:
Sorry, I don't have time to review either.

PR #874, opened 2017-05-31
interval measured in seconds. update readme

zeki893:
interval measured in seconds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -783,7 +783,7 @@ Below is an advanced example configuration. You will be able to (optionally) ove
 username: oxidized
 password: S3cr3tx
 model: junos
-#in seconds
+#interval in seconds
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -783,6 +783,7 @@ Below is an advanced example configuration. You will be able to (optionally) ove
 username: oxidized
 password: S3cr3tx
 model: junos
+#interval in seconds
```
ytti: You don't need two commits to add this. Also I'd prefer it to be in-line:

interval: 3600  # measured in seconds

I'm curious if 3600 was ambiguous. What other unit makes sense? milliseconds? minutes? hours?

laf:
Merged in with #1093

Issue #758, opened 2017-03-07
Trying to create new model Audiocodes Mediant


ecm75:
Hi,

me again, this time I'm trying to create a new model for Audiocodes Mediant.
I started by cloning a simple existing model and want to use only ssh.
When connecting to the node oxidize always runs into timeouts without producing any input debug...
I then enabled ssh debug in ssh.rb by adding ```:verbose => :debug``` to ssh_opts and I get this:
```D, [2017-03-07T13:49:39.945331 #17245] DEBUG -- net.ssh.transport.session[1b02d6c]: establishing connection to 192.168.7.17:22
D, [2017-03-07T13:49:39.947466 #17245] DEBUG -- net.ssh.transport.session[1b02d6c]: connection established
I, [2017-03-07T13:49:39.947788 #17245]  INFO -- net.ssh.transport.server_version[1aff978]: negotiating protocol version
D, [2017-03-07T13:49:40.006377 #17245] DEBUG -- net.ssh.transport.server_version[1aff978]: remote is `SSH-2.0-AudioCodes'
D, [2017-03-07T13:49:40.006533 #17245] DEBUG -- net.ssh.transport.server_version[1aff978]: local is `SSH-2.0-Ruby/Net::SSH_3.0.2 x86_64-linux-gnu'
D, [2017-03-07T13:50:46.017408 #17245] DEBUG -- socket[1b02790]: read 0 bytes
D, [2017-03-07T13:50:46.964922 #17245] DEBUG -- net.ssh.transport.session[1bb6ce0]: establishing connection to 192.168.7.17:22
D, [2017-03-07T13:50:46.967232 #17245] DEBUG -- net.ssh.transport.session[1bb6ce0]: connection established
I, [2017-03-07T13:50:46.967534 #17245]  INFO -- net.ssh.transport.server_version[1bb1664]: negotiating protocol version
D, [2017-03-07T13:50:47.007672 #17245] DEBUG -- net.ssh.transport.server_version[1bb1664]: remote is `SSH-2.0-AudioCodes'
D, [2017-03-07T13:50:47.007931 #17245] DEBUG -- net.ssh.transport.server_version[1bb1664]: local is `SSH-2.0-Ruby/Net::SSH_3.0.2 x86_64-linux-gnu'
```
It seems SSH doesn't even get past protocol negotiation.. any ideas?

ytti:
I don't know what it should print next, but what is the actual error you're getting, in logs or crashfile.

[...]

PR #889, opened 2017-06-12
Use newer net-ssh as default requirement

Fauli83:
For some devices oxidized has problems to connect using ssh with the current Version. If you install a newer Version of net-ssh (4.1.0) The problem is solved.

An Example is the following issue:
Trying to create new model Audiocodes Mediant #758 

I needed to modify ssh.rb because in newer versions of net-ssh it seems you cannot pass an nil value for Proxy.

I also added a new model AudioCodes, which will only work with the changes above.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,10 +1,10 @@
 PATH
   remote: .
   specs:
-    oxidized (0.18.0)
+    oxidized (0.20.0)
       asetus (~> 0.1)
       net-ssh (~> 3.0.2)
-      net-telnet
+      net-telnet (~> 0)
       rugged (~> 0.21, >= 0.21.4)
       slop (~> 3.5)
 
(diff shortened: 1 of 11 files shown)


lib/oxidized/input/ssh.rb
```diff
@@ -24,20 +24,23 @@ def connect node
...
+                :auth_methods => %w(none publickey password keyboard-interactive),
+                :number_of_password_prompts => 0,
+        }
+
       if proxy_host = vars(:ssh_proxy)
         proxy_command =  "ssh "
         proxy_command += "-o StrictHostKeyChecking=no " unless secure
         proxy_command += "#{proxy_host} -W %h:%p"
         proxy =  Net::SSH::Proxy::Command.new(proxy_command)
+        ssh_opts[:proxy] = proxy
```
ytti: Why change in same commit the way ssh proxy is done?

lib/oxidized/model/audiocodes.rb
```diff
@@ -0,0 +1,32 @@
+class AudioCodes < Oxidized::Model
```
ytti: This belongs in another commit.

lib/oxidized/input/ssh.rb
Fauli83: You can't pass an empty proxy in ssh_opts or you'll receive errors in the log. That's why I've changed it.

lib/oxidized/model/audiocodes.rb
Fauli83: you're right. removed it and created new request

ytti:
Looks good. Can you also squash these commits so this does not bring any audicodecs addition and deletion to the history. 

Thanks!

Fauli83:
New request with cleaner commits .> #909

PR #893, opened 2017-06-15
Update report git commits script

asenci:
Update report git commits script to use bare repo instead of working directory, dropping the need for a temporary directory and the clone/pull process.


ytti:
@nickhilliard please ack

nickhilliard:
looks good to me.

PR #894, opened 2017-06-15
Disable output paging on FortiOS

asenci:
Disable output paging on FortiOS, speeding up the execution and avoiding job timeouts when using SSH.

### DIFF
--- a/lib/oxidized/model/fortios.rb
+++ b/lib/oxidized/model/fortios.rb
@@ -55,6 +55,18 @@ cfg << cmd('end') if @vdom_enabled
   end
 
   cfg :telnet, :ssh do
+    post_login 'config global'
+    post_login 'config system console'
+    post_login 'set output standard'
+    post_login 'end'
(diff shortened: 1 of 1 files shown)


ytti:
We can't do this. We cannot change user devices config.

 1. We don't know if Oxidized has the permission
 2. We don't know the _current_ state, which means, you may end up _enabling_ paging for people who currently do disable it

We could potentially guard this under some var like fortios_disable_pager, but we cannot do this unconditionally.

danilopopeye:
@asenci could you rework this in a new PR with the `vars(:fortios_disable_pager)` like @ytti suggested? In the meantime I'm going to close this PR, ok? Thanks for the time and code, but I'm with @ytti for not changing the device without the user explicit knowing.

asenci:
Hi @danilopopeye, agreed. Will work on implementing the var and detecting the current state for rolling back after the backup as soon as I have some spare time.

PR #895, opened 2017-06-16
net-ssh 4.1

crami:
My try to get to the newest net-ssh version, as it supports more cyphers Im able to use ssh on more nodes. I tested my changes on  a few devices and that looks good. What is untested is the ssh proxy feature. Is someone able to verify this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -477,7 +477,7 @@ source:
     map:
       name: ip
       model: model
-      username: username
+      user: username
       password: password
     vars_map:
       enable: enable
--- a/lib/oxidized/input/ssh.rb
+++ b/lib/oxidized/input/ssh.rb
@@ -24,20 +24,22 @@ module Oxidized
       secure = Oxidized.config.input.ssh.secure
       @log = File.open(Oxidized::Config::Log + "/#{@node.ip}-ssh", 'w') if Oxidized.config.input.debug?
       port = vars(:ssh_port) || 22
-      if proxy_host = vars(:ssh_proxy)
-        proxy_command =  "ssh "
-        proxy_command += "-o StrictHostKeyChecking=no " unless secure
-        proxy_command += "#{proxy_host} -W %h:%p"
-        proxy =  Net::SSH::Proxy::Command.new(proxy_command)
-      end
       ssh_opts = {
         :port => port.to_i,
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -508,7 +508,7 @@ source:
     map:
       name: ip
       model: model
-      username: username
+      user: username
```
ytti: Why this? node.rb uses ```username```

lib/oxidized/input/ssh.rb
```diff
@@ -24,20 +24,22 @@ def connect node
...
       }
+
+      if proxy_host = vars(:ssh_proxy)
+        proxy_command =  "ssh "
+        proxy_command += "-o StrictHostKeyChecking=no " unless secure
+        proxy_command += "#{proxy_host} -W %h:%p"
+        proxy =  Net::SSH::Proxy::Command.new(proxy_command)
+        ssh_opts[:proxy] = proxy
+      end
+
```
ytti: Why did you change this block from the current? Both are functional, seems unnecessary.

lib/oxidized/model/ironware.rb
```diff
@@ -67,7 +67,7 @@ class IronWare < Oxidized::Model
     # match expected prompts on both older and newer
     # versions of IronWare
     username /^(Please Enter Login Name|Username):/
-    password /^(Please Enter Password |Password):/
+    password /^(Please Enter Password ?|Password):/
```
ytti: This really should be in another PR.

README.md
crami: Your are righ, changed it at the wrong location ... sql uses user not username...

lib/oxidized/model/ironware.rb
crami: Right ... can you tell me how to move that to the other pull request ?

lib/oxidized/input/ssh.rb
crami: No, they are not the same. In old version :proxy gets no value, which the ssh module complains about, in my version :proxy only gets set if there is need for a proxy.

ytti:
I have another pull request for net-ssh 4.1 So maybe just scratch this PR, and create another with just the model change. 
Very helpful that you've arrived to same conclusion about the ssh_proxy!

Thank you.

PR #896, opened 2017-06-19
Create esa.rb Email Security Appliance CISCO

cd67-usrt:


ytti:
You will also need to update README.md. If the operating system is called ESA, the name is fine. We should call the model as the operating system always. If we can't figure out what the operating system is called, I think CiscoESA might be safer name.

ytti:
Please update README.md too about the new model, thanks!

PR #897, opened 2017-06-19
Create sma.rb  Security Management Appliance CISCO

cd67-usrt:


lib/oxidized/model/sma.rb
```diff
@@ -0,0 +1,45 @@
+class SMA < Oxidized::Model
+
+	# ESA prompt "mail.example.com> "
+	prompt /^\r*([-\w. ]+\.[-\w. ]+\.[-\w. ]+[#>]\s+)$/
```
casdr: Maybe `\r*([-\.\w]+[#>]\s+)$` is better

ytti:
You need to update README.md too about new models. Is the OS called SMA? If not and you can't find the OS name. Maybe call the model CiscoSMA to avoid stealing potentially 'valuable' TLA, which might be name of an actual OS.

lib/oxidized/model/sma.rb
cd67-usrt: capture
No.
match on config file

ytti:
you need to also update README.md, to include the new model. Thanks!

lib/oxidized/model/sma.rb
casdr: Whoops, sorry :)

PR #906, opened 2017-07-01
Create audiocodes.rb for AudioCodes devices

Fauli83:
Please add this device config to the default repo.

This depends on ssh 4.1

Pull config from AudioCodes mediant devices >Version 7

PR #910, opened 2017-07-02
Create audiocodes.rb for AudioCodes devices

Fauli83:
Follow up of #906 with cleaner commits.


ytti:
Thanks. Bit surprised about the SSH, I guess it's another 'noauth' device, which always succeeds SSH auth, then presents prompt.
Also we can define many telnet/ssh blocks, so the shared part could be in common ```:ssh, :telnet``` block. But as it's single line, I guess repetition is ok.

PR #914, opened 2017-07-04
Add documentation for Huawei VRP devices

tuxis-ie:


ytti:
I guess this could be comment in the model itself. I wonder if VRP supports 'views' like IOS, where you don't need to give any specific privilege level, but can iterate exactly what command are permissible. 

Anyhow merged.

tuxis-ie:
Documentation suggests that something like that could be possible. But me nor the TAC made that work.. I also thought about putting the documentation in the model, but I figured this would be cleaner and more obivous to users.

PR #923, opened 2017-07-06
Fixes for Netgear

### LINKED ISSUE #920: Problems with Netgear GSM7352 and telnet
Trying to connect to GSM7352 over telnet. Sample manual session:

~~~
# telnet wrn-coresw1
Trying 192.168.0.3...
Escape character is '^]'.

(wrn-coresw1)
User:admin
Password:**********
(wrn-coresw1) >ena

(wrn-coresw1) #
~~~

Problem 1: oxidized doesn't recognise the "User:" prompt.  It looks like this is not configurable, so I patched it:

~~~
--- oxidized-0.20.0/lib/oxidized/model/netgear.rb.orig	2017-07-06 10:22:47.528428313 +0000
+++ oxidized-0.20.0/lib/oxidized/model/netgear.rb	2017-07-06 12:38:55.978924414 +0000
@@ -8,6 +8,10 @@
     cfg
   end

+  cfg :telnet do
+    username /^User:/
+  end
+
   cfg :telnet, :ssh do
     if vars :enable
       post_login do
~~~

Problem 2: I need oxidized to type "enable", but not to enter an enable password.  In the netgear.rb source, it looks like this should be possible by setting the "enable" var to a (ruby) "True".

It took some head-scratching as to how to do this, but I eventually worked out [from the source](https://github.com/ytti/oxidized/blob/master/lib/oxidized/config/vars.rb) that I could do this:

~~~
models:
  gsm7352:
    vars:
      enable: true
model_map:
  gsm7352: netgear
~~~

I had previously tried mapping "enable" to a csv column, but that didn't work (presumbly "true" was treated as a String)

Problem 3: it still couldn't login. The debug log file showed it hanging at this point:

~~~
User:admin
Password:**********
(wrn-coresw1) >enable

(wrn-coresw1) #
~~~

It looks like the code is wrong; it should be using "send" rather than "cmd". The following change makes it login successfully:

~~~
   cfg :telnet, :ssh do
     if vars :enable
       post_login do
-        cmd 'enable'
+        send "enable\n"
         # Interpret enable: true as meaning we won't be prompted for a password
         unless vars(:enable).is_a? TrueClass
           expect /[pP]assword:\s?$/
-          cmd vars(:enable) + "\n"
+          send vars(:enable) + "\n"
         end
         expect /^.+[#]$/
       end
~~~

Problem 4: it now gets as far as dumping the config.  But it fails when logging out:

~~~
D, [2017-07-06T13:00:58.854397 #11695] DEBUG -- : lib/oxidized/input/cli.rb Running pre_logout commands at wrn-coresw1.example.net
D, [2017-07-06T13:00:58.854495 #11695] DEBUG -- : Telnet: exit @wrn-coresw1.example.net
...
W, [2017-07-06T13:01:18.941648 #11695]  WARN -- : 192.168.0.3 raised Timeout::Error with msg "timed out while waiting for more data"
~~~

Debug log file:

~~~
...
router rip
exit
router ospf
exit
exit
terminal length 0

(wrn-coresw1) #exit

(wrn-coresw1) >
~~~

So for some reason it doesn't accept the returned prompt, even though it appears to match the prompt regexp in netgear.rb:

~~~
prompt /^(\([\w-]+\)\s[#>])$/
~~~

Now, the netgear.rb code has a pre_logout of both "exit" and "quit". Actually this is not necessary; you can "quit" directly from enable mode.  But then this has another trap for the unwary:

~~~
(wrn-coresw1) >logout

The system has unsaved changes.
Would you like to save them now? (y/n)
~~~

So I think it is safer simply to disconnect, which can be done by getting rid of the two `pre_logout` statements.

Here is the final patch which makes this particular Netgear switch work for me:

~~~
--- oxidized-0.20.0/lib/oxidized/model/netgear.rb.orig	2017-07-06 10:22:47.528428313 +0000
+++ oxidized-0.20.0/lib/oxidized/model/netgear.rb	2017-07-06 13:11:11.569577765 +0000
@@ -8,21 +8,23 @@
     cfg
   end

+  cfg :telnet do
+    username /^User:/
+  end
+
   cfg :telnet, :ssh do
     if vars :enable
       post_login do
-        cmd 'enable'
+        send "enable\n"
         # Interpret enable: true as meaning we won't be prompted for a password
         unless vars(:enable).is_a? TrueClass
           expect /[pP]assword:\s?$/
-          cmd vars(:enable) + "\n"
+          send vars(:enable) + "\n"
         end
         expect /^.+[#]$/
       end
     end
-    pre_logout 'exit'
-    pre_logout 'quit'
   end

   cmd 'show running-config' do |cfg|
~~~

candlerb:
- match the User: prompt on telnet
- fix the sending of enable password
- don't issue exit/quit commands on logout, because an
  additional prompt may or may not be encountered

Closes #920

### DIFF
--- a/lib/oxidized/model/netgear.rb
+++ b/lib/oxidized/model/netgear.rb
@@ -8,21 +8,29 @@ class Netgear < Oxidized::Model
     cfg
   end
 
+  cfg :telnet do
+    username /^User:/
+  end
+
   cfg :telnet, :ssh do
     if vars :enable
       post_login do
-        cmd 'enable'
+        send "enable\n"
         # Interpret enable: true as meaning we won't be prompted for a password
         unless vars(:enable).is_a? TrueClass
           expect /[pP]assword:\s?$/
-          cmd vars(:enable) + "\n"
+          send vars(:enable) + "\n"
         end
         expect /^.+[#]$/
       end
     end
     post_login 'terminal length 0'
-    pre_logout 'exit'
-    pre_logout 'quit'
+    # quit / logout will sometimes prompt the user:
+    #
+    #     The system has unsaved changes.
+    #     Would you like to save them now? (y/n)
+    #
+    # So it is safer simply to disconnect and not issue a pre_logout command
   end
 
   cmd 'show running-config' do |cfg|


ytti:
Probably cleaner would be something like:

```
post_login do
  if vars(:enable) == true
    cmd "enable"
  elsif vars(:enable)
    send "enable\n"
    cmd vars(:enable)
  end
end
```

candlerb:
> Probably cleaner would be something like

Agreed - I was just trying to make minimal change to existing code.

Issue #913, opened 2017-07-04
Cannot backup large configuration Cisco UCS


WiXZlo:
Hello.
I have UCS Blade server, UCS Manager based on NXOS with different commands. I created new model:
```
class NXOS_Cisco_UCS < Oxidized::Model

  prompt /^(\r?[\w.@_()-]+[#]\s?)$/
  comment '! '

  cmd 'show version brief' do |cfg|
    cfg = cfg.each_line.take_while { |line| not line.match(/uptime/i) }
    comment cfg.join ""
  end

  cmd 'show chassis detail' do |cfg|
    comment cfg
  end

  cmd 'show fabric-interconnect detail' do |cfg|
    comment cfg
  end

  cmd 'show configuration all | no-more' do |cfg|
    cfg.gsub! /^!Time:[^\n]*\n/, ''
  end

  cfg :ssh, :telnet do
    post_login 'terminal length 0'
    pre_logout 'exit'
  end

  cfg :telnet do
    username /^login:/
    password /^Password:/
  end
end
```
I see in backup only output of inventory's commands but output of command 'show configuration all | no-more' is empty.
There are no errors in log file:
```
I, [2017-07-04T12:07:27.682893 #4711]  INFO -- : lib/oxidized/nodes.rb: Loaded 1 nodes
D, [2017-07-04T12:07:28.215668 #4711] DEBUG -- : lib/oxidized/core.rb: Starting the worker...
D, [2017-07-04T12:07:28.215999 #4711] DEBUG -- : lib/oxidized/worker.rb: Jobs 0, Want: 1
D, [2017-07-04T12:07:28.216295 #4711] DEBUG -- : lib/oxidized/worker.rb: Added ucs1 to the job queue
D, [2017-07-04T12:07:28.216435 #4711] DEBUG -- : lib/oxidized/worker.rb: 1 jobs running in parallel
D, [2017-07-04T12:07:28.216658 #4711] DEBUG -- : lib/oxidized/job.rb: Starting fetching process for ucs1 at 2017-07-04 09:07:28 UTC
D, [2017-07-04T12:07:28.217192 #4711] DEBUG -- : lib/oxidized/input/ssh.rb: Connecting to ucs1
D, [2017-07-04T12:07:28.531855 #4711] DEBUG -- : lib/oxidized/input/ssh.rb: expecting [/^(\r?[\w.@_()-]+[#]\s?)$/] at ucs1
[...]

PR #943, opened 2017-07-24
Add support for Cisco UCS model

WiXZlo:
Added support for Cisco UCS
#913


waza-ari:
Thank you for writing this plugin! +1 for this

danilopopeye:
@WiXZlo could you also include the device in the README?

PR #950, opened 2017-07-28
added Fire Linux OS (FirePOWER) and SonicOS (Sonicwall) support

rgnv:


lib/oxidized/model/aireos.rb
```diff
@@ -45,6 +45,7 @@ def clean cfg
     cfg.each_line do |line|
       next if line.match /^\s*$/
       next if line.match /rogue (adhoc|client) (alert|Unknown) [\da-f]{2}:/
+      next if line.match /interface nat-address management set \d+.\d+.\d+.\d+/
```
danilopopeye: not sure this belong to this PR. this was intended?

lib/oxidized/model/sonicos.rb
```diff
@@ -0,0 +1,47 @@
...
+    cfg.gsub! /^: [^\n]*\n/, ''
+    clean cfg
+  end
+
+  cfg :telnet, :ssh do
+    post_login 'no cli pager session'
+    pre_logout 'exit'
+  end
+
+  def clean cfg
```
danilopopeye:
```ruby
def clean cfg
  cfg.map do |line|
    next if line.match /system-time \"\d{2}\/\d{2}\/\d{4} \d{2}\:\d{2}:\d{2}.\d+\"/
    next if line.match /system-uptime "(\s+up\s+\d+\s+)|(.*Days.*)"/
    next if line.match /checksum \d+/
    line.strip
  end.compact.join("\n") << "\n"
end
```

wdyt? 😃

lib/oxidized/model/sonicos.rb
```diff
@@ -0,0 +1,47 @@
...
+    out = []
+    cfg.each_line do |line|
+      next if line.match /system-time \"\d{2}\/\d{2}\/\d{4} \d{2}\:\d{2}:\d{2}.\d+\"/
+      next if line.match /system-uptime "(\s+up\s+\d+\s+)|(.*Days.*)"/
+      next if line.match /checksum \d+/
+      line = line[1..-1] if line[0] == "\r"
```
danilopopeye: the `#strip` in the next line would remove the `\r`, so we can remove this line.

```ruby
"\tgoodbye\r\n".strip   #=> "goodbye"
```

rgnv:
Wrong branch pushed out for pull request.

PR #729, opened 2017-02-16
New Slack hook - posts colourized diffs to a channel

natm:
This contribution includes a native [Slack](http://slack.com) hook which can be used to post colourized diffs to a Slack channel of your choice.

screen shot 2017-02-16 at 16 35 16

When expanded:

screen shot 2017-02-16 at 16 35 27

Configurable as follows:

```yaml
hooks:
  slack:
    type: slackdiff
    events: [post_store]
    token: #########################
    channel: "#oxidized-prod"
```

PR #951, opened 2017-07-29
Added hook for XMPP MUC

mortzu:
We're using XMPP instead of Slack. So I've rewritten the hook for XMPP

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ RUN gem install oxidized-web --no-ri --no-rdoc
 # dependencies for hooks
 RUN gem install aws-sdk
 RUN gem install slack-api
+RUN gem install xmpp4r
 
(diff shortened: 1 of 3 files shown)


danilopopeye:
@mortzu could you update the `README.md` and `Dockerfile` like the #729 PR ? Other than that, the code seem 👌

mortzu:
Updated README.md and Dockerfile and added some changes we've made in the meantime

mortzu:
Updated again to match new docs refactoring

mortzu:
Please merge it?

laf:
If you can provide a way to test this with a publicly accessible xmpp service then I'll try it out and merge in.

mortzu:
You can register an account at https://www.jabber.org/

laf:
jabber.at got me there :)

PR #952, opened 2017-07-29
Create enterasys.rb (C3/B3 series switches)

koenvdheuvel:
Used for the lanparty Frag-o-Matic

Tested with the following switches:
- Enterasys Securestack C3G124-48
- Enterasys Securestack B3G124-48

This adds basic support. We will try and maintain this file if we find valuable additions in the future.


danilopopeye:
@koenvdheuvel could you include the device in the README also?

laf:
@danilopopeye I've updated the readme so if you're happy I guess we can merge in.

PR #956, opened 2017-08-01
Driver for Westell 8178G and Westell 8266G devices

ignaqui:
We use Westell devices in our network (Westell 8178G and Westell 8266G), so I wrote a driver.
SSH daemon is broken on those devices, so only telnet-based shell works.

### DIFF
--- a/lib/oxidized/model/weos.rb
+++ b/lib/oxidized/model/weos.rb
@@ -1,0 +1,22 @@
+class WEOS < Oxidized::Model
+
+  #Westell WEOS, works with Westell 8178G, Westell 8266G
+
+  prompt /^(\s[\w.@-]+[#>]\s?)$/
+
+  cmd :all do |cfg|
+    cfg.each_line.to_a[1..-2].join
+  end
+
+  cmd 'show running-config' do |cfg|
+    cfg
+  end
+
+  cfg :telnet do
+    username /login:/
+    password /assword:/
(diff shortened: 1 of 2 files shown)


lib/oxidized/model/weos.rb
```diff
@@ -0,0 +1,22 @@
...
+    cfg.each_line.to_a[1..-2].join
+  end
+
+  cmd 'show running-config' do |cfg|
+    cfg
+  end
+
+  cfg :telnet do
+    username /login:/
+    password /assword:/
```
danilopopeye: why the missing `p` here? 😕

danilopopeye:
@ignaqui could you include the device in the README also?

lib/oxidized/model/weos.rb
ignaqui: It's not missing, it could be `P` or `p`. This regex greps both occurrences

ignaqui: > there is no `version` command in this device?
Honestly, I never used this command on the device, so I don't know

danilopopeye: you can use `/password:/i` then. 😉 
I gather these devices have both prompts then?

ignaqui:
Exactly, but I can't recollect which has what prompt. It's been awhile when
I worked on that project

PR #965, opened 2017-08-08
Added individual repository per device option for git output

bheum:
Additional to the existing options of a single repository for all
devices or one repository per group, this option allows to have one
repository for every individual device.

This comes in handy as i addressed this issue:
https://github.com/ytti/oxidized-web/issues/115

Large repositories slow down the process of finding the last configuration change, because the walker has to iterate through a lot of updates from other devices.

### DIFF
--- a/README.md
+++ b/README.md
@@ -607,7 +607,7 @@ output:
     repo: "/var/lib/oxidized/devices.git"
 ```
 
-And for groups repositories:
+For a single repositories per group:
 
 ``` yaml
 output:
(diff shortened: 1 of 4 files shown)


danilopopeye:
@bheum not sure if this is the best solution for the problem you mentioned. seems like the rugged::walker code need some optimization instead. maybe the group options could solve the slowness. wdyt?

laf:
No response to the questions asked above, closing this PR. @bheum feel free to respond and we can re-open this.

Issue #867, opened 2017-05-24
Switch Blade Nortel/HP (IBM BladeCenter and old HP Blade) only password in telnet


bonigeek:
I'm create a model for switch blade, for olds Nortel/HP equipments, but don't work. 

This is the model 
```
class Blade < Oxidized::Model
# Switches Blade HP e Nortel

  prompt /^[>> ]*[\w]*[#]\s?$/
  comment '# '

  cmd '/info/sys/dump' do |cfg|
    comment cfg
  end

  cmd '/cfg/dump'

  cfg :telnet do
    password /^(Enter password:)$/
    pre_logout 'exit'
  end

end
```
This is the debug

```
Enter password: Trying 10.10.45.9...
Connected to 10.10.45.9.


Enter password: Trying 10.10.45.9...
Connected to 10.10.45.9.


Enter password: Trying 10.10.45.9...
Connected to 10.10.45.9.


Enter password: Trying 10.10.45.9...
Connected to 10.10.45.9.


Enter password: Trying 10.10.45.9...
Connected to 10.10.45.9.


Enter password:
```

And the line in router.db

`10.10.45.9:blade:GROUP:password:password`

The device login:

```
telnet 10.10.45.9

Enter password:
System Information at  6:37:40 Sun Jul  1, 2073

Layer 2-7 Gigabit Ethernet Switch Module for IBM eServer BladeCenter

Switch is up 283 days, 1 hour, 40 minutes and 56 seconds.
Last boot:  0:00:58 Thu Jan  1, 2070 (power cycle)

MAC Address: xx:xx:xx:xx:xx:xx    Management IP Address (if 250): 10.10.45.9

Note - When the measured temperature inside the switch EXCEEDs
       the high threshold at 62 degree Celsius a syslog message
       will be generated.

Software Version 21.0.1.1 (FLASH image1), active configuration.

PCBA Part Number:      XXXXXXX-A
FAB Number:            XXXXXXXXXX
Serial Number:         XXXXXXXXXX
Manufacturing Date:    XXX
Hardware Revision:     X
PLD Firmware Version:  3.6
FP date (version):     2004/08/03 (18)

Temperature Sensor 1 (Warning):   44.0 C (Warn at 75.0 C/Recover at 70.0 C)
Temperature Sensor 2 (Shutdown):  48.0 C (Warn at 90.0 C/Recover at 80.0 C)

------------------------------------------------------------
[Main Menu]
     info     - Information Menu
     stats    - Statistics Menu
[...]

PR #974, opened 2017-08-16
IBM/Blade Networks/Lenovo Support

jackson-tim:


lib/oxidized/model/ibm.rb
```diff
@@ -0,0 +1,64 @@
...
+    cfg.gsub! /^(\s+(?:password|secret)) (?:\d )?\S+/, '\\1 <secret hidden>'
+    cfg.gsub! /^(tacacs-server \S+host \S+ ekey) (\S+)/, '\\1 <secret hidden>'
+    cfg
+  end
+
+  cmd 'show version' do |cfg|
+    comments = []
+    comments << cfg.lines.first
+    lines = cfg.lines
+    lines.each_with_index do |line,i|
```
danilopopeye: why not use `#each`? the index doesn't look like is needed here

lib/oxidized/model/ibm.rb
```diff
@@ -0,0 +1,64 @@
...
+        if line.match /^Software Version\s+(.+)\s+\(FLASH (.+)\).*/
+            comments << "Software: #{$1}, Booted from: #{$2}"
+        end
+
+        if line.match /^Boot kernel version (.+)/
```
danilopopeye: you should join all these `if` so we don't need to test three times each line 😉

danilopopeye: now that I come to think about it, why not just comment the whole command output?

laf:
@jackson-tim Any feedback on the requested changes for this?

zombah:
This looks like same as in #867 
Regarding IBM site correct name for OS is BladeOS

mortzu:
Can you please fix the conflict and add your change to CHANGELOG

Issue #866, opened 2017-05-24
Setting git output's single_repo option to true breaks version output if no groups


Merlin83b:
Similar to #413, if you set the `single_repo` option to `true` in a configuration not using groups, trying to look at a particular version of a device configuration results in the "Version not found" error.

The option should probably be silently ignored in this case, since the output module would use a single git repository anyway.

ytti:
This is probably oxidized-web issue, if I understood correctly.

fractalbrain:
Hi!  I'm having the same issue (I'm using github).  My workaround was to assign everything a "default" group, keep the "single_repo: true", and restart oxidized.

shepherdjay:
This appears to be addressed in an old pull request #979 - not sure if that fix can be manually applied while the pull request is reviewed and pending as I'm not familiar with Ruby.

PR #979, opened 2017-08-21
Fix version output (single git repo without group case)

Tibshot:
When trying to get backup versions in single repo without groups, I was getting 'Versions not found'.


ytti:
This is probably mistake in somewhere else in the code. I think we should guarantee group to be nil or false if we don't have group. Why is it empty string? Can we fix it in the earlier point?

shepherdjay:
Related to #866

mortzu:
Any news on this?

Issue #760, opened 2017-03-07
enhancement for "oxidized.service" file


zxyz:
on my system oxidized startup with systemd fails after a reboot because the network is not ready. 
Adding `After=network-online.target` in the `[Unit]` section solved this for me

```
[Unit]
Description=Oxidized - Network Device Configuration Backup Tool
After=network-online.target

[Service]
ExecStart=/usr/local/bin/oxidized
User=oxidized

[Install]
WantedBy=multi-user.target
```

ytti:
Send pull request, thanks!

davama:
@zxyz  Id noticed that too

https://github.com/ytti/oxidized/pull/921

laf:
@davama Why have default being optional? Should it not be on by default?

davama:
placed it that way because everyones env is diff. so you can set or not. up to you
if you want to change that PR then resubmit. it's the same to me

laf:
I would have thought though network services would be a requirement no matter what though :)

davama:
agreed
when i first installed i never had it. only later in life of our system did i have to add them. the lines are there so thats a start

danilopopeye:
@laf + @davama we can close this one, right?

laf:
@danilopopeye I don't see why not, I think it would be better to have it enabled as network services seems a key requirement to oxidized to start.

If others agree I'll open a PR.

danilopopeye:
> If others agree I'll open a PR.

fine by me 😃 closing this one

PR #980, opened 2017-08-21
Updated oxidized.service file to rely on networking

laf:
Follow on from #760

Oxidized relies on network connectivity so updated the oxidized.service file to reflect this + small doc update within the same file.


laf:
Thanks for the merge :)

PR #987, opened 2017-08-23
Added some API documentation

### LINKED ISSUE #940: Additional hooks for API calls
I'm proposing an additional `post_store` hook, if the configuration pull was initialized through an API call (`/node/next/[node_name]`).

The scenario I'm thinking of is this: 
Oxidized has a scheduled run once a day (at night for example). Changes found during these scheduled runs should trigger a notification script through the `post_store` hook and inform us about undocumented changes (it could be a security issue).

Now, during work hours, configuration changes are made by an admin and an API call is issued to Oxidized after logout from the switch (`/node/next/...`) . This change should not trigger the `post_store` hook, because this change is wanted and will be documented. So an additional `post_store` hook is needed which only applies to API calls. Also, the post_store hook needs to know it was not triggered by the API.

Is this something others would want? I'm trying to figure out how this could be done in the code, but I'm fairly new to Ruby, so I can't make any promises for a PR 🙁 Trying my best, though!

The possibility to edit the commit message while doing the API call would be nice (maybe by using a HTTP header?), but that's just wishful thinking.

mrplow87:
Added restful API documentation and fixed the mention of commit user and message in the introduction.

closes #940

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Oxidized is a network device configuration backup tool. It's a RANCID replacemen
 * automatically adds/removes threads to meet configured retrieval interval
 * restful API to move node immediately to head-of-queue (GET/POST /node/next/[NODE])
   * syslog udp+file example to catch config change event (ios/junos) and trigger config fetch
-  * will signal ios/junos user who made change, which output modules can use (via POST)
+  * optional mention of commit user and message (via PUT)
   * The git output module uses this info - 'git blame' will for each line show who made the change and when
 * restful API to reload list of nodes (GET /reload)
 * restful API to fetch configurations (/node/fetch/[NODE] or /node/fetch/group/[NODE])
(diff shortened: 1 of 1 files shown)


laf:
This is definitely needed but imho I think we should start splitting the readme out, this is prime to be put in a new file.

@ytti  / @danilopopeye - Thoughts?

mrplow87:
Also, if we split it into additional files, shouldn't this be moved to oxidized-web?

laf:
I guess although I think most people end up here.

danilopopeye:
> This is definitely needed but imho I think we should start splitting the readme out, this is prime to be put in a new file.

yeah, I was thinking about this yesterday. haven't decided how to do it yet.



> Also, if we split it into additional files, shouldn't this be moved to oxidized-web?

i guess we could move this doc to oxidized-web and include a link in the readme here?

laf:
> i guess we could move this doc to oxidized-web and include a link in the readme here?

Seems sensible.

danilopopeye:
@mrplow87 could you reopen this changes to oxidized-web? Then we can link it here.

PR #989, opened 2017-08-23
More than one output

Tibshot:
Tested with:

`default: git, file`

Git output is displayed in WebUI, conf are saved with git + file

yaml conf:

```
output:
  default: git, file
  git:
    single_repo: true
    user: oxidized
    email: config@example.com
    repo: /var/lib/oxidized/devices.git
    debug: true
  file:
    directory: /var/lib/oxidized/files
```

### DIFF
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -149,11 +149,13 @@ module Oxidized
     end
 
     def resolve_output opt
-      output = resolve_key :output, opt, Oxidized.config.output.default
-      if not Oxidized.mgr.output[output]
-        Oxidized.mgr.add_output output or raise MethodNotFound, "#{output} not found for node #{ip}"
+      outputs = resolve_key :output, opt, Oxidized.config.output.default
+      outputs.split(/\s*,\s*/).map do |output|
+        if not Oxidized.mgr.output[output]
+          Oxidized.mgr.add_output output or raise MethodNotFound, "#{output} not found for node #{ip}"
+        end
+        Oxidized.mgr.output[output]
       end
-      Oxidized.mgr.output[output]
     end
 
     def resolve_model opt
@@ -231,11 +233,13 @@ module Oxidized
     end
 
     def is_git? opt
-      (opt[:output] || Oxidized.config.output.default) == 'git'
+      outputs = (opt[:output] || Oxidized.config.output.default)
+      outputs.split(/\s*,\s*/).map.include?('git')
     end
 
     def is_gitcrypt? opt
-      (opt[:output] || Oxidized.config.output.default) == 'gitcrypt'
+      outputs = (opt[:output] || Oxidized.config.output.default)
+      outputs.split(/\s*,\s*/).map.include?('gitcrypt')
     end
 
   end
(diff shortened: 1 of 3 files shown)


lib/oxidized/node.rb
```diff
@@ -149,11 +149,13 @@ def resolve_input opt
     end
 
     def resolve_output opt
-      output = resolve_key :output, opt, Oxidized.config.output.default
-      if not Oxidized.mgr.output[output]
-        Oxidized.mgr.add_output output or raise MethodNotFound, "#{output} not found for node #{ip}"
+      outputs = resolve_key :output, opt, Oxidized.config.output.default
+      outputs.split(/\s*,\s*/).map do |output|
+        if not Oxidized.mgr.output[output]
```
danilopopeye:
``` ruby
unless Oxidized.mgr.output[output]
```

💅

lib/oxidized/node.rb
```diff
@@ -231,11 +233,13 @@ def resolve_key key, opt, global=nil
     end
 
     def is_git? opt
-      (opt[:output] || Oxidized.config.output.default) == 'git'
+      outputs = (opt[:output] || Oxidized.config.output.default)
+      outputs.split(/\s*,\s*/).map.include?('git')
```
danilopopeye: this `#split` is repeated several times. could you extract it to a private method? something like:

``` ruby
def output_list
  @output_list ||= outputs.split(/\s*,\s*/)
end
```

lib/oxidized/worker.rb
```diff
@@ -44,13 +44,15 @@ def process job
...
         msg += " with message '#{node.msg}'" if node.msg
-        output = node.output.new
-        if output.store node.name, job.config,
-                              :msg => msg, :user => node.user, :group => node.group
-          Oxidized.logger.info "Configuration updated for #{node.group}/#{node.name}"
-          Oxidized.Hooks.handle :post_store, :node => node,
-                                             :job => job,
-                                             :commitref => output.commitref
+        node.output.each do |output|
+          output = node.output.new
```
danilopopeye: not really sure this code do what you want. 😕 

you receive a `output` variable from the block `#each` but you shadow it here. if I guess I think you mean something like:

``` ruby
node.output.each do |output|
  current_output = output.new
  // ...
end
```

danilopopeye:
there's also a point I don't know what we should do here.

In the `#process` method since we will have more the one output, the hook `:post_store` will be invoked for each output configured. But we don't pass any reference as from which output it's being invoked from. Maybe we don't want to fire several `:post_store` for each node, I'm not convinced this is the best path forward.

wdyt? /cc @ytti @laf @aakso @ElvinEfendi

laf:
@danilopopeye Sorry I'm not a Ruby dev so this one is out of my ability to check :/

ytti:
I think we should have this feature, but I accept the plugin issue needs to be answered, and I don't have answer right now.

mortzu:
Any news on this?

ytti:
> Any news on this?

I have no plans at least to work on refactoring plugin for output model awareness. Considering dumping flat files from github is oneliner you. can run from crontab seems odd  that the only reason people want this, is for git+flat file, communicating that likely whole feature request is operator mistake.

mortzu:
Then I'll close this issue, ok?

PR #995, opened 2017-08-25
Add support for Alcatel Lucent (Nokia) SR-OS devices.

ZacharyPuls:
Tested on ALU (Nokia) 7750 routers.

PR #1004, opened 2017-08-30
Update README to indicate that Alcatel Lucent routers running SR OS (TiMOS) are supported.

ZacharyPuls:
In PR #995, I was a bit confused on the support for Alcatel-Lucent (Nokia) devices, as the README only mentioned support for the omniswitch platform.


laf:
Probably should be SROS without a space?

ZacharyPuls:
I would be happy to change it, but the Nokia documentation writes it with a space https://networks.nokia.com/products/sros.

PR #1006, opened 2017-08-31
Updated to ruby 2.4.1

chriseckhardt:
It passes the specs (mostly...modified the SSH proxy spec), but I haven't actually tried this out, yet.


laf:
For what it's worth, builds and runs fine on CentOS 7.

Not able to test past that.

wk:
The `ssh_spec.rb` fix from this PR has been implemented in #1267.

`Gemfile.lock` has been eliminated in #1263 to improve portability of the gem built and prevent de-syncs with the already pessimistic `oxidized.specfile`. Bumping `minitest` version from `5.8` to `5.10` does not appear to provide any benefit as all tests currently appear to pass under both.

@chriseckhardt Are you OK with closing this PR?

PR #1010, opened 2017-09-01
Added gsub filter to remove encrypted MD5 keys

colin-stubbs:
Added gsub filter to remove encrypted MD5 keys which change on every display of config.


lib/oxidized/model/fortios.rb
```diff
@@ -17,6 +17,7 @@ class FortiOS < Oxidized::Model
   cmd :secret do |cfg|
     cfg.gsub! /(set (?:passwd|password|secondary-secret|rsso-secret|psksecret|secret|key ENC)).*/, '\\1 <configuration removed>'
     cfg.gsub! /(set private-key).*-+END ENCRYPTED PRIVATE KEY-*"$/m , '\\1 <configuration removed>'
+    cfg.gsub! /(set md5-key 1 "ENC).*$/, 'set md5-key 1 "ENC <encoded key removed>"'
```
danilopopeye: could you rewrite to use the backreference?

``` ruby
cfg.gsub! /(set md5-key 1 "ENC).*/, '\\1 <encoded key removed>"'
```

PR #1024, opened 2017-09-13
Suppress timestamp response to ios commands

### LINKED ISSUE #1022: ASR 9k outputs a timestamp with every command
My first time posting anything so I am sorry if I go about this the wrong way. I recently set Oxidized up to backup the config of a new ASR 9k router. Unfortunately, it outputs the timestamp after every command which makes it think that the configuration has changed with every poll. My workaround has been to add `post_login 'terminal exec prompt no-timestamp'` at line 146 to turn this feature off. Could this be added to the code?

ja-frog:
At least some versions of ios output a timestamp after each input command. This will cause Oxidized to have a different version of the config for every poll of the device. I added the per-session command to suppress timestamps.

fixes #1022

### DIFF
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -140,7 +140,6 @@ class IOS < Oxidized::Model
     end
     post_login 'terminal length 0'
     post_login 'terminal width 0'
-    post_login 'terminal exec prompt no-timestamp'
     pre_logout 'exit'
   end
 
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -140,6 +140,7 @@ class IOS < Oxidized::Model
     end
     post_login 'terminal length 0'
     post_login 'terminal width 0'
+    post_login 'terminal exec prompt no-timestamp'
     pre_logout 'exit'
   end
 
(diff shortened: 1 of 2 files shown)


ja-frog:
Issue 1022

danilopopeye:
@ja-frog + @laf this command (`terminal exec prompt no-timestamp`) is default for all IOS devices? Shouldn't we create a flag for enabling this?

laf:
@danilopopeye Actually it's not, doesn't seem to be in at least 12.2 releases and I can't find when it may have been introduced.

ZacharyPuls:
@ja-frog Given that this is for an ASR 9K, wouldn't the change need to be under [iosxr.rb](https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/iosxr.rb)? From what I can see, `terminal exec prompt no-timestamp` isn't a valid command on any IOS/IOS-XE devices, just IOS-XR. It does accept `terminal exec prompt timestamp`, so this may be something you are able to turn on conditionally.

ja-frog:
@ZacharyPuls That is correct. I'm sorry, Cisco is new to me and I didn't think to check the other ios-type files. I will get that updated.

ja-frog:
Fixed, I believe. Hopefully without committing any gaffes.

PR #1034, opened 2017-09-19
enable/disable ssl certificate validation for githubrepo remote server

jodast:
The githubrepo hook currently lacks of a "secure: false" option to disable SSL verification for fetch and push operations on the remote server.

### DIFF
--- a/lib/oxidized/hook/githubrepo.rb
+++ b/lib/oxidized/hook/githubrepo.rb
@@ -11,11 +11,11 @@ class GithubRepo < Oxidized::Hook
 
     fetch_and_merge_remote(repo)
 
-    remote.push([repo.head.name], credentials: credentials)
+    remote.push([repo.head.name], credentials: credentials, certificate_check: certificate_validation)
   end
 
   def fetch_and_merge_remote(repo)
-    result = repo.fetch('origin', [repo.head.name], credentials: credentials)
+    result = repo.fetch('origin', [repo.head.name], credentials: credentials, certificate_check: certificate_validation)
     log result.inspect, :debug
 
     unless result[:total_deltas] > 0
@@ -43,6 +43,10 @@ class GithubRepo < Oxidized::Hook
   end
 
   private
+
+  def certificate_validation
+    lambda { |valid, host| cfg.has_key? 'secure' and cfg.secure == false ? true : valid }
+  end
 
   def credentials
     @credentials ||= if cfg.has_key?('username') && cfg.has_key?('password')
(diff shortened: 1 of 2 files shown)


tobbez:
I think this shouldn't be included.

If the hook is used with GitHub as a remote, you always want certificate validation.

If used with another Git host, that host should have a valid certificate, or one signed by your private CA, with its roots installed on the server running oxidized.

If the maintainers disagree and wish to include it, this PR needs to document the new option in the README.

laf:
@jodast please update the README to document this.

@tobbez Whilst it is good practice to enforce valid SSL, sometimes this isn't always possible so giving users the option to disable seems sane to me. Hopefully others will chime in.

jodast:
@tobbez you're definitely right. An ssl certificate always should be verified. But for simple testing purpose, where it is possible that you only have a self signed certificate it comes in pretty handy to be able to just switch the verification off.
Furthermore rugged resp. libgit2 removed support for GIT_SSL_NO_VERIFY environment variable and "http.sslverify" there is no way to bypass the veification.

lib/oxidized/hook/githubrepo.rb
```diff
@@ -44,6 +44,10 @@ def fetch_and_merge_remote(repo)
 
   private
 
+  def certificate_validation
+    lambda { |valid, host| @cfg.secure == false ? true : valid }
```
tobbez: Should use `cfg` rather than `@cfg` to be consistent with the code in all existing hooks.

Would be cleaner to explicitly check if the key exists as well, i.e. 
```
cfg.has_key? 'secure' and cfg.secure == false ? true : valid
```
rather than just
```
cfg.secure == false ? true : valid
```

jodast: These changes make perfectly sense for me. I will apply them right away.

tobbez:
Please squash these commits into one.

You can do this by first issuing `git rebase -i HEAD~3`, changing the first word on the lines for the two most recent commits to `fixup`, i.e:
```
pick c0a422f enable/disable ssl certificate validation for githubrepo remote server
fixup e3c0541 Update README.md
fixup 97ee141 Apply changes from review of @tobbez
```
save, exit and then force push the branch (`git push --force-with-lease`).

Other than that (and my initial objections that were overruled), LGTM.

danilopopeye:
@jodast could you update the tests for this class too, please?

jodast:
I'm sorry for the delayed reply. I was on vacation during the last two weeks. But I will update the tests very soon.

laf:
@jodast any update on this?

jodast:
Again, sorry for the late reply. Unfortunately I wasn't able to write the tests yet. But I'il try to finish them as soon as possible.

laf:
Thanks very much :)

laf:
@jodast Any update on this?

mortzu:
any news on this?

PR #1036, opened 2017-09-19
Add support for Supermicro SSE-G48-TG4

tobbez:
This model (and possibly others) runs an OS with a different set of commands than the Supermicro model that already exists.

I didn't find anything resembling a name of the OS, so I just used the model name for the class. If you think this name is fine I can update the PR to add it to the README. If not, please suggest a better name (maybe just Supermicro2?).

### DIFF
--- a/docs/Supported-OS-Types.md
+++ b/docs/Supported-OS-Types.md
@@ -128,7 +128,8 @@
  * Siklu
    * [EtherHaul](lib/oxidized/model/siklu.rb)
  * Supermicro
-   * [Supermicro](lib/oxidized/model/supermicro.rb)
+   * [Supermicro](lib/oxidized/model/supermicro.rb): Known to work with the Supermicro SSE-G2252.
+   * [Supermicro2](lib/oxidized/model/supermicro2.rb): Known to work with the Supermicro SSE-G48-TG4.
  * Symantec
    * [Blue Coat ProxySG / Security Gateway OS (SGOS)](lib/oxidized/model/sgos.rb)
  * Trango Systems
--- a/lib/oxidized/model/supermicro2.rb
+++ b/lib/oxidized/model/supermicro2.rb
@@ -1,0 +1,48 @@
+# Developed against:
+# #show version
+# Switch ID       Hardware Version                Firmware Version
+# 0               SSE-G48-TG4   (P2-01)           1.0.16-9
+
+class Supermicro2 < Oxidized::Model
+  prompt (/^(\e\[27m)?[ \r]*\w+# ?$/)
+
+  cfg :ssh do
+    post_login 'no cli pagignation'
+    pre_logout 'exit'
+  end
+
+  cmd :all do |cfg|
+    # * Drop first line that contains the command, and the last line that
+    #   contains a prompt
+    # * Strip carriage returns
+    cfg.delete("\r").each_line.to_a[1..-2].join
+  end
+
+  cmd :secret do |cfg|
+    cfg.gsub(/^(snmp community) .*/, '\1 <hidden>')
+  end
+
+  cmd 'show system information' do |cfg|
+    cfg.sub! /^Device Up Time.*\n/, ''
+    cfg.delete! "\r"
+    comment(cfg).gsub(/ +$/, '')
+  end
+
+  cmd 'show running-config' do |cfg|
+    comment_next = 0
+    cfg.each_line.map { |l|
+      next '' if l.match /^Building configuration/
+
+      if l.match /^Switch ID.*Hardware Version.*Firmware Version/ then
+        comment_next = 2
+      end
+
+      if comment_next > 0 then
+        comment_next -= 1
+        next comment(l)
+      end
+
+      l
+    }.join.gsub(/ +$/, '')
+  end
+end


laf:
I'm unsure if this should be merged into the existing model or not but the OS name should be generalised a bit more maybe. supermicro-sse? What's the difference with the other supermicro model and this for the kit it supports?

tobbez:
I don't think they should be merged, since the CLI interfaces seem to be completely different.

| Supermicro | Supermicro SSE-G48-TG4 | Comment |
| --- | --- | --- |
| show running-config | show running-config |
| show access-list tcam-utilization | show access-lists | May not be equivalent
| show memory | show meminfo | Probably not equivalent output - the SSE shows all memory usage information, which you don't want (since config changes would be detected every run)
| show system | show system information | The SSE includes uptime, which doesn't seem to be the case for the existing model
| show version | show version | 
| show watchdog | - |
| terminal length 0 | no cli pagination | 
| community ... | snmp community ... |

The SSE also doesn't seem to include passwords in `show running-config`.

Supermicro-sse wouldn't be a good name, since the commit message that adds the already included model states:

    Add supermicro support tested on a SuperMicro SSE-G2252

lib/oxidized/model/supermicrosseg48tg4.rb
```diff
@@ -0,0 +1,45 @@
...
+      if comment_next > 0 then
+        comment_next -= 1
+        next comment(l)
+      end
+
+      l
+    }
+    # drop first line that contains the command, and
+    # the last line that contains a prompt
+    cfg[1..-2].join
```
danilopopeye: you could put this in a `cfg :all` block so you do not repeat it in both `cmd`

lib/oxidized/model/supermicrosseg48tg4.rb
```diff
@@ -0,0 +1,45 @@
...
+    cfg.delete! "\r"
+    comment(cfg.each_line.to_a[1..-2].join).gsub(/ +$/, '')
+  end
+
+  cmd 'show running-config' do |cfg|
+    comment_next = 0
+    cfg = cfg.delete("\r").gsub(/ +$/, '').each_line.map {|l|
+      next '' if l.match /^Building configuration/
+
+      if l.match /^Switch ID.*Hardware Version.*Firmware Version/ then
```
danilopopeye: do you mind posting a sample of the problem you are solving here?

lib/oxidized/model/supermicrosseg48tg4.rb
```diff
@@ -0,0 +1,45 @@
...
+  prompt (/^(\e\[27m)?[ \r]*\w+# ?$/)
+
+  cfg :ssh do
+    post_login 'no cli pagignation'
+    pre_logout 'exit'
+  end
+
+  cfg :secret do |cfg|
+    cfg.gsub!(/^(snmp community) .*/, '\1 <hidden>')
+    cfg
```
danilopopeye: 💅 no need for this since `#gsub!` already returns the modified `cfg` 🤓

lib/oxidized/model/supermicrosseg48tg4.rb
tobbez: The output of `show running-config` starts with these lines:
```
Building configuration...
Switch ID       Hardware Version                Firmware Version
0               SSE-G48-TG4   (P2-01)           1.0.16-9
                                  
ip address [...]
```

`Building configuration...` should be removed, and the `Switch ID[...]` and subsequent line is not actually configuration, and should thus be commented.

tobbez:
Fixed the `:secret` issue.

Also fixed the duplication between commands, except for `gsub(/ +$/, '')`, which must be performed last and as such needs to be called for each individual command. It would be useful to be able to specify a command that executes after all individual commands (i.e. `:all` -> `:secret` -> `cmd 'command'` -> `:all_last`), but there's nothing in place for that right now.

For the `show running-config` logic see the separate (in-line) comment: https://github.com/ytti/oxidized/pull/1036#discussion_r141085391

danilopopeye:
> Also fixed the duplication between commands, except for `gsub(/ +$/, '')`, which must be performed last

@tobbez not sure why we need to remove these trailing white space, not sure is worth the CPU. any real gain by doing this?

danilopopeye:
@tobbez can you also update the README with this new model, please

tobbez:
Why? Because both machines and people like clean configs, without a lot of extraneous whitespace characters in them. I doubt the amount of CPU required by this is significant in comparison to that used by everything else Oxidized does.

Regarding the README, in my initial comment I wrote:

> I didn't find anything resembling a name of the OS, so I just used the model name for the class. If you think this name is fine I can update the PR to add it to the README. If not, please suggest a better name (maybe just Supermicro2?).

but you didn't address it. Does this mean you find the current model name good enough as it is?

tobbez:
The `user_pending` label on this pull request is incorrect, I'm waiting for a reply to my previous comment.

laf:
@tobbez Personally I don't see this model number working. It's far too specific unless genuinely it is the only model this will work with but that looks unlikely.

danilopopeye:
> The user_pending label on this pull request is incorrect, I'm waiting for a reply to my previous comment.

@tobbez yeah, I forgot to remove after you replied. my bad!

danilopopeye:
@laf maybe we could name it `SuperMicro2` and list the supported versions for each in the docs. This become a bikeshedding 😞 wdyt?

tobbez:
Rebased on master, renamed to Supermicro2, and README updated.

wk:
@laf, @ytti, @yeled, @tobbez

Sorry to further the bike-shed discussion, but this is actually a somewhat special and interesting situation. SuperMicro does not have its own network OS (or network kit), but re-brands whitebox equipment from a number of vendors, sometimes very different vendors (and operating systems) under rather similar model names.

The current `supermicro.rb` should work with the following SuperMicro devices:

```text
Supermicro SSE-G2252
Supermicro SSE-G2252P
```

* Actual OS: Edge-corE EdgeCOS
* Hardware: [ECS4210-52P](https://www.edge-core.com/productsList.php?cls=2&cls2=13&cls3=54) (non-existent model)

Thus, this can be renamed into `edgecos.rb` (for generic Edge-corE support), and supermicro.rb can `require` `edgecos.rb` verbatim for now as backwards compatibility shim.

The proposed `supermicro2.rb` in this PR (#1036) should work with the following SuperMicro devices:

```text
Supermicro SSE-G48-TG4 
Supermicro SSE-G24-TG4 
Supermicro SSE-X24S 
Supermicro SSE-X24SR 
Supermicro SSE-X3348S 
Supermicro SSE-X3348SR 
Supermicro SSE-X3348T 
Supermicro SSE-X3348TR 
Supermicro SBM-GEM-X2C 
Supermicro SBM-GEM-X2C+ 
Supermicro SBM-GEM-X3S+ 
Supermicro SBM-XEM-X10SM 
```

* Actual OS: Aricent [ISS](https://www.aricent.com/solutions/enabling-software/intelligent-switching-solution-iss) (Intelligent Switching Solution)
* Hardware: Various Marvell

This operating system type can also be encountered as Oracle SEFOS and Lenovo N/OS.

Thus, this can be renamed `aricent.rb`, and `supermicro2.rb`, `lenovo.rb`, as well as `oracle.rb` can `require` it verbatim as links.

Perhaps a bigger question here is what's the way forward to handling re-branded equipment and model type linking, so that it remains easily discoverable for users, easily extensible with model specifics, and is suitable for a use case where one brand actually ships multiple different operating system types under different models.

Special thanks to @erik-smit for his help with the requisite OEM archaeology.

wk:
One potential approach:

* Introduce a `deprecated` helper function for model `cmd` post-processing that appends a standard short comment to each command output suggesting the model currently used is about to become deprecated in a future version (to allow, in this case, transitioning away from `supermicro.rb` with minimal impact for users over a release or two).
* Keep names of actual OS types as close as possible to the underlying OS (so here, `edgecos.rb` and `aricent.rb` or `iss.rb`).
* Document compatibility mapping in [Supported OS Types](https://github.com/ytti/oxidized/blob/master/docs/Supported-OS-Types.md) by linking products to model.

PR #1037, opened 2017-09-20
Add better documentation for Model

tobbez:


laf:
This looks great. I do think we should structure the docs a little more as it's getting to be a big README.

tobbez:
Can we get this merged first, and then split the README into several files in another pull request?

It would still be a definite improvement, even though it makes the README longer. I don't think it makes sense to wait for an eventual split, nor incorporating it into this pull request.

PR #1039, opened 2017-09-22
support passphrases on ssh key in githubrepo

yeled:
Due to https://github.com/libgit2/rugged/issues/530 we need a newer libssh2 to support a passphrase on an SSH key.

This necessarily needs Dockerfile hacking.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,25 @@
-FROM phusion/baseimage:0.9.18
+FROM phusion/baseimage:0.9.22
 MAINTAINER Samer Abdel-Hafez <sam@arahant.net>
 
 RUN add-apt-repository ppa:brightbox/ruby-ng && \
 	apt-get update && \
   apt-get install -y ruby2.3 ruby2.3-dev libsqlite3-dev libssl-dev pkg-config make cmake libssh2-1-dev git g++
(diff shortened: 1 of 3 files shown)


danilopopeye:
seeing all the contortion to install the needed `libssh2` package to ubuntu 16 maybe it's time to start using the official ruby image. the `2.4-strech` already has `libssh2` __1.7__ in the official debian repo.

wdyt? /cc @ytti @mikebryant @laf

yeled:
I chose the safer phusion image due to init magic. I didn't look very hard (none at all) etc the https://github.com/ytti/oxidized/tree/master/extra directory but assumed there was some tie-in there.

laf:
As in the one from [here](https://hub.docker.com/r/library/ruby/tags/)? I'd say yes if it meant the docker build was less hacky.

danilopopeye:
@laf exactly, I guess `ruby:2.4-stretch` already has most of the dependencies.

> I chose the safer phusion image due to init magic.

Docker from `1.13` has [tini](https://github.com/krallin/tini) built in

yeled:
Docker is good at hiding a mess, and making it repeatable.

Considering that this patch is ready *now* and works, I suggest just pushing the Dockerfile hacking so we can use ssh-keys with passphrases.

yeled:
fixed in https://github.com/ytti/oxidized/issues/1204

PR #1042, opened 2017-09-25
Docs refactor

laf:
This is by no means complete or perfect.

I've took a stab at sensibly splitting out the README file into multiple docs whilst trying to leave the install instructions in README as that's where users will first end up.

#987 Could then be added to this by including a new API file.

Two other options if this isn't the path people want to go down would be to use the Wiki or a full docs.oxidized.x site(?)

Tagging @ytti, @nopedial and @danilopopeye for comments.


nopedial:
Looks good to me, great work!

danilopopeye:
great work @laf 👏 

wdyt about breaking some more about topics: `output`, `hooks`, etc ?

laf:
@danilopopeye more than happy to do that and I'll fix the ``` issues

laf:
@danilopopeye split more out and also fixed the ``` back ticks.

I was wondering also, we do a fair bit of support for oxidized in the ##librenms irc channel and our community forum. Gitter is quite quiet so I was wondering if we you'd be open to us running a section of community.librenms.org dedicated to Oxidized support and put it in the readme?

danilopopeye:
> Gitter is quite quiet so I was wondering if we you'd be open to us running a section of community.librenms.org dedicated to Oxidized support and put it in the readme?

absolutely @laf, I'll create a user there and give you a ping afterwards 😉

PR #1043, opened 2017-09-27
Allow configure mode in prompt regex

mrplow87:
To properly use oxidized-script and switch to configure terminal, the prompt regex needs to accept `<hostname>(config)# ` and all other context layers (vlan-... etc.).

With the given regex, oxidized-script times out once the context is switched to configure terminal because the regex does not match anymore.

### DIFF
--- a/lib/oxidized/model/nos.rb
+++ b/lib/oxidized/model/nos.rb
@@ -2,7 +2,7 @@ class NOS < Oxidized::Model
 
   # Brocade Network Operating System
 
-  prompt /^(?:\e\[..h)?[\w.-]+(\([\w.-]+\))*# $/
+  prompt /^(?:\e\[..h)?[\w.-]+(\([\w.-]+\))?# $/
   comment  '! '
 
   cmd :all do |cfg|
--- a/lib/oxidized/model/nos.rb
+++ b/lib/oxidized/model/nos.rb
@@ -2,7 +2,7 @@ class NOS < Oxidized::Model
 
   # Brocade Network Operating System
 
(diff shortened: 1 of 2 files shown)


mrplow87:
"_But wait, there's more_". Some older firmwares of HP Procurve do not accept <code>password username plaintext password</code> which caused the prompt regex to fail when issuing <code>password manager</code>.

<pre>host(config)# password manager
New password for manager: ...
Please retype new password for manager: ...
host(config)# </pre>

Basically <code>^\r?([\w .-]+: |([\w.-]+(\([\w.-]+\))*#) )$</code> would allow any prompt ending with <code>: </code> additionally to a proper prompt. But I'm not sure, if this is the way we want to go here.

mrplow87:
Similar changes to allow configure terminal may apply to other hardware as well.
* nos.rb -> <code>^(?:\e\[..h)?[\w.-]+(\([\w.-]+\))*# $</code>

danilopopeye:
the block `(config)` you mentioned can repeat? if not you could replace the `*` for `?` :neckbeard:

mrplow87:
Oh, that's my laziness and old habit to get zero occurances of that group matched. Will be replaced!

ytti:
Definitely a legit problem. But I think we probably should have separate regexp for different state.

Possibly we should make prompt an instance of new prompts class. And for backward compatibility 'prompt = x' would be equivalent to 'prompt.exec = x'. Then we add 'prompt.config = y'. 

But then the class probably should indicate WHEN should we expect each prompt.

I think this needs more thought.
7

mortzu:
Please fix the conflict and add your change to CHANGELOG

PR #1047, opened 2017-09-28
Update fortios.rb

sfini:
Modify script for discard changes private key encryption every time config is downloaded  e modify show with show full-configuration 
https://github.com/ytti/oxidized/issues/931

PR #1057, opened 2017-10-03
Cluster uptime spam information

brunobritocarvalho:
Everyhour, a new message about the "Cluster uptime". I believe this was added in 5.6 version of Forti.

### DIFF
--- a/lib/oxidized/model/fortios.rb
+++ b/lib/oxidized/model/fortios.rb
@@ -18,6 +18,11 @@ class FortiOS < Oxidized::Model
     cfg.gsub! /(set (?:passwd|password|secondary-secret|rsso-secret|psksecret|secret|key ENC)).*/, '\\1 <configuration removed>'
     cfg.gsub! /(set private-key).*-+END ENCRYPTED PRIVATE KEY-*"$/m , '\\1 <configuration removed>'
     cfg.gsub! /(IPS Malicious URL Database).*/, '\\1 <configuration removed>'
+    cfg.gsub! /.*(APP-DB).*/, '\\1 <configuration removed>'
+    cfg
+  end
+
+  cmd :reduce do |cfg|
     cfg.gsub! /(Cluster uptime).*/, '\\1 <configuration removed>'
     cfg
   end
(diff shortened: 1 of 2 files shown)


jsynack:
This is added to :secret which is only called when `remove_secret: true` is set. 

Should probably be in more generic spot for the model? 

Using this example - I may want to backup my secrets (that is `remove_secret: false`) but I still don't want the repo/config churn because of an uptime stamp that is being incremented.

💭 This also probably brings up a need/desire to have a new method generically for models - that there should be a `:reduce_churn` (or whatever name works) - removing operational counters/things that is outside of the scope of `:secret` (which has other implications). 

This would also give a place to do all the 'normal' gsub fixups such as that.

 But I may be overthinking it. I can see it being useful to have the gsubs right under the device command that was just executed for ease of troubleshooting during model misbehavior versus having to hunt for them in a larger `model.rb`.

laf:
Well spotted. @brunobritocarvalho can you move this so you don't need to enable secrets for it to work please.

brunobritocarvalho:
Point taken, I agreee. Kinda wanted to do that at first but guess I got lazy along the way. 
This should do the trick I believe. Thanks for the input.

laf:
I was just more referring to moving the new line outside of the config check.

I'm not personally (not the projects position) sure that adding another config option to remove what you'd expect to happen by default is the way forward and we should strip out the bad data like this for all.

brunobritocarvalho:
Hmm yeah, also makes sense. I just figured some people would like to have that information and some other would not. I'll make that change tomorrow then.

ytti:
I'm on the fence about reduce_churn. Isn't the expectation that dynamic data, like uptime is not included? Why would @brunobritocarvalho request be unique? 

If it's not unique, we'd need to change all models to output dynamic data, unless variable is set?

laf:
I'd vote not to put a config option in for this and just remove this type of config churn.

jsynack:
Sorry - I wasnt as clear as I could have been.

 My thought process for those comments was to break out the `gsub` and dynamic data removal/reductions in a different method within the model file - but not have it be controlled by a configuration setting / additional variable.  Just have it grouped for easier understanding of what happens in each method.  Sort of a 'this method pulls the config', 'this method scrubs the config' thought process. I am not tied to it at all. 

There is/should be an expectation that a model will behave itself 'out of the box' - and that each poll interval does not cause churn.

wk:
This has been closed by #1047 which introduced unconditional uptime filtering for the `fortios.rb` model.

PR #1062, opened 2017-10-10
Add haproxy example

denvera:


laf:
Is this specific to redhat and / or debian based distros. If only one of them is it worth changing the filename to offer more clarity?

bitfiretech:
This should apply to any haproxy package across distributions.

denvera:
Hey, this should be generic across distributions, though the default/global sections are based on a default installation on Ubuntu. The backend and frontend sections are definitely applicable to any haproxy installation. Happy to rename if you deem necessary.

laf:
> Hey, this should be generic across distributions, though the default/global sections are based on a default installation on Ubuntu. The backend and frontend sections are definitely applicable to any haproxy installation. Happy to rename if you deem necessary.

As in they reference file paths specific for Ubuntu? If so, could you rename it to `extra/oxidized-ubuntu.haproxy` or similar?

denvera:
Have renamed to `oxidized-ubuntu.haproxy` to be explicit, as some of the default/global section paths may not work for other distros.

PR #1064, opened 2017-10-13
Add docs & error checking

MajesticFalcon:
Add FreeBSD documentation for mysql.
Add error checking in case user forgets/misformats the source mapping. Don't fail with a hard to track error message, let the user know why it didn't work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - allow any max length for username/password in GcomBNPS (@freddy36)
 - relax prompt requirements in ciscosmb (@Atroskelis)
 - fortios model strips uptime even without remove_secrets (@jplitza)
+- HP ProCurve now accepts ">" as apart of the prompt (@magnuslarsen)
 
 ## [0.27.0] - 2019-10-27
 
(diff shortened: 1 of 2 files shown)


ytti:
What type of error does it catch? Is it not possible to have it entirely in source.rb? Seems not very DRY.

laf:
@MajesticFalcon Any response to the last comment?

wk:
Is an empty source map meant to be an error condition in the first place?

It seems like a plausible outcome, especially for `http` defined source maps, that the external map provider may temporarily have no nodes, but will have them in the future. It might also be a viable outcome for the `csv` map if hot-reload (via for example `rb-inotify`) functionality is added.

README.md
```diff
@@ -91,7 +91,9 @@ Install all required packages and gems.
 
 ```shell
 pkg install cmake pkgconf
+pkg install mysql-connector-c-6.1.11
```
wk: Since this only applies for users of the `MySQL` source, it seems like this would be a better fit for `docs/Sources.md` under the `Source: MySQL` section?

README.md
```diff
@@ -91,7 +91,9 @@ Install all required packages and gems.
...
 gem install oxidized
+gem install mysql2 sequel
```
wk: Ditto here...

docs/Sources.md
```diff
@@ -39,6 +39,7 @@ source:
   default: sql
   sql:
     adapter: mysql2
+    host: server_host
```
wk: Specifying a host that isn't localhost is indeed a common use case, so this enhances the example!

laf:
> It seems like a plausible outcome, especially for http defined source maps, that the external map provider may temporarily have no nodes, but will have them in the future. It might also be a viable outcome for the csv map if hot-reload (via for example rb-inotify) functionality is added.

Or when you first setup LibreNMS and hook Oxidized into it, no nodes will be present so it shouldn't bail out at that point and instead allow oxidized to start still.

@MajesticFalcon Any response to all the comments in this PR?

PR #1067, opened 2017-10-17
Allow whitespaces in hostname

stiltzkin10:
Without this change a prompt that look like this will fail: "MY SWITCH#"

### DIFF
--- a/lib/oxidized/model/procurve.rb
+++ b/lib/oxidized/model/procurve.rb
@@ -2,7 +2,7 @@ class Procurve < Oxidized::Model
 
   # some models start lines with \r
   # previous command is repeated followed by "\eE", which sometimes ends up on last line
-  prompt /^\r?([\w.-]+# )$/
+  prompt /^\r?([\w\s.-]+# )$/
 
   comment  '! '
 


cppmonkey:
I understand you want to make it as flexible as possible, but when spaces aren't permitted within a hostname. This would only be applicable in exceptional cases. Why can't you rename it and give it one that complies with internet standards?

> The Internet standards (Requests for Comments) for protocols mandate that component hostname labels may contain only the ASCII letters 'a' through 'z' (in a case-insensitive manner), the digits '0' through '9', and the minus sign ('-'). The original specification of hostnames in RFC 952, mandated that labels could not start with a digit or with a minus sign, and must not end with a minus sign. However, a subsequent specification (RFC 1123) permitted hostname labels to start with digits. **No other symbols, punctuation characters, or white space are permitted.**

Baptizein:
In the model file (in my example, /var/lib/gems/2.3.0/gems/oxidized-0.21.0/lib/oxidized/model/boss.rb) :
prompt /[^\s#>]+[#>]$/

wk:
While having spaces in hostnames of network equipment is a dubious operational practice, it is probably not up to Oxidized to pass judgement on how users configure their equipment - as long as the configuration is tolerated by the underlying OS type. 

@stiltzkin10 are there any concrete ambiguities in the input this relaxation might inadvertently match?  Is there a restriction on hostname lenght for these devices which can be used to make the match more specific?

stiltzkin10:
I impleneted Oxidized for a client that had spaces in their hostnames. Couldn't really tell them to change them.

Not really, kinda hard to make it more specific.

wk:
I would expect the hostname length procurve tolerates to not be infinite (maybe 32-64 characters). If we could place a number on this then that would allow at least constraining total prompt length by number of characters matched (including spaces)

ytti:
I'm ok with this, but if we wanted more safety, perhaps we should match for space surrounded by non-spaces, instead of space verbatim.

mortzu:
Please fix the conflict and add your change to CHANGELOG

PR #1071, opened 2017-10-18
Cisco ASA: gather inventory and status for the secondary machine in a redundant pair.

ik5pvx:
Please consider adding this to get complete inventory from a redundant pair. Useful when
oxidized is controlled by librenms, because in that case you can't add the secondary device to monitoring.


mortzu:
Please add it to changelog

mortzu:
duplicate of #1106

PR #1077, opened 2017-10-25
Hide snmp community

brunobritocarvalho:
Community is displayed next to this line (old ios versions)

### DIFF
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -26,7 +26,7 @@ class IOS < Oxidized::Model
 
   cmd :secret do |cfg|
     cfg.gsub! /^(snmp-server community).*/, '\\1 <configuration removed>'
-    cfg.gsub! /^(snmp-server host \S+( vrf \S+)?( version (1|2c|3))?)\s+\S+((\s+\S*)*)\s*, '\\1 <secret hidden> \\5'
+    cfg.gsub! /^(snmp-server host \S+( vrf \S+)?( version (1|2c|3))?)\s+\S+((\s+\S*)*)\s*/, '\\1 <secret hidden> \\5'
     cfg.gsub! /^(username \S+ privilege \d+) (\S+).*/, '\\1 <secret hidden>'
     cfg.gsub! /^(username \S+ password \d) (\S+)/, '\\1 <secret hidden>'
     cfg.gsub! /^(username \S+ secret \d) (\S+)/, '\\1 <secret hidden>'
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -26,7 +26,7 @@ class IOS < Oxidized::Model
 
   cmd :secret do |cfg|
     cfg.gsub! /^(snmp-server community).*/, '\\1 <configuration removed>'
-    cfg.gsub! ^(snmp-server host \S+( vrf \S+)?( version (1|2c|3))?)\s+\S+((\s+\S*)*)\s*, '\\1 <secret hidden> \\5'
+    cfg.gsub! /^(snmp-server host \S+( vrf \S+)?( version (1|2c|3))?)\s+\S+((\s+\S*)*)\s*, '\\1 <secret hidden> \\5'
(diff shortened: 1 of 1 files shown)


ZacharyPuls:
Won't this remove some valid config?

We could use this RegEx instead:

`cfg.gsub! ^(snmp-server host \S+( vrf \S+)?( version (1|2c|3))?)\s+\S+((\s+\S*)*)\s*, '\\1 <secret hidden> \\5'`

Test inputs:

`snmp-server host 10.10.25.5 version 2c net.ksfiber.configmgmt config `
`snmp-server host 10.10.25.5 net.ksfiber.configmgmt config `
`snmp-server host 1.2.3.4 vrf Mgmt-intf version 2c test  mpls-vpn `
`snmp-server host 1.2.3.4 test `
`snmp-server host 1.2.3.4 test mpls-vpn aaa_server `

Test outputs:

`snmp-server host 10.10.25.5 version 2c <secret hidden> config`
`snmp-server host 10.10.25.5 <secret hidden> config`
`snmp-server host 1.2.3.4 vrf Mgmt-intf version 2c <secret hidden> mpls-vpn`
`snmp-server host 1.2.3.4 <secret hidden> `
`snmp-server host 1.2.3.4 <secret hidden> mpls-vpn aaa_server`

brunobritocarvalho:
I did not encounter those examples when I looked at my config files, but your regex does sound a lot better. My main concern was to remove those community ASAP, guess I did not spend enough time looking at it, thanks a lot!

laf:
@brunobritocarvalho you updating this PR?

@ZacharyPuls thanks for the input :)

brunobritocarvalho:
@laf guess it should work now, I'll test it tomorrow, sadly I can't do it tonight. Thanks guys.

laf:
@ZacharyPuls You happy with this?

@brunobritocarvalho Ace, if you can confirm that would be great.

brunobritocarvalho:
Hey guys, just tested today, we have this use case still appearing: 

+snmp-server host $IP $PASSWORD

So Zachary's regex does not entirely work in our context. As for the other scenarios, it looks alright.

ZacharyPuls:
@laf Looks good to me!
@brunobritocarvalho
I just tested this again on regexr, this should be working, here's what I'm getting: https://screenshots.firefox.com/HG72yFnrAvA05whu/regexr.com

Thanks for adding in the /^/ for me, forgot to add it back when pasting from the RegEx tool.

laf:
@ZacharyPuls Thanks :D

@brunobritocarvalho Let me know when this is updated to the final version.

zombah:
Maybe it is possible to make this change as additional version of ios.rb like "extra secure ios"?
For example i have miriad of different communities on my ios devices and i will need to make my own version after this change, to not hide communities, because i need to backup them and track their changes.

brunobritocarvalho:
@zombah sounds like an idea, any idea how to do it, I am no ruby expert sadly. 
Would it be something like this? 
`  cmd :extra_secure do |cfg| 
    cfg.gsub! /^(snmp-server host \S+( vrf \S+)?( version (1|2c|3))?)\s+\S+((\s+\S*)*)\s*/, '\\1 <secret hidden> \\5'
    cfg
  end 
`
@ZacharyPuls I tested the regex again, it does seem to work, don't know why it failed last time. : 
`cfg = 'snmp-server host 127.0.0.1  Psdadasdasw2123123'
cfg.gsub! /^(snmp-server host \S+( vrf \S+)?( version (1|2c|3))?)\s+\S+((\s+\S*)*)\s*/, '\\1 <secret hidden> \\4'
puts cfg
`

laf:
Personally (not the projects stance). Having more config options for things like this seems ott. We already have the secure config option which if set is where this change affects so if you want to see communities then you shouldn't set the secure flag.

I do realise that what should be classed as 'secure' is subjective but this model already removes the communities from other config lines so to me this seems like it was just missed in the first place.

brunobritocarvalho:
@laf Makes sense, that's also what I thought at first. As for me, I think everything is fine now, the existing code should work fine.

zombah:
Ok i'll go for "remove_secret: false" then

Issue #1087, opened 2017-11-08
New versions of Cumulus use FRR rather than Quagga


User4574:
The current model for Cumulus is currently hard coded to get the routing daemon config from /etc/quagga.

In modern versions of Cumulus Quagga has been replaced with FRR, and the directory is no longer symlinked by default.

The model will need updating to support devices running FRR, either by automatically detecting this, or with a configuration variable to select which routing daemon should be backed up.

PR #1089, opened 2017-11-08
Add support for frr in new versions of cumulus

User4574:
This fixes the issue I raised, #1087.

### DIFF
--- a/lib/oxidized/model/cumulus.rb
+++ b/lib/oxidized/model/cumulus.rb
@@ -15,6 +15,10 @@ class Cumulus < Oxidized::Model
   
   #show the persistent configuration
   pre do
+    # Set frr or quagga in config
+    routing_daemon = vars(:cumulus_routing_daemon)
+    routing_daemon_shout = routing_daemon.upcase
+
     cfg = add_comment 'THE HOSTNAME'
     cfg += cmd 'cat /etc/hostname'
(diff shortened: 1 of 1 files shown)


laf:
Actually just looking at this again, do you need to set the default to quagga so it doesn't break old installs on update? @User4574

ytti:
I think we should set the variable if vars does not, to be backward compatible.

bobthebutcher:
Hello Team,

If @User4574 is unable to address the comments would you consider someone else taking over this PR ? I would be happy to take over with a bit of guidance. I am pretty new to ruby but with some guidance I think I can handle this one. 

Its usefully for my $job so I am keen to get this one across the line.

Cheers,

Brad

wk:
@bobthebutcher based on what I can see, there are three elements here that could use attention:

* Re-base commit on current version of `cumulus.rb` (from what I can see, only cosmetic changes)
* Look into giving `routing_daemon` a sensible default value and pay attention to capitalization of default(s) and FRR names (so that devices with no frr still work as they do today).
* Documentation (perhaps by adding a file to `/docs/Model-Notes`) and `CHANGELOG.md` update.

Together with a bit of testing, it would be a welcome enhancement.

wk:
Closed by #1375

PR #1094, opened 2017-11-10
Fix prompt pattern for model/comware.rb to handle ASCII Bell response +fix minor README.md typo

udha:
Changes `prompt /^\0*(<[\w.-]+>)$/` to `prompt /^\0*(<[\w.-]+>).?$/` to handle situations where ASCII Bell character is sent at the end of the prompt by Comware HP/H3C/3Com devices.

Also fixes minor type of 'HTT' to 'HTTP' in README.md file.


lib/oxidized/model/comware.rb
```diff
@@ -42,6 +42,10 @@ class Comware < Oxidized::Model
     end
 
     post_login 'screen-length disable'
+    # optionally you can add into system-view:
```
laf: The screen-length disable command is run all the time anyway so I'm not sure those notes are useful.

laf:
@Ultra2D / @amarti2038 - As you were the last two users to edit this model, would either of you be able to test this or confirm it's ok to merge?

udha:
Do you want me to remove the line disable comments?

udha:
I can also provide a debug dump of a connection failing before this change and then working successfully after the change if that helps

Ultra2D:
Works for me, but note that I only manage 1 Comware stack (2 devices).

laf:
> Do you want me to remove the line disable comments?

Yes please :)

> I can also provide a debug dump of a connection failing before this change and then working successfully after the change if that helps

Probably not necessary, just trying to avoid breaking anything :)

udha:
> Yes please :)

All cleared, thanks for the guidance 👍

PR #1095, opened 2017-11-10
feature Allow oxs to call for single host

laf:
Before hand, when using oxs it would load the entire list of nodes up. Using http as a source means a big api call to make if you have a lot of devices with the possibility of this taking a while.

We've been running this live for a couple of weeks now with no noticeable issues. I've tested the sql source in my dev install and it seems to work but I'm not able to test that in anger.


ytti:
Needs conflict resolution

laf:
Thanks for the reminder. I've rebased and fixed the conflicts now.

laf:
Thanks @ytti :)

PR #1096, opened 2017-11-10
Basic support for ArbOS from Arbor Networks.

jsynack:
ArbOS is the generic name for Arbor Networks (https://www.arbornetworks.com/) OS that runs on some of their devices.

Found and imported from https://github.com/claranet/oxidized . Credit to Claranet for the creation.

Locally validated on Peakflow SP v8.2+

The module is quite simple as the command 'config show' automatically disables pagination. There is a pagination-enabled version of the command as well.


laf:
@sanjmonkey are you ok with this being included?

sanjmonkey:
Sure, happy for it to be included @laf.  I just pushed https://github.com/claranet/oxidized/commit/e31f7b42ff7dac3b613536685c18bee1eb4923f9 which includes showing the active IP ACL's - @jsynack - does that also work for you?

laf:
@sanjmonkey thanks for getting in touch :)

@jsynack - if you can take a look at that last commit and see if it's ok to be in here that would be ace.

jsynack:
The IP access configuration rules are shown in the 'config show' for my output on an Arbor SP 8.2 device. I haven't looked at a TMS yet.  Perhaps a difference of the product lines or versions ?

If it is needed for other product lines - since the output of 'ip access show' is not 'pure config' but a command output, should this be done as a comment when stored in oxidized?

sanjmonkey:
In SP 8.1.1, configured IP ACL's are shown in `config show` too, though `ip access show` also shows the immutable auto-configured rules, which I had a use for seeing.

Point taken re: commenting, see: https://github.com/claranet/oxidized/commit/89f6f813e77cf336702b0b5cf611bceebffcc94f

laf:
@sanjmonkey Thanks again.

@jsynack Does this work for you now?

jsynack:
Sorry for the delay - After putting this `arbos.rb` into place and restarting , I started to get timeouts on the Arbor device when attempting to poll it. This is without any resource problems on the oxidized host or reachability problems to the Arbor device.

When I watch the output manually of the Arbor CLI `config show` dump on the arbor device - it takes about 31-32seconds(measured with stopwatch).  When I activate the input debugging for oxidized - it (oxidized) is still receiving the output when the timeout cancels the connection at 120seconds. 

This was done with starting oxidized and moving to the head of the line with the API call `/node/next/<nodename>` , with `next_adds_job: true` .  So there were probably other nodes running based on the interval and node count. 

I then reverted this change and went back to the original `arbos.rb`, as well as turned off debugging in oxidized to see if there was some bottleneck that I was running into that was unrelated to this. Continued to hit the same timeout for this specific device.

I am going to let oxidized startup settle down and see if during normal/natural polling from the node queue works OK.  

I think this may point to some other I/O blocking or limiting issue within the oxidized framework during device polling (and probably unrelated to this/Arbor). 

This is very odd where the manual running of the command is a 31-32second output and oxidized triggers a timeout at 120seconds. When I have some more information/debugging I will open a proper issue and include all of this information. And yes, the oxidized host has plenty of spare resources and low latency/no loss to the device that is timing out. 

When I look at the configuration that had been previously pulled - it is 689KB in the backend git repo at about 5.6K lines. I have other devices with larger configurations that are different models. 

All that is to say that I think the Arbor change is fine - but it highlights something else perhaps that only triggered during a restart and a `/node/next` .

jsynack:
It looks like the removal of `exec true` caused the problems. 

Adding this back in gets me back to normal config fetching in the expected timeframe. There is some code path that must be getting hit that loops forever as the oxidized process burns up 100% CPU during this time (on a single core) until the timeout fires.

@sanjmonkey - Do you have any other local customizations that make the `exec true` take place elsewhere?  Do you have success without the `exec true` in `arbos.rb` ?

sanjmonkey:
@jsynack - I've made no other customisation to run this in an exec channel.  Plus, I've just tested with using both shell and exec channels and both work for me (SP 8.1.1) and both get config in ~8s on one of our SP's.  Current Arbos SSHd version: `OpenSSH_6.7p1, OpenSSL 1.0.1e-fips 11 Feb 2013`

However, I do recall ripping out `exec true` in one of my earlier `arbos.rb` commits and I seem to remember it being related to the Arbos OpenSSH version not closing the exec channel properly on exit, but this was whilst on a different version of SP - probably a version of 7, I don't remember exactly which one - and as we were planning an upgrade to 8.1 we didn't bother taking it further.

So, I dont mind if you want to keep `exec true` in, it doesn't affect me.  This thread might help those planning on using oxidized with earlier versions of Arbos / those hitting similar issues.

laf:
I've pushed a model note to say to remove exec true if people are using 7 or less. Thanks for putting all this together @jsynack (took me a while to realise what your username was :) and @sanjmonkey.

PR #1097, opened 2017-11-10
Add support for Broadband Network Platform Software from GCOM Technologies Co.,Ltd.

freddy36:


docs/Supported-OS-Types.md
```diff
@@ -76,13 +76,17 @@
    * [XOS](/lib/oxidized/model/xos.rb)
  * F5
    * [TMOS](/lib/oxidized/model/tmos.rb)
+ * Fiberstore
```
laf: How come this is in here?

lib/oxidized/model/gcombnps.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+  end
+
+  cfg :telnet, :ssh do
+    pre_logout 'exit'
+  end
+
+end
+
+# vim: set noai ts=2 sw=2:
```
laf: Can you remove these two lines please.

lib/oxidized/model/gcombnps.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  cmd 'show running-config' do |cfg|
+    cfg
+  end
+
+  cmd 'show interface sfp' do |cfg|
+    cfg = cfg.each_line.to_a
+
+    cfg = cfg
+      .reject { |line| line.match /^  Temperature/ }
```
laf: A lot of the other models do it this way:

https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/aireos.rb#L45

Can you switch to that or was it done like this for a reason?

docs/Supported-OS-Types.md
freddy36: GCOM also produces OEM hardware for other companies
The Fiberstore S3800 is one switch I'm aware of.
I added it so other users can easily find the correct module.

laf: Ok, makes sense.

freddy36:
requested changes have been made

PR #1099, opened 2017-11-15
Netgear model: Permit . as valid character in prompt

pv2b:
I have Netgear switch with hostname like "switch.example.net" which results in prompt like:

    (SWGE001-STO-DC1.ita.local) #

This is not considered valid by current prompt regex, so I added . as a valid character in the character class.


laf:
Thanks @pv2b :)

PR #1104, opened 2017-11-17
Procurve: Handle switch selection for stack commanders

### LINKED ISSUE #1070: HP Procurve stacking
Hello,

Trying to backup a HP Procurve with two members in a stack. Seems like Oxidized gets stuck during login when selecting member stack:

```
Enter switch number to connect to or <CR>:

  SN MAC Address   System Name      Device Type           Status                   
  -- ------------- ---------------- --------------------- -------------------------
  0  002347-457980 SW0         ProCurve 2810-48G    Commander Up             
  1  002347-459880 SW1      ProCurve 2810-48G    Member Up 
```               

I tried to add:  

```
expect /Enter switch number to connect to or <CR>:/ do
    send "\n"
    ""
end
```

to procurve.rb without success.

pv2b:
When trying to connect to a HP Procurve Switch with stacking enabled,
you are asked to choose which switch to connect to to manage.

This patch makes it so that if this question is encountered, just press
"Enter" to choose the stack commander.

This won't make backing up stack members work (they still need to be
configured seperately and given IP addresses, making stacking useless)
but at least it makes it possible to back up a commander in a stack
seperately.

Fixed #1070

Credit to @stiltzkin10 for this fix.

### DIFF
--- a/lib/oxidized/model/procurve.rb
+++ b/lib/oxidized/model/procurve.rb
@@ -21,6 +21,11 @@ class Procurve < Oxidized::Model
     ""
   end
 
+  expect /Enter switch number/ do
+    send "\n"
+    ""
+  end
+
   cmd :all do |cfg|
     cfg = cfg.each_line.to_a[1..-2].join
     cfg = cfg.gsub /^\r/, ''


pv2b:
This is the exact same change as #1108 so I'll close this one.

Issue #741, opened 2017-02-22
Issue with reported last config change on Cisco ASA


ecm75:
Simmilar to the earlier reported uptime causing constant config updates, I had an issue with Cisco ASAs reporting slightly different timestamps for the last config change.

Here's a patch to work around this issue by simply cutting the seconds part:
```
--- a/lib/oxidized/model/asa.rb
+++ b/lib/oxidized/model/asa.rb
@@ -23,6 +23,7 @@ class ASA < Oxidized::Model
     # avoid commits due to uptime / ixo-router01 up 2 mins 28 secs / ixo-router01 up 1 days 2 hours
     cfg = cfg.each_line.select { |line| not line.match /(\s+up\s+\d+\s+)|(.*days.*)/ }
     cfg = cfg.join
+       cfg.gsub! /(Configuration last modified by) (\S+) at (\d+:\d+)(:\d+\.\d+) (.*)/, '\1 \2 \3 \5'
     comment cfg
   end
```

ytti:
But last modified only changes when config is written, so why does it cause constant change?

I'm fine by removing it from the output, just curious.

ecm75:
Yea that's what you'd expect to see, but here's what I see on several of our devices when issuing several subsequent show version commands:
```
router# show version | include last modified
Configuration last modified by enable_15 at 11:19:34.290 CEST Wed Nov 2 2016
router# show version | include last modified
Configuration last modified by enable_15 at 11:19:34.450 CEST Wed Nov 2 2016
router# show version | include last modified
Configuration last modified by enable_15 at 11:19:34.720 CEST Wed Nov 2 2016
router# show version | include last modified
Configuration last modified by enable_15 at 11:19:34.210 CEST Wed Nov 2 2016
```

The date kind of "jitters"...
I still have to investigate if it's just on some fw versions...

jsynack:
+1 on this.

I  have seen this on several older ASA units. The msec can jitter, so removing the msec and second from the output should keep things running cleaner with just the hour:minute in the output.  While there is the edge case that a jittering second could 'flip' the minute - it is probably not worth the extra code to handle that case.

As @ecm75 shows - this jitters even on the same CLI session , with no actual changes that have taken place for months on the unit. And it is not impacted by NTP sync status/etc. It can skip forward or go backwards 

We modified our local `asa.rb` for this - I wanted to confirm that someone else has seen this in the wild as well.

This was run back-to-back:

```
dsfw3058> show ver | inc mod
[...]

PR #1106, opened 2017-11-18
Implement two fixes in ASA model behavior

jsynack:
1. Fix for Issue #741 

 Some ASAs / revs will jitter the 'configuration last update' time even when there has not been a configuration change.

 Trim off the msec and seconds to avoid most jitter scenarios as a compromise but still retain the last configuration modification hour and minute.

 credit to @ecm75 for the adjustment.

This will shows a diff for any installed base of ASA devices due the msec and seconds being trimmed out.

Closes ytti/oxidized#741

2.  Implement PR #1071 in a slightly different fashion.

Failover information gathering commands added, wrapped in a test of the failover active status from the 'show failover' command. 

 credit to @ik5pvx for the desired command list and original PR.
 
 The 'headers' are added so that it is clear the next command is reporting information on the firewall mate. 

 Devices that do not have Failover active will not have any additional commands run.

Both have been tested on various ASAs in production environment.

### DIFF
--- a/lib/oxidized/model/asa.rb
+++ b/lib/oxidized/model/asa.rb
@@ -25,10 +25,16 @@ class ASA < Oxidized::Model
     @is_multiple_context = cfg.include? 'multiple'
   end
 
+  # check for failover environment
+  cmd 'show failover' do |cfg|
+    @is_failover_environment = cfg.include? 'Failover On'
+  end
+
   cmd 'show version' do |cfg|
     # avoid commits due to uptime / ixo-router01 up 2 mins 28 secs / ixo-router01 up 1 days 2 hours
     cfg = cfg.each_line.select { |line| not line.match /(\s+up\s+\d+\s+)|(.*days.*)/ }
     cfg = cfg.join
+    cfg.gsub! /(Configuration last modified by) (\S+) at (\d+:\d+)(:\d+\.\d+) (.*)/, '\1 \2 \3 \5'
     comment cfg
   end
 
@@ -36,7 +42,19 @@ class ASA < Oxidized::Model
     comment cfg
   end
 
+  #
+  # Capture failover items if we are Failover On
+  #
   post do
+     if @is_failover_environment
+        comment collect_failover_info
+     end
+  end
+
+  post do
+    #
+    # Determine configuration gathering for contexts
+    #
     if @is_multiple_context
       multiple_context
     else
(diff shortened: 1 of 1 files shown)


pv2b:
Before looking at the report, I already imagined the edge case where it would jitter at a minute boundary, which @jsynack already mentioned in #741:

>While there is the edge case that a jittering second could 'flip' the minute - it is probably not worth the extra code to handle that case.

If we want to go even further down, imagine saving a configuration exactly on midnight on Jan 1 of a year. Then you'd get jitter on every single component of the date!

After thinking about this, I come to the conclusion that it's probably mathematically impossible to completely handle this bug in the Cisco code without looking at the previous commited output.

If this bug has been fixed in more recent Cisco software - perhaps the "correct" fix is to have a flag to completely omit this date? Even so, this will still work fine approximately 98.3% of the time so it's probably better than nothing.

jsynack:
Great points @pv2b  - 

In general my thinking was to solve for the cases that I had seen with my installed base of ASA units that had experienced this issue ("more than 500 - less than 1000") .  I had not seen a minute jitter in the sample size I had.

There is certainly the ability to chop out the timestamp completely - it was an attempt to have a compromise and retain *some* information from it :)   Didn't want to spoil it for everyone unless it was a much larger issue. 

Since it was an easy adjustment that we had made locally I didn't collect any more stats on it. There doesn't seem to be much other report of it from the oxidized community - and the software revs are older where we have seen this The defect is slowly going away.

I thought about 'checking' the device as you describe - execute the command multiple times and observing the result - but opted not to do it to avoid punishing the majority of units that don't have this problem and punishing the AAA server with more commands to authorize per `interval` :)

laf:
I'll get this tested pretty soon. Sorry for the delay @jsynack and thanks for the PR though.

laf:
Jitter correction works fine from what I can see.

I'm going to let the failover side run for a bit longer to see if it generates noise but so far seems pretty good as well.

laf:
@jsynack So I've tested the failover command being in it as well and it does flip between not being able to get the state and working fine. I don't think that's an issue on our setup and happens across a good number of devices.

Have you seen any issues on your side?

jsynack:
@laf - So you see differences/diffs between different oxidized runs on the device as it can get the failover info and not get the information on other runs? (information churn).

We have not seen any problems/churn and have been running with this deployed as our `asa.rb` since the original PR timeframe - Roughly mid Nov-2017.  As a percentage of our ASA deployments - HA/FO is not that large as compared to other vendors/models - so it may simply be that we do not have the critical mass with ASAs or version-specific items.

Do you have any AAA logs of what commands get run or any debug log of the session?  Does it work clean on any units ? Break always on specific units?

laf:
@jsynack I'll have to do some more testing to find out. Leave it with me.

wk:
@jsynack is this PR abandoned? Can it be closed?

jsynack:
I am just about to get our instances caught up to `master` in the next few days - and I will see where things land and give it another go. For cleanliness - this can be closed out and I will open a new one for any changes.

mortzu:
Can you please fix the conflicts

PR #1107, opened 2017-11-18
Misc Doc updates

jsynack:
A few misc documentation updates that may be helpful to other users.

This also starts a framework of having docs/Model-Notes/<filename>.md for model-specific notes with the linkage from the README.md file.


docs/Sources.md
```diff
@@ -72,6 +72,36 @@ source:
...
+```sql
+query: "SELECT * FROM nodes WHERE enabled = True"
+```
+
+Since this is an SQL query, you can also provide a more advanced query to assist in more complicated oxidized deployments. The exact deployment is up to you on how you design your database and oxidized fetchers.
+
+In this example we limit the nodes to two "POPs" of `mypop1` and `mypop2`. We also require the nodes to have the `enabled` boolean set to `True`.
+
+```sql
+query: "SELECT * FROM nodes WHERE pop ('mypop1','mypop2') AND enabled = True"
```
pv2b: I think the correct SQL syntax is:

SELECT * FROM nodes WHERE pop IN ('mypop1','mypop2') AND enabled = True

(Missing IN)

laf:
Thanks for the doc updates :)

docs/Sources.md
jsynack: Thanks! Good catch. Absolutely required 'IN' - thanks for the fixup.  I had originally done the example as multi-POP. Then thought to make it simpler and started to do the example as a single-POP example. Then decided to go back to multi-POP as I only wanted two query examples as not to break the flow of the `Sources.md` (probably start to split them out to `Sources-<source>.md`).
 
So this is what happens when you change your mind two times mid-query :)

PR #1110, opened 2017-11-19
Initial checkin of PanOS API support

pv2b:
My second attempt at closing #440. This code actually works, unlike the approach that I tried for #1103 which proved to be impractical and problematic to implement.

No generic implementation of http unfortunately, but I can't see how that would work for this use case.

### DIFF
--- a/lib/oxidized/input/panos_http.rb
+++ b/lib/oxidized/input/panos_http.rb
@@ -13,9 +13,7 @@ module Oxidized
       @node = node  
       @node.model.cfg['panos_http'].each { |cb| instance_exec(&cb) }  
       @log = File.open(Oxidized::Config::Log + "/#{@node.ip}-panos-http", 'w') if Oxidized.config.input.debug?  
-      # Don't use @node.ip here, because this will make Net::HTTP not send a  
-      # correct HOST header.  
-      @http = Net::HTTP.new(@node.name, vars(:https_port) || 443)
+      @http = Net::HTTP.new(@node.ip, vars(:https_port) || 443)
       @http.use_ssl = true  
       # TODO make this configurable, this is not desirable for every environment
       @http.verify_mode = OpenSSL::SSL::VERIFY_NONE
--- a/lib/oxidized/input/panos_http.rb
+++ b/lib/oxidized/input/panos_http.rb
@@ -1,0 +1,77 @@
+module Oxidized
+  require 'net/http'
+  require 'net/https'
+  begin
+    require 'nokogiri'
+  rescue LoadError
+    raise OxidizedError, 'nokogiri not found: sudo gem install nokogiri'
+  end
+
+  class Panos_HTTP < Input
+
+    def connect node  
+      @node = node  
+      @node.model.cfg['panos_http'].each { |cb| instance_exec(&cb) }  
+      @log = File.open(Oxidized::Config::Log + "/#{@node.ip}-panos-http", 'w') if Oxidized.config.input.debug?  
+      # Don't use @node.ip here, because this will make Net::HTTP not send a  
+      # correct HOST header.  
+      @http = Net::HTTP.new(@node.name, vars(:https_port) || 443)
+      @http.use_ssl = true  
+      # TODO make this configurable, this is not desirable for every environment
+      @http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+
+      apikey_response = apicall(URI.encode_www_form({
+        :user => node.auth[:username],
+        :password => node.auth[:password],
+        :type => 'keygen'
+      }))
+
+      status = apikey_response.xpath('//response/@status').first
+      if status.to_s != 'success'
+        msg = apikey_response.xpath('//response/result/msg').text
+        disconnect
+        raise OxidizedError, ('Could not generate PanOS API key: ' + msg)
+      end
+      @apikey = apikey_response.xpath('//response/result/key').text.to_s
+
+      connected?
+    end
+
+    def connected?
+      @apikey != nil
+    end
+
+    def cmd(params)
+      apicall("#{params}&key=#{CGI::escape(@apikey)}").to_xml(:indent => 2)
+    end
+
+    # I do not understand these two methods, I just copied them from ftp.rb
+    def send my_proc
+      my_proc.call
+    end
+
+    def output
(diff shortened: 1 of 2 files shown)


laf:
@yeled What do you think on this one? Seems very specific but it does look like it could be tricky to make it very generic for different vendors.

Maybe it's worth just having it as part of the model rather than an input otherwise we could end up with a huge range of inputs before you know it?

ytti:
I do like the direction of the original attempt more. If we can't do it, via ```cfg``` or otherwise to be generic. Then we have some more fundamental problem in how models work, and we should extend the models infrastructure to allow this use case.

pv2b:
Unfortunately the original attempt ended up creating backups that were not restorable (and therefore was only usable for change tracking, not for backup/restore). So we have to use HTTP api.

I could try refactoring this again and simply not use a model at all (or maybe use a "dummy model"). In this case I think that adding another layer of abstraction doesn't really make the code more clear.

rocketnateynate:
Is this in the works still? Seems like a major thing if you can't use the backup to restore the device.

pv2b:
I'm running pretty much the code that's in this PR in our deployment. It works for me. I'm not likely to put much more effort into this, since my particular itch is scratched for the moment.

wk:
Having per-model http implementations and depending on additional 3rd party libraries would increase the hurdle for both users and future developers. In its current form, it may not be perfectly fit for inclusion.

Since between the time this PR was created and today `Oxidized` got an [`http`](https://github.com/ytti/oxidized/blob/master/lib/oxidized/input/http.rb) input, it would be great if this model was re-based on it for inclusion - on quick inspection, it seems much of the functionality at that level is already provided by `http`. 

Parsing XML is a more challenging task. There, we'd need to decide if we wish to include `nokogiri` (or an alternative such as `rexml`) as an explicit dependency (as we do want users to be able to use all the models provided 'out of the box') - in which case the dependency should be added to the spec file, `Dockerfile`, and installation instructions if necessary; or the model could attempt to mimic the parsing with regexp in order to avoid having such a dependency in the first place.

pv2b:
I could see if I could take a crack at refactoring this with the new http model. As a beginner level Ruby programmer (it's not a language I ever use), I really wouldn't know how to make an architectural change like that to the model, but now that that hard part's done, perhaps I could finish this up by myself.

As for parsing XML with regex, I really don't think that's a good idea. XML is an extremely complex standard, that does not lend itself to being parsed by regex in the general case. That means that even a minor any software update to the firewall might break existing code in Oxidized, if something about how the XML is emitted by the firewall changes, even though the actual document remains the same. Worst case, that can even break silently.

Aside from this model, there are at least two other models that parse and manipulate XML today, namely the OPNsense and pfSense models. By coincidence, I happen to use those two models as well. If an XML parser were included globally in oxidized, I could refactor those models to use that XML parser as well.

As for Nokogiri vs rexml, I have no strong opinion on this, because of my lack of experience with Ruby in general and XML parsing libaries for Ruby in particular. When I initially programmed this, I was very surprised that Ruby doesn't have a library for this out of the box. I can't remember my exact thinking now, but I likely picked Nokogiri because I read somewhere that it was the most popular XML parser for ruby. That in itself is probably a good reason to pick Nokogiri, it's a good chance it's already installed. Another good reason is that the current code already uses Nokogiri, which means I don't have to change that part.

I'm fairly sure this module won't be the last time someone needs to parse XML. In fact, searching over the codebase, I find that the OPNsense and pfSense models also use XML, and currently use Regex to manipulate the configuration files.

Anyway, as I said, my particular itch is scratched by the state of the commit as it looks now. I'll be happy to attempt to refactor the code to use the http model of Oxidized, but I'm not going to write code that parses XML using regex, because therein lies madness.

wk:
Upon deeper review, the http input appears to optionally load `mechanized` which in turn depends on `nokogiri`. Since the `Dockerfile` provides mechanized, the entire dependency chain on that side is already met. 

On manual installs, when `mechanized` is not available and the http input is loaded, a message is emitted instructing the user to install it - so it seems the question of library choice and availability appears to already be resolved.

Porting this model over to rely on the http input should basically do the trick, with XML parsing left intact.

reubenvirtuas:
Hi I am a bit of a beginner to Oxidized and I am trying to get it working with exporting Palo Alto XMLs. I have added the panos_http.rb and panos_api.rb files in the proper locations, but I am a bit confused as to how I am supposed to format my config and router.db files in order to work with this. Any help would be appreciated, thanks!

michael-ironnet:
Hello All, sorry to drag this up again. I'm relatively new to oxidized, however am interested in getting this to work. I have access to a lab Palo Alto device which currently does export a config in json. I would like to export a config in XML so it can be imported. 

Was this fix ever made functional? I didn't see it in the master branch. Appreciate anyones help with this!

RBVirt:
I've had good success using this - https://github.com/pv2b/oxidized/tree/panos-apiinput-model/lib/oxidized. I think its just two files you need to copy over.

michael-ironnet:
> I've had good success using this - https://github.com/pv2b/oxidized/tree/panos-apiinput-model/lib/oxidized. I think its just two files you need to copy over.

Thanks for the response! I did find this file https://github.com/pv2b/oxidized/blob/panos-apiinput-model/lib/oxidized/input/panos_http.rb, which was the other file that is required? 

Also, do you know where I would need to insert these files and what the config would be like? Currently I'm making use of groups in oxidized to make this work for the credentials, and I have it pulling devices from librenms and remapping the group variable based on the OS detected. If the API key simply needs to be inserted into the oxidized config file for the group, that would make this fairly simple. 

Appreciate anyones help.

RBVirt:
Should be this one https://github.com/pv2b/oxidized/blob/panos-apiinput-model/lib/oxidized/model/panos_api.rb. Once you have that you need to edit the config file and add  the panos_http file to the input section. I just added apikey as an input to my router DB file and added the API key there.

michael-ironnet:
Thanks so much for your help, going to attempt this on my lab in the next week or so. Getting busy with other projects. Appreciate you taking the time to point those files out. Now just gotta figure out how to get an additional input to function with my Librenms automation.

pv2b:
Closing this pull request in favor of my new, fresh PR #2360

PR #1111, opened 2017-11-20
Add support for Opnsense

pv2b:
This is just a very small change to the pfsense model to make it compatible with opnsense (which is a fork of pfsense - https://opnsense.org/)


laf:
Can you update the https://github.com/ytti/oxidized/blob/master/docs/Supported-OS-Types.md as well please :)

Issue #213, opened 2015-10-07
Auto enable like RANCiD ?


FTBZ:
I didn't find a way to not provide the enable password in oxidized like it's in RANCiD? There's a way to do that?

ytti:
Not sure I understand. Just not configure enable and it won't be used.

FTBZ:
Sorry I'm wrong. I thought that my user was in "auto enable" on the Cisco device but it's not the case. I'm using a specific set of "privilege exec" for this user. This role based method is working with RANCiD and seems to be not working with Oxidized.

I noticed that the RANCiD "rancid.types.conf" is taking about this and uses the full command as workaround (show running-config view full).

An idea if I need to create a custom model or if it's something that will be fixed?

ytti:
I'm not sure I fully understand your problem. If you're using RBAC you can allow 'show run' command, and no parameters should be needed.

If there is particular reason why this should work, we probably could hack something in the model to support both use cases. Maybe something in line with


cfg @@sh_run do |cfg|
   #parse_show_running_config
end

````
cfg :telnet, :ssh do
  @@sh_run  = "show running-configuration"
  @@sh_run += " view full" if vars("some_variable_blaa_blaa")
  ..
....
````

FTBZ:
Based on my experience with RANCiD, when using RBAC it's not enough to allow only "show run". It will only loads these informations without "enabled mode":

>Current configuration : 183 bytes
>!
>! Last configuration change at 06:45:26 MET Thu Oct 8 2015 by p22893
>! NVRAM config last updated at 17:10:21 MET Wed Oct 7 2015
>!
>boot-start-marker
>boot-end-marker
>!
>!
>!
>!
>!
>!
>end

I'm not a code expert, but I will test your workaround.

ytti:
When you login to the box normally with the credentials you have, and you run 'show run', do you get full configuration? If not, then we need to change the code or you need to change the config. 

I know that you can configure IOS RBAC so that 'show running-config' will display full config.

FTBZ:
Nope, with my settings the "sh run" only shows the configuration pasted in the preceding post. I need to specify the full "show running-config view all" to get the configuration.

I appreciate the time you take, I will promote this project on my side.

ytti:
Can you show me your RBAC configuration?

FTBZ:
[...]

PR #1115, opened 2017-11-28
model/ios: handle RBAC-enabled IOS shells

costasd:
RBAC-enabled shells and users in IOS, need show running-config view full in order to return meaningful results.

As not all IOS versions support this command, run view full version first, and if this returns less than 5 lines, (ie \n% Invalid input detected\n\n), fallback to get output from 'show running-config'.

Code was proposed in ytti/oxidized#213, I'm just testing and creating the request.

Testing:
- Works as it should (view full output wins) with my RBAC-enabled cisco shell and oxidize user being a non-privileged user.
- As I currently don't have any non-RBAC cisco shells to test the fallback, did the following to test it:
  - changed the command to `show running-config view fall` instead of `full`
  - this behaves as `show run view full` would in non-RBAC equipment (ie. "% Invalid input" etc.).
  Code fallbacks successfully  to `show run` command output in that case.

Thanks!

Fixes: #213

### DIFF
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -115,7 +115,23 @@ class IOS < Oxidized::Model
     comment cfg
   end
 
+  cmd 'show running-config view full' do |cfg|
+    @running_full_config_rbac = process_running_config(cfg)
+  end
+
   cmd 'show running-config' do |cfg|
+    # if running-config view full returned actual output and
+    # not the 3-line "Invalid input detected....", accept it.
+    if @running_full_config_rbac.lines.count > 5
+      ""
+    else
+      cfg
+    end
+  end
+
+  # add an intermediate method to process running config
+  # in order to control running-config variations
+  def process_running_config cfg
     cfg = cfg.each_line.to_a[3..-1]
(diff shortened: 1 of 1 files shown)


laf:
@FTBZ Are you able to test this please and confirm it works for you?

FTBZ:
At this time it's working fine 👍 Not sure, but seems more longer to poll all devices.

laf:
> This is out of the scope of this PR, but the show vtp status is a pain. Every time that a change occurs in vlans all switches return a new configuration.

Yeah we get the same. I've binned off the vtp stuff from the ios model in our install

ytti:
This would leave running-config unprocessed on non-RBAC nodes.

laf:
@costasd Any comment to the last reply?

costasd:
Hi there,

Unfortunately I don't have a non-RBAC node currently, I'm trying to find one to confirm ytti's comment, just freeze it for now.

Thanks!

dalamanster:
Or there is option to specify RBAC boxes manually via vars. 
@ytti proposed it [here](https://github.com/ytti/oxidized/issues/213)

mortzu:
duplicate of #1841

Issue #1058, opened 2017-10-03
Adtran Operating System


ja-frog:
I created a simple class for handling the Adtran Operating System (AOS). This is to pull configs from our Total Access 5000 and NetVanta equipment. Two kind of minor issues popped up:

1. What to call it for a pull request? AOS is already taken for Alcatel. Perhaps Adtran_AOS?
I might be going out on a limb in assuming that anyone would want to fold it in.
2. As part of the captured output, I am getting the initial command at the top, show running-config, without the prompt preceding it. Also, I get the prompt at the very end. What could be causing it to capture these as well?

jsynack:
(disclaimer: not official policy of oxidized, just generic thoughts)

1. Whatever is unique enough not to stomp on other vendors but applies to the product families that it is supposed to cover.  Ruby naming convention is to CamelCase Class names that are more than a single word. So for example, `AdtranTotalAccess`, `AdtranNetVanta` or just `NetVanta`. 

2. There is some discussion on other issues that this takes place sometimes when the wrong methods are used when interacting with the device - for example, using `send` versus using `cmd`.   The best way to get debugging help on the code is to let others see it - it will probably be a quick fix. As always - start from a working model is always good - especially one that is fairly mainstream or that you have in your environment so you can adjust or reference what is going on.

Including debug logs (with setting `debug: true` in your configuration as well as your input method) can also be very helpful.  This is important since due to the vast array of devices out there - not everyone has spares or access to them. If you can provide remote access during the development process it can also go a long way to getting a model developed properly.

laf:
> What to call it for a pull request? AOS is already taken for Alcatel. Perhaps Adtran_AOS?
> I might be going out on a limb in assuming that anyone would want to fold it in.

AdtranOS maybe seeing as the A in AOS means Adtran?

> As part of the captured output, I am getting the initial command at the top, show running-config, without the prompt preceding it. Also, I get the prompt at the very end. What could be causing it to capture these as well?

Prompt not correct?

Might be worth you submitting the new model as is and people can comment against it.

laf:
No response on the proposed PR which has been closed. Please feel free to comment on it and we can re-open the issue or the PR.

wk:
An Adtran model has been introduced in #1576

PR #1122, opened 2017-12-04
Create adtranos.rb

ja-frog:
While this works, it is including the command at the start of a config pull and the prompt at the end in the output.

Used NetVanta and TA5000 for testing. Version 9.7 #1058


laf:
You need to update the models doc: https://github.com/ytti/oxidized/blob/master/docs/Supported-OS-Types.md

Sounds like the prompt detection isn't 100% if you see the prompt on the output.

laf:
@ja-frog Any update on this one?

laf:
@ja-frog Feel free to push an update for the docs and we can re-open and merge this in.

PR #1149, opened 2018-01-16
New FTP option - Passive

cppmonkey:
Working with Xyzel XGS4600 series, they require passive to be set to false.

To make this possible, the ftp session is established after setting the passive setting.

The default is true, keeping the behaviour unchanged for anyone else using FTP to retrieve files.

### DIFF
--- a/lib/oxidized/input/ftp.rb
+++ b/lib/oxidized/input/ftp.rb
@@ -13,7 +13,9 @@ module Oxidized
       @node       = node
       @node.model.cfg['ftp'].each { |cb| instance_exec(&cb) }
       @log = File.open(Oxidized::Config::Log + "/#{@node.ip}-ftp", 'w') if Oxidized.config.input.debug?
-      @ftp = Net::FTP.new @node.ip, @node.auth[:username], @node.auth[:password]
+      @ftp = Net::FTP.new(@node.ip)
+      @ftp.passive = false
+      @ftp.login  @node.auth[:username], @node.auth[:password]
       connected?
     end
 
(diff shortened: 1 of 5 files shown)


lib/oxidized/input/ftp.rb
```diff
@@ -32,15 +28,6 @@ def cmd file
       @ftp.getbinaryfile file, nil
     end
 
-    # meh not sure if this is the best way, but perhaps better than not implementing send
```
laf: How come the rest of this code block was removed?

lib/oxidized/input/ftp.rb
```diff
@@ -1,25 +1,21 @@
 module Oxidized
   require 'net/ftp'
   require 'timeout'
-  require_relative 'cli'
-
+  require 'oxidized/input/cli'
   class FTP < Input
     RescueFail = {
-      :debug => [
```
laf: I know it's mainly just comments that's been removed but any reason you've done that? It may be helpful to others when reading the code.

cppmonkey: They're doing nothing. they are commented out.

By all means leave them in. Personally didn't see much point them being there, if they are needed the user would still have to edit the code.

lib/oxidized/input/ftp.rb
cppmonkey: a) to mirrored tftp.rb
b) The send my_proc define isn't called anywhere and define output was removed as the output log was empty and could have been related. Still empty but its removal hasn't broken anything

cppmonkey:
I've refactored the code. I'll resubmit

PR #1151, opened 2018-01-17
Don't post diff to Slack if there are no diffs

### LINKED ISSUE #1150: Slackdiff hook throws error when there is no diff
Hello,

I started oxidized for the first time, using CSV source, and git output. I configured the slack hook, and I received the below error.

```
[2018-01-16T22:08:02.817068 #15] ERROR -- : Hook slack (#<SlackDiff:0x000055c35b519c98>) failed (#<TypeError: no implicit conversion of Symbol into Integer>) for event :post_store
```

I investigated a little bit and it seems when [this line](https://github.com/ytti/oxidized/blob/master/lib/oxidized/hook/slackdiff.rb#L36) runs, it throws a TypeError seen above

I used `diff.inspect` to see what was up and it showed just `no diffs` instead of a hash. This makes sense as the configuration hasn't changed from what was on the remote, however I did not even have `diff` turned on for the Slack hook.

Looks like there needs to be a check to only post when there is a diff.

yzguy:
If there are no diffs then do not try to post to Slack.

Fixes #1150

### DIFF
--- a/lib/oxidized/hook/slackdiff.rb
+++ b/lib/oxidized/hook/slackdiff.rb
@@ -30,14 +30,16 @@ class SlackDiff < Oxidized::Hook
         if diffenable == true
           gitoutput = ctx.node.output.new
           diff = gitoutput.get_diff ctx.node, ctx.node.group, ctx.commitref, nil
-          title = "#{ctx.node.name.to_s} #{ctx.node.group.to_s} #{ctx.node.model.class.name.to_s.downcase}"
-          log "Posting diff as snippet to #{cfg.channel}"
-          client.files_upload(channels: cfg.channel, as_user: true,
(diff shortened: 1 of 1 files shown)


laf:
I responded in your issue asking to see the config for you slack hook, would you mind posting it so we can see.

laf:
I'm not sure this might be the best place for this, it may be better suited before the hooks are called so we don't run post_store hooks if we have no diff. For now I'll merge this in though.

yzguy:
I agree. At least this fixes the bug at the moment. I don't mind looking into moving some validations back up into `lib/oxidized/hook.rb`.

laf:
Every help we can get is appreciated, especially with the code side of things :)

PR #1155, opened 2018-01-18
Added documentation for setting up JunOS with Oxidized

cppmonkey:
I've been testing this with an SRX220H2 and should be applying this to several EX and QFX switches and our MX series routers.

So far only the J2300 has shown any errors and that was due to _omit_ within the _show configuration | display omit_ command. Its running such an old version of JunOS the command doesn't exist.

It isn't extensively tested, to ensure its locked down 100%, but at least Oxidized won't be accessing with root privileges.
It was assembled using Junipers own documentation on their [knowledge base](## https://kb.juniper.net/InfoCenter/index?page=content&id=KB23038&actp=METADATA)

### DIFF
--- a/docs/Model-Notes/JunOS.md
+++ b/docs/Model-Notes/JunOS.md
@@ -1,0 +1,38 @@
+JunOS Configuration
+========================
+
+Create login class cfg-view
+
+```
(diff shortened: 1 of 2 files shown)


laf:
Thanks very much for contributing :)

PR #1161, opened 2018-01-22
FTP Passive/Active modes & Xyzel XGS4600 series Docs

cppmonkey:
Having refactored the code/docs.

The Xyzel XGS4600 series require an active FTP session to enable the config file to be pulled.

Unfortunately only admin can pull the config and thus a CLI version is in the works.

Hopefully the passive/active option will be helpful for other users

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -74,6 +74,15 @@ vars_map:
 ...
 ```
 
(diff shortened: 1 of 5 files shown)


laf:
Thanks for contributing this @cppmonkey, appreciate it :)

Issue #887, opened 2017-06-12
Oxidized "slows Down"


jasonyates:
I'm not too sure how to describe this issue. Every 1-2 months we see a problem where Oxidized simply stops backing up 90% of our devices. There is nothing in the log file that suggests a problem, the tool is not failing to login due to connectivity or credential issues, it simply doesn't try. Most devices in the GUI turn blue. It's almost as if it slows down and decides it can only back up 1 device every 5 minutes instead of running its usual threads. 

My polling configuration looks like the below & we have 750 devices:
interval: 3600
threads: 30
timeout: 45
retries: 2

I have a hook configured to push collection metrics in to InfluxDB and generate an alert when our success metric drops over a 15 minute period so the "slow down" can be visualized. We change from an average of 4 devices every minute to 1 every 5 minutes with no configuration changes on our side.

https://www.dropbox.com/s/12m79h1i4nfhkde/Screenshot%202017-06-12%2006.38.00.png?dl=0

Normal operation is restored by restarting Oxidized.

Thanks

ytti:
This might be bit of a pickle to debug. As you need to wait 1-2 months to get more data. For shit and giggles, if it's easy, try to update oxidized, ruby and kernel to latest and greatest and see if problem remains constant.

laf:
@jasonyates Have you managed to test updating the software like ytti has said?

InsaneSplash:
I have about 2000 devices at the moment and I find the same thing happen but after a much shorter time. I find that I need to restart the service regularly to keep it operational.....

jasonyates:
I have just updated to 0.20.0 ( from 0.19.0) along with updating Ruby.

I'll monitor for the next few weeks and see if we get the same slow down.

Is there any form of debugging I can enable to help diagnose?

Thaks

jasonyates:
So this still happens after the upgrade to 0.20.0, it actually died after only a few hours this time around.

I noticed the following in the logs right before it fell back to only running once every 5 minutes

W, [2017-09-11T09:02:51.983719 #27673]  WARN -- : 172.x.x.x raised Net::SSH::ConnectionTimeout (rescued RuntimeError) with msg "timeout during server version negotiating"
W, [2017-09-11T09:02:52.818668 #27673]  WARN -- : mydevicename status no_connection, retry attempt 1
I, [2017-09-11T09:08:23.954798 #27673]  INFO -- : Configuration updated for mygroup/mydevice

laf:
[...]

PR #1175, opened 2018-02-05
Added support for setting ssh auth methods

laf:
This reverts https://github.com/ytti/oxidized/pull/128 which will cause a breaking change for Arista EOS devices. However, `keyboard-interactive` shouldn't be in by default as it causes some models (at least `IOS` and `NXOS`) that fail auth to hang on an interactive prompt waiting for a password which never times out.

To compensate for that, in EOS you can change the ssh authentication mode as per the model specific notes or by enabling `keyboard-interactive` in the oxidized config.

I've also included a fix (but am happy to move this to another PR) for when jobs_done incremented past the total number of nodes. For this we now reset the counter back to 0.


laf:
I've rebased this now to fix the conflict.

yeled:
Is there an issue # for the hanging forever bug?

How about you ship with the feature that doesn’t change for people using `keyboard-interactive` and the people affected by the hanging bug can take advantage of your new setting to remove the login method that stalls.

also is there a # for the `jobs_done` problem?

laf:
@yeled Thanks for taking a look.

The issue for the hanging is #887.

I can definitely change the defaults to include it so it remains the same as is now. However the default originally was not to have it and it was added in to fix one model that was broken so it does seem overall better to revert back to how it was before. I'm not bothered which way we go.

No issue for the job_done issue, I just found it by adding loads of debug output whilst trying to find why Oxidized ground to a halt after a bit. I can split that out if needs be.

ytti:
Needs conflict resolution

laf:
Done and rebased.

I blame @wk for these :D

wk:
@laf laf/oxidized#1

laf:
@wk :)

Thanks for that. I'd actually just pushed: https://github.com/ytti/oxidized/pull/1175/commits/7701ee6ba338348acce08c8fc941244b909842cf

What's the difference out of interest - I'm not a ruby dev :)

wk:
@laf the original laf/oxidized#1 was submitted just before you've had a chance to fix the tests, so the remaining difference was just the syntactic %w :)

This now passes CI and can go in.

laf:
> @laf the original laf/oxidized#1 was submitted just before you've had a chance to fix the tests, so the remaining difference was just the syntactic %w :)
> 
> This now passes CI and can go in.

Thanks @wk. Appreciate the insight :)

Issue #1000, opened 2017-08-29
AOSW Model not pulling entire configuration of Instant Controller


tyler-8:
I'm using the `https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/aosw.rb` model and it doesn't appear to be grabbing the entire config. Looking through the diffs, it appears that it grabs varying levels of the configuration each time. Some times it'll make it further into the config than other times. I've verified there's no paging issue (`show running-config` just dumps the entire config in one screen). So I'm not sure what to tweak.

Is the session being closed too early? Is there a setting I can tweak that would improve this behavior? It doesn't appear to be a connection speed or stability issue either. I have other devices (IOS/Procurve) that are being pulled successfully from the same location, it's only the AOSW device. 

I'm happy to provide any other information to help narrow in on the cause. Thanks in advance!

laf:
You should enable debugging and post the output of the devices debug file.

tyler-8:
@laf Appreciate the reply. I turned on debug and the entire config was pulled via SSH and the session closed with an `exit` successfully. So it would seem that the config is not making it into the local git repo? If I download the config trough Oxidized it's incomplete however.

laf:
Can you post it? Just trim the config so no sensitive info is in.

tyler-8:
I can't I'm afraid (it's a dev environment and I can only access it through a funky web client with no copy/paste ability right now).

I have manually combed the debug output and aside from some parse errors for non-existent commands I don't see any errors. Attachment of the parse errors.
![aoswoutput](https://user-images.githubusercontent.com/17618971/29842187-17e8334e-8cd6-11e7-9fb1-23dfd78141ea.jpg)

laf:
Is that not the device telling you those commands aren't running?

tyler-8:
@laf The commands that return the Parse error are expected based on the AOSW model - it's just attempting them without failing the whole workflow, but the `show running-config` does return correct output as shown in the screenshot.

laf:
Ok but this isn't showing the oxidized part to see what's going on. Can you not run a cli pastebin client to upload the text?

tyler-8:
I'm using the debug config outlined here: https://github.com/ytti/oxidized#debugging and all of the files it generates only contain the SSH outputs from the devices, even the devices that are working correctly in oxidized. The last line I see on the debug file for example is `test_rtr#exit`. Is there another way to enable debugging for Oxidized's processing?

laf:
[...]

PR #1182, opened 2018-02-09
[AWOS] Make the space between # mandatory rather than optional in the promt

thanegill:
Per discussion in https://github.com/ytti/oxidized/issues/1000#issuecomment-363444488. Fixes: #1000


laf:
Stealing my fix :D Thanks for testing and submitting it, glad it's working for you all now.

PR #1183, opened 2018-02-12
Comnet switch model

jaylik:


laf:
Thanks for submitting this @jaylik. Can you please also update https://github.com/ytti/oxidized/blob/master/docs/Supported-OS-Types.md to include this new model.

jaylik:
Ok now - could you check and merge, https://github.com/ytti/oxidized/pull/1183/commits/24aaf2d44b10063ada4fa8dfe6c61a5c67cb70df

lib/oxidized/model/comnet.rb
```diff
@@ -0,0 +1,44 @@
+class Comnet < Oxidized::Model
```
laf: I've just had a look and Comnet is the vendor and provides lots of products. The model should be named after the firmware the devices run like IOS, JUNOS, etc.

Can you change the name to match the firmware, if they don't provide firmware versions then it will need to at least be a name to cover the models they support.

docs/Supported-OS-Types.md
```diff
@@ -60,6 +60,8 @@
    * [TMOS (8800)](/lib/oxidized/model/corianttmos.rb)
    * [8600](/lib/oxidized/model/coriant8600.rb)
    * [Groove](/lib/oxidized/model/coriantgroove.rb)
+ * Comnet
+   * [Swithes](/lib/oxidized/model/comnet.rb)
```
yeled: spelling

yeled:
> Ok now - could you check and merge, 24aaf2d

How are you getting this error?

laf:
> Ok now - could you check and merge, 24aaf2d
> How are you getting this error?

It would also be better to split that out into another PR to avoid confusion.

lib/oxidized/model/comnet.rb
laf: Ok but if this model doesn't cover the entire comnet hardware range of devices (not just switches) then we should consider a better name imho.

laf: @jaylik Any update on this?

laf:
@jaylik Any update on this?

PR #1188, opened 2018-02-16
Update br6910.rb

dhooper6430:
The BR6910 switch can be configured to display anything on the prompt, by default it's vty-, I have modified the prompt statement to allow for whatever the user has configured the prompt to be.

### DIFF
--- a/lib/oxidized/model/br6910.rb
+++ b/lib/oxidized/model/br6910.rb
@@ -1,7 +1,7 @@
 
 class BR6910 < Oxidized::Model
 
-  prompt /^Vty-[0-9]\#$/
+  prompt /^([\w.@()-]+[#>]\s?)$/
   comment  '! '
 
   # not possible to disable paging prior to show running-config


laf:
Any chance you can provide some examples this will match against?

laf:
@dhooper6430 Any update on this?

dhooper6430:
Hi Laf,

I stole the prompt from the existing ios.rb code.

On the Brocade 6910 switch the prompt is configurable, by default it's just vty-n#, if the operator chooses to configure it, it supports an alphanumeric string up to 255 characters.

From the Brocade documentation:

> prompt
> This command customizes the CLI prompt. Use the no form to restore the default prompt.
> Syntax
> prompt string
> no prompt
> string - Any alphanumeric string to use for the CLI prompt. (Maximum length: 255
> characters)
> Default Setting
> Console

https://www.brocade.com/content/dam/common/documents/content-types/configuration-guide/brocade6910-2400-configguide.pdf

laf:
Thanks @dhooper6430. Just wanted to make sure this covers the intended prompt :)

PR #1210, opened 2018-03-12
ArgumentError with msg "invalid byte sequence in UTF-8"

tobzsc:


laf:
> There are bunch of models we have which, after this PR, unnecessarily mangle invalid UTF8 bytes to something else.

@ytti Is that a concern? Should we have this as a flag to enable using vars()?

ytti:
I'm not sure. I mean it doesn't hurt anything to unnecessarily mangle the strings, I think. 

But maybe someone WANTS to enforce UTF8? So that their consumers further down the pipeline can work on assumption it's UTF8 clean?

So maybe it should be global configuration option? ```config_is_utf8: true``` and if set to non-default false, we set this out.b?

laf:
@tobzsc Would you mind updating this PR to take those comments into account?

Basically default will be as is but you can set config_is_utf8 to be false if you want the behaviour you see here.

tobzsc:
In my eyes this is good solution and should work fine.

laf:
@tobzsc Are you able to update this PR then?

tobzsc:
@laf I don't know how do edit the code so that the config option is accepted. Sorry.

ytti:
Untested, due to clinically lazy

manoafv:
Hello,
Any news on this issue please?

PR #1219, opened 2018-03-14
Document the process of creating and extending Oxidized models

wk:
Provide skeleton documentation and examples for creating and extending Oxidized models.


laf:
Wow this is awesome. Thanks @wk :)

Issue #1221, opened 2018-03-15
VyOS/Vyatta/EdgeOS expose secrets under show configuration commands


yzguy:
In https://github.com/ytti/oxidized/issues/1163 it was brought up that `show configuration commands` provides a better backup rather than `show configuration`, this change was merged.

Normally when you run `show configuration` it automatically hides all the secrets with `********`. However, when running `show configuration commands` it no longer hides these secrets, therefore exposing them in the backup.

This is not an issue for those who want that behavior, however if a user specified `remove_secrets: true`, they are still captured.

Files to add to

https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/edgeos.rb
https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/vyatta.rb

Non-exhaustive list to cover:

```
encrypted-password
plaintext-password
password
pre-shared-key
```

PR #1222, opened 2018-03-15
add removal of secrets in edgeos, vyatta

yzguy:
Fixes https://github.com/ytti/oxidized/issues/1221

There are probably more, such as the MD5 authentication for BGP, etc. I don't have that configured in my environment so I am unable to make sure it's right.

However, the ones below should cover the most high risk secrets

Tested on Ubiquiti EdgeRouter X and VyOS devices.


yzguy:
This will cover BGP MD5 passwords too, sample command would be like

`set protocols bgp 65500 neighbor 10.10.10.10 password`

So the password one will take care of it

PR #1231, opened 2018-03-17
Supermicro support refactor (extends #1036) 

wk:
_Change is largely cosmetic, but untested. Assumes correctness of base PR #1036._

* Renames `supermicro.rb` to `edgecos.rb` to reflect which os type is actually being supported.
* Introduces a backward compatibility shim `supermicro.rb` which links to `edgecos.rb`.
* Renames proposed `supermicro2.rb` from #1036 to `aricentiss.rb` to reflect which os type is actually being supported. No backward compatibility shim here as this support is brand new.
* Documents the modules supported by each of the two types.

### DIFF
--- a/docs/Supported-OS-Types.md
+++ b/docs/Supported-OS-Types.md
@@ -141,8 +141,7 @@
 * Supermicro
   * [SSE-G2252, G2252P](/lib/oxidized/model/edgecos.rb)
   * [SSE-G48-TG4, G24-TG4](/lib/oxidized/model/aricentiss.rb)
-  * [SSE-X24S, X24SR, X3348S, X3348SR, X3348T,
-  * X3348TR](/lib/oxidized/model/aricentiss.rb)
+  * [SSE-X24S, X24SR, X3348S, X3348SR, X3348T, X3348TR](/lib/oxidized/model/aricentiss.rb)
   * [SBM-GEM-X2C, GEM-X2C+, GEM-X3S+, XEM-X10SM](/lib/oxidized/model/aricentiss.rb)
 * Symantec
   * [Blue Coat ProxySG / Security Gateway OS (SGOS)](/lib/oxidized/model/sgos.rb)
--- a/docs/Supported-OS-Types.md
+++ b/docs/Supported-OS-Types.md
@@ -128,7 +128,8 @@
  * Siklu
    * [EtherHaul](lib/oxidized/model/siklu.rb)
  * Supermicro
-   * [Supermicro](lib/oxidized/model/supermicro.rb)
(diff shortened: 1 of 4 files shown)


wk:
@tobbez @danilopopeye Can you check if this is a sensible compromise for #1036 for you, test it if possible, and merge or close with PR?

tobbez:
@wk Please don't make actual code changes when you can't/won't test them. The `post_login` command in my pull request is correct (not a typo).

2978d671a195a024893150cdf27b8a136d18e003 needs to be reverted, because `pagination` is (mis)spelled as `pagignation` in the switch OS.

laf:
@tobbez why not just submit a pull request to resolve it? @wk is doing an amazing job of getting a lot of things tidied up and inevitably something will break along the way.

wk:
@tobbez @laf 

This is definitely not the case per the [manual](https://www.supermicro.com/manuals/network/SMC_switches_CLI_manual.pdf) nor per the latest firmware blob for actual model [SSE-G24-TG4](ftp://ftp.supermicro.com/Firmware/Switch/SBM-GEM-X2C_SSE-G24-TG4_SSE-G48-TG4/Release-1.0.18-15/) where the correct spelling `no cli pagination` is used.

You seem to be using the same model, `SSE-G48-TG4`, but an older version of the firmware. Could you try upgrading to `1.0.18` and see if that solves it for you?

I've also encountered a SuperMicro knowledge [base article](https://www.supermicro.com/support/faqs/faq.cfm?faq=17258) that uses the incorrect spelling `no cli pagignation`, but it is quite old, and does not clarify on which models or releases this incorrect spelling applies to.

Since this model supports 12 different devices, if this is reintroduced, it would be best reintroduced conditional on a firmware check in-model to only apply to models which are affected.

tobbez:
It's definitely the case for 1.0.16 on SSE-G48-TG4.

Upgrading is not an option, at least short term. The firmware image for 1.0.18 definitely only contains `pagination` though, so it seems they fixed it in 1.0.18 (or maybe 1.0.17 if it exists). Regardless, the model should support both versions.

Anyway, conditionally issuing one or the other won't be faster or use less resources than just issuing both unconditionally (https://github.com/ytti/oxidized/pull/1234).

wk:
@tobbez 

That's indeed what I suspect (a quiet change). I have validated this against the 1.0.18 image prior to introducing the change, but not against running equipment, hence the untested remark.

As SuperMicro seems to have removed the older version of the firmware from their site, I was unable to verify if the version you've developed the model against (1.0.16) legitimately contains this misspelling, or if this was a typo in the model.

So long as issuing the wrong version of the command does not generate a blocking error or prompt that has to be handled during pre-exec phase, the solution in #1234 seems perfectly pragmatic.

Thank you for working through this!

PR #1239, opened 2018-03-21
Update nxos.rb

uwerler:
Otherwise the running config will not be stored.


ytti:
gsub! returns the changed string on match, and nil on non-match. So apparently we are trying to remove string which does not exist, so what is the whole purpose of the line? 

I could as well be ```cmd 'show running-config'```

PR #1243, opened 2018-03-25
Rubocop

cppmonkey:
Could be used to help make formatting more consistent and hopefully help improve code quality.

A lot of things highlighted are down to style and everyone had their own style.

||/or &&/and !/not all function the same

It is also not always clear is assignments expressions in conditional inspections are intended. Easy to put = when you mean == and there are 12 instances

I do like some of the suggestions, such as the guard clauses, rather than the nested if statements improves maintainability/readability


.rubocop
```diff
@@ -0,0 +1,2 @@
+--rails
```
wk: AFAIK, oxidized is not a rails application. What is the reason this is enabled?

.rubocop
```diff
@@ -0,0 +1,2 @@
...
+--display-cop-names
```
wk: This is default in Rubocop 0.54.0, is there a reason to force it on?

.rubocop.yml
```diff
@@ -0,0 +1,12 @@
...
+
+AllCops:
+  Include:
+    - Rakefile
+
+StringLiterals:
+  Enabled: false
+
+LineLength:
+  Enabled: false
```
wk: Based on a cursory observation, oxidized has a preference for verbose style hash methods (in terms of instances in the codebase), perhaps something along these lines as well?

```yaml
# Stick to verbose until https://bugs.ruby-lang.org/issues/10177 is closed.
Style/PreferredHashMethods:
  EnforcedStyle: verbose
```

lib/oxidized/input/tftp.rb
```diff
@@ -28,14 +28,10 @@ def cmd file
...
       config.read
     end
-    
-    private
-    
-    def disconnect
+  
+    private def disconnect
       # TFTP uses UDP, there is no connection to close
-    ensure
```
wk: This is presumably restructured to prevent Rubocop from breaking, but might be better as a separate commit / issue (can't immediately see why the `ensure` clause was here in the first place). Perhaps something along these lines to make this a pure Rubocop implementation commit in .rubocop.yaml instead?

```yaml
AllCops:
  Exclude:
    - 'lib/oxidized/input/tftp.rb'
```

laf:
@cppmonkey Any response to the comments from @wk here?

@ytti any thoughts / input on moving forward with this once it's ironed out?

wk:
Now that a CI pipeline is available again via #1265, this can be integrated into the pipeline to run for PRs going forward, allowing both structurally cleaning out the current inconsistencies and preventing new ones over time.

Something along these lines can be added to Rakefile as part of this PR to enable this:

```ruby
# Integrate Rubocop if available
begin
  require 'rubocop/rake_task'

  RuboCop::RakeTask.new
  task(:default).prerequisites << task(:rubocop)
rescue LoadError
  task :rubocop do
    puts 'Install rubocop to run its rake tasks'
  end
end
```

PR #1250, opened 2018-04-05
Rework the Exclusions for FortiOS

vppencilsharpener:
I had a few more fields to be masked and figured I could make the model better as well.

I moved the DB version masking to the "get system status" section and modified the replacement text to clearly indicate what was replaced. 

After adding the new fields to be masked I figured the RegEx patterns could be made a little more generic to cover additional cases without the need to update the model every time so I gave it a go. I tried to explain what and why in comments to make it easier for future changes. 

What I found after doing this was that the old line was masking non-secret commands incorrectly. (Looking at it again, it might just be the change from `.*` to ` .+`, with a space, at the end that corrected this.) 
For example `set password-expiry-warning disable` was masked as `set password <configuration removed>` and `set password-renewal disable` was masked as `set password <configuration removed>`

The diff of the resulting config (old & new model) for my device is attached.

[DiffOut.txt](https://github.com/ytti/oxidized/files/1878399/DiffOut.txt)

Let me know if you have any questions, if I should do anything differently or if I can do anything better. 

Thanks for doing what you do. 
Also my second attempt at getting this working is going much better.


vppencilsharpener:
So you don't have to pull it out of the diff file. I tried this on a FortiGate 600D running version 5.4.6

wk:
@vppencilsharpener this looks good, but could use a rebase on the current master to resolve conflicts, and a removal of the commented out cruft (the previous version remains in git history, so there does not seem to be any advantage to having it in the model).

wk:
@laf @vppencilsharpener this has been cleaned up a little, re-based, and adapted to pass rubocop. Completely untested.

wk:
@benjaminAco since you've recently contributed to this model, also focusing on secret scrubbing - could you perhaps review and test these changes?

wk:
Specifically interested to know if we're accomplishing anything here beyond what's been done in #1345 - since regexp generalizations, while elegant, occasionally bring with them a risk of false positives.

vppencilsharpener:
The changed code removes only the sensitive portion of the configuration. Prior to this change, commands like `set password-expiry-warning disable` were being masked as `set password <configuration removed>`.  This means that the configuration output from Oxidized may not be complete or accurately reflect the running config of the device.

PR #1260, opened 2018-04-07
eliminate inverse methods from models

wk:
swap out blocks which inverse their method across all models to improve readability

Passes `Rubocop` with `Style/InverseMethods` defaults.


wk:
Updates `.rubocop_todo.yml` to eliminate relevant `Style/InverseMethods` exceptions.

PR #1261, opened 2018-04-08
standardize on verbose hash methods

wk:
While there is some [discussion](https://bugs.ruby-lang.org/issues/10177) about `has_key` and `has_value` methods, the bulk of oxidized codebase currently uses these verbose variants. This eliminates the few short variants currently in use to make the entire codebase consistent.

Passes `Rubocop` with:
```yaml
Style/PreferredHashMethods:
  EnforcedStyle: verbose
```


wk:
Updates `.rubocop_todo.yml` to eliminate relevant `Style/PreferredHashMethods` exceptions.

PR #1262, opened 2018-04-08
Disable paging for "show full-configuration" on FortiOS

asenci:
Disable output paging on FortiOS, speeding up the execution and avoiding job timeouts when using SSH.

New take on PR #894 without changing the device configuration.

### DIFF
--- a/lib/oxidized/model/fortios.rb
+++ b/lib/oxidized/model/fortios.rb
@@ -49,7 +49,7 @@ class FortiOS < Oxidized::Model
 
 cfg << cmd('end') if @vdom_enabled
 
-    cfg << cmd('show full-configuration')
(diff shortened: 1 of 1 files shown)


laf:
@asenci Will this work across all versions of FortiOS?

asenci:
@laf I have tested the command "show | grep ." from 5.0 onwards (5.0, 5.2, 5.4, 5.6 and 6.0) and it worked as expected.

I can run further tests against older versions if you like.

laf:
@asenci I'm not sure it's 100% needed, I just don't use these devices to know or be able to test.

asenci:
Just FYI, I have tested on version 4.0-MR2-P7 (oldest version I could find a VM image for) and it also works as expected.

Issue #1061, opened 2017-10-09
Class for handling models that do not request a username


ja-frog:
I am trying to create a class to retrieve the configuration for a Calix B6 device. These aren't prompting Oxidized for a username when it tries to sign into them, only a password. From reading #573, I gather that these can be handled by dropping the stanza for username from the class for the model. I tried this, as seen in the code below, but it still seems like it is waiting on a username. 

Would someone have time to provide some guidance?

Error message in server log: <IP> raised Oxidized::PromptUndetect with msg "Password:  not matching configured prompt (?-mix:^[\w .-]+(#|>) $)"

```ruby
class B6 < Oxidized::Model

  # Calix B6 Series

  prompt /^[\w .-]+(#|>) $/
  comment  '! '

  cmd 'show version' do |cfg|
    comment cfg
  end

  cmd 'show running-config'

  cfg :ssh do
    password /^Password:/
  end

  cfg :ssh do
    if vars:enable
      post_login do
        send "enable\n"
          cmd vars(:enable)
        end
      end
    post_login 'terminal length 0'
    pre_logout do
      send "exit\n"
      send "\n\n"
    end
  end

end
```

Fhajad:
SSH B6 should prompt you for a username (name: cli) regardless. Telnet just prompts for a password, skipping the username portion

ja-frog:
I'm sorry, I have been away for a while.

What you say is accurate when connecting from a standard telnet or SSH client. However, when I tail the debug log for Oxidized for a B6 I am only seeing a password prompt. I verified with a packet capture that Oxidized is indeed trying to connect with SSH and not telnet.

I tried turning off exec channels just in case that is the difference yet it had no impact.

laf:
Would you be able to put a device online for me to test?

Fhajad:
@ja-frog I'll try your class out on some of my blades to see if I get the same issue or not. What software version are you running for comparison sake? I'm in a mix of 8.0.30.0 and 8.0.40.0.

@laf I'll let you know if it fails for me as well, then I can provide a remote session for one from my lab. These devices are basically never on direct internet (They're POTs Blades, DSL/POTs, GPON FTTH, L2 Switches, TDM to MGCP, and active ethernet "switches"). Luckily they all backup the same.

ja-frog:
We are in the process of upgrading, so some are on 7.3.61.0 while others are running 8.0.30.0. I see the same behavior on both versions.

ja-frog:
[...]

PR #1266, opened 2018-04-09
#1061 Handling devices that only prompt for password via SSH

ja-frog:
For #1061 
An approach to handling devices that do not prompt for a username, only a password when connecting via ssh. The Calix B6 inspired this.


laf:
You might also need to rebase from master as the travis ci checks are currently failing.

laf:
@wk What do you think to this change?

wk:
This looks like an improvement that should not have any side effects.

In the broader context, it might be desirable to also handle the case where a username prompt has been configured and observed, but resulted in a successful login immediately after the username was sent without being subsequently prompted for a password. It might also make sense to see if standardizing on the same logic both here and in `telnet.rb` is possible.

ytti:
Or maybe rather:

```ruby
    def login
      match_re = [ @node.prompt ]
      match_re << @username if @username
      match_re << @password if @password
      match = expect match_re
      until match == @node.prompt
        cmd(@node.auth[:username], nil) if match == @username
        cmd(@node.auth[:password], nil) if match == @password
        match = expect match_re
      end
    end
```

To honour stated desire in model, so that we don't even try to do passowrd/user if they are not asked in model, to avoid working with broken models.

laf:
@ja-frog Mind adding the above in to this PR and give it a test, I'll do another test once it's ready.

ytti:
Anyone able to test this? I have exactly zero devices which use SSH but authenticate terminal. Pong @ja-frog

ja-frog:
I should be able to try this out tomorrow.

ja-frog:
@ytti @laf  I'm sorry for having such a long turn around. I tested this and it passed with all of our ssh devices. That would be the B6s along with Netonix and Cisco switches and router.

Issue #1265, opened 2018-04-09
Fix tests so that CI/CD pipeline via Travis CI works


wk:
Travis CI was recently reactivated after a few years of absence (the [![Build Status](https://travis-ci.org/Shopify/oxidized.svg)](https://travis-ci.org/Shopify/oxidized) badge in README dates to a two year old build). Currently, 8 tests fail due to having fallen out with current code and need to be fixed. This is also an opportunity to introduce #1243 into the CI/CD pipeline.

- [x] Fix [failing test](https://travis-ci.org/ytti/oxidized) in [spec/input/ssh_spec.rb](https://github.com/ytti/oxidized/blob/master/spec/input/ssh_spec.rb) (Resolved in #1267)
- [x] Fix [failing tests](https://travis-ci.org/ytti/oxidized) in [spec/githubrepo_spec.rb](https://github.com/ytti/oxidized/blob/master/spec/githubrepo_spec.rb) to account for new username handling model via rugged proc introduced in #1257 (Resolved in #1269, thanks @erik-smit)
- [x] Update [README.md](https://github.com/ytti/oxidized/blob/master/README.md) to reflect build badge from the correct account (Resolved in #1270)
- [ ] Introduce Rubocop into Rakefile after #1243

wk:
@ElvinEfendi @danilopopeye I'm a bit stumped at how to correctly express the Proc'ing call to :credentials through Rugged in mocha expectations in order to update the githubrepo tests. Could you perhaps take a look at the change introduced in #1257 and offer a hint?

wk:
The CI pipeline is whole again. Thanks @ytti, @laf, @erik-smit

> Finally, the demons are appeased with offerings, which include local spirits – arrak, palm wine (tuak), and white rice wine (berem); these are poured over the banana trunk. The demons are then sent home.
>
> -- Angela Hobart: Dancing shadows of Bali: theatre and myth. 232 pp., 40 plates. London and New York: Kegan Paul International, 1987

wk:
Closing this issue as resolved, #1243 which was originally in issue scope will be addressed separately if there is collaborator consensus on implementing it.

PR #1267, opened 2018-04-10
Update ssh_spec.rb to pass successfully

wk:
Part of #1265


wk:
And likewise to you, @laf :smile_cat:

PR #1268, opened 2018-04-10
model/netgear.rb support for older models

ZsZs73:
Hi,
At first, thanks for this nice piece of software.

This pull request will add support for older models with FW version 5.x.y.z, like GS110TP, GS748v4.
- accepts spaces in prompt as the default prompt is '(Broadcom FASTPATH Switching)' without quotes
- hides password hashes following 'encrypted' as the config line looks as follows:
`users passwd "admin" encrypted 0b4....5d3`
unlike the newer fw versions:
`username "admin" password 335....2df level 15 encrypted`
- expects either 'User:' or 'Applying Interface configuration, please wait ...' as request for entering username. The later looks strange, but that is how it is.

It has been tested with following devices:
GS110TP fw 5.4.2.30 (latest available)
GS748Tv4H1 fw 5.4.2.27 (not latest)
Original function tested with:
GS724Tv4 fw 6.3.1.16 (latest available)
GS748Tv5 fw 6.3.1.16 (latest available)

This is how I configured my Netgear devices with old/new FW in:
- oxidized config:
```
    map:
      name: 0
      model: 1
      username: 2
      password: 3
    vars_map:
      enable: 4
      telnet_port: 5
```
- router.db
```
switchOldFW:netgear:admin:adminpw::60000
switchNewFW:netgear:admin:adminpw:true:60000
```
Note that both older and newer fw versions need 'enable', but:
- the older model prompts for 'Password:' and expects empty password (enable='')
- newer model does not prompt for password at all. (enable=true)

### DIFF
--- a/docs/Model-Notes/Netgear.md
+++ b/docs/Model-Notes/Netgear.md
@@ -38,9 +38,9 @@ The main differences are:
 
 Configuration for older/newer models: make sure you have defined variable 'enable':
 - `'true'` for newer models
-- `''` empty strgng: for older models
+- `''` empty string: for older models
 
-One possoble configuration:
+One possible configuration:
 - oxidized config
 ```
 source:
--- a/docs/Model-Notes/Netgear.md
+++ b/docs/Model-Notes/Netgear.md
@@ -1,0 +1,66 @@
+Netgear Configuration
+=====================
+
+There are several models available with CLI management via telnet (port 60000), but they all behave like one of the following:
+- older models:
+```
+Connected to 192.168.3.201.
+
+(GS748Tv4) 
+Applying Interface configuration, please wait ...admin
+Password:********
+(GS748Tv4) >enable
+Password:
+
+(GS748Tv4) #terminal length 0
+
+(GS748Tv4) #show running-config
+```
+
+- newer models:
+```
+Connected to 172.0.3.203.
+
+User:admin
+Password:********
+(GS724Tv4) >enable
+
+(GS724Tv4) #terminal length 0
+
+(GS724Tv4) #show running-config
+```
+
+The main differences are:
+- the prompt for username is different (looks quite strange for older models)
+- enable password
+  - the older model prompts for enable password and it expects empty string
+  - the newer model does not prompt for enable password at all
+
+Configuration for older/newer models: make sure you have defined variable 'enable':
+- `'true'` for newer models
+- `''` empty strgng: for older models
+
+One possoble configuration:
+- oxidized config
+```
+source:
+  default: csv
+  csv:
+    file: "/home/oxidized/.config/oxidized/router.db"
+    delimiter: !ruby/regexp /:/
+    map:
+      name: 0
+      model: 1
+      username: 2
+      password: 3
+    vars_map:
+      enable: 4
+      telnet_port: 5
+```
+- router.db
+```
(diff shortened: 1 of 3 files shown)


ZsZs73:
I am pretty sure that it is intended to work like this. There are no debug messages on the console.
I think that the CLI over telnet port 60000 was not intended to be used publicly.
Actually the credit for this info should go to: https://idefix.net/~koos/newsitem.cgi/1334773743
I've just tried with a GS110TP after factory reset and it prints this strange message instead of "User:"

Following is the beginning of the switchip-telnet file from the log directory
```
Connected to 172.0.3.201.

(switchname) 
Applying Interface configuration, please wait ...admin
Password:********
(switchname) >enable
Password:

(switchname) #terminal length 0

(switchname) #show running-config
0x4e470x010x00GS748Tv4H1          5.4.2.27            0x000000000x00000000000000
! The line above is the NSDP Text Configuration header. DO NOT EDIT THIS HEADER
!Current Configuration:
!
!System Description "GS748Tv4H1"
!System Software Version "5.4.2.27"
```

laf:
Actually also confirmed it does as I have GS110TP.

laf:
So i've tested this and definitely gets access to the device but without specifying an enable password then it can't get to enter enable mode to get the config. How is it working for you?

```
Connected to 192.168.90.251.

(Broadcom FASTPATH Switching)
Applying Interface configuration, please wait ...admin
Password:********
(Broadcom FASTPATH Switching) >terminal length 0
                               ^
% Invalid input detected at '^' marker.

(Broadcom FASTPATH Switching) >show running-config
                                    ^
% Invalid input detected at '^' marker.

```

ZsZs73:
"without specifying an enable password" -> enabled=nil
As described in my original post: you should set the variable 'enable' to empty string `''`. See the config snippet and the relevant router.db part in my original post.

If I click on the hostname on the GUI, it says: `:vars=>{:enable=>"", :telnet_port=>"60000"}`
I guess it is `:vars=>{:enable=>"nil", :telnet_port=>"60000"}` for you.
- `enable='nil'` bypasses sending 'enable' command completely. I am not aware of any Netgear models not expecting 'enable' at all.
- `enable='true'` sends 'enable', but does not expect 'Password:'. This is for newer models.
```
Connected to 172.0.3.203.

User:admin
Password:********
(GS724Tv4) >enable

(GS724Tv4) #terminal length 0

(GS724Tv4) #show running-config
```
- `enable=''` sends 'enable' and replies empty string to 'Password:'. This is for older models, like yours.
```
Connected to 172.0.3.201.

(GS748Tv4) 
Applying Interface configuration, please wait ...admin
Password:********
(GS748Tv4) >enable
Password:

(GS748Tv4) #terminal length 0

(GS748Tv4) #show running-config
```
- `enable='something'` sends 'enable' and replies 'something' to 'Password:'. I am not aware of any Netgear models expecting password after 'enable'.

Note the strange versioning of the HW:
- GS110TP: runs FW version 5.x.y.z (old model)
- GS748Tv4: runs FW version 5.x.y.z (old model)
- GS724Tv4: runs FW version 6.x.y.z (new model)

ZsZs73:
I am not sure if it is enough to push them to my repo, but apparently it showed up here as well.

PR #1269, opened 2018-04-11
Update githubrepo_spec.rb to pass successfully

wk:
Adapts tests to take into account changes introduced in #1257


wk:
Thanks @erik-smit for the breakthrough pointer on how to express this in mocha. :+1:

wk:
Part of #1265

PR #1273, opened 2018-04-11
introduce rubocop (closes #1243)

wk:
Partial rework of #1243:
* Introduces Rubocop with barebone configuration into the pipeline
* Updates `mocha/mini_test` to `mocha/minitest`
* Excludes rather than resolves `lib/oxidized/input/tftp.rb` from Rubocop to restrict change to pipeline


wk:
PR to test effective build pipeline. Not meant for merging. Closing.

wk:
This dependency was in the original user request #1243 by @cppmonkey, and presumably was used by that user. This is a development dependency, and is not included in the as-shipped gem. 

The [tool](https://github.com/flyerhzm/rails_best_practices) itself appears to include `haml` code quality measurement capability (which I presume is the context in which the user made use of it, as no other capabilities seem relevant), and may currently be a better fit as a dependency for `oxidized-web`.

Removed as part of #1372.

PR #1276, opened 2018-04-11
Add keepalive => true to the ssh_opt.

jsynack:
This will help to detect SSH sessions that zombie out at the protocol level by sending SSH Keepalives on the session.

This has been tested in a production environment for 3-weeks.

This may be the reason that oxidized 'slows down' by having worker thread slots occupied by wedged and non-productive SSH sessions. Discussion in https://github.com/ytti/oxidized/issues/887

Even if this is not the core of the problem mentioned in #887 - it is a good thing to have active.

### DIFF
--- a/lib/oxidized/input/ssh.rb
+++ b/lib/oxidized/input/ssh.rb
@@ -28,6 +28,7 @@ module Oxidized
       ssh_opts = {
                 :port => port.to_i,
                 :password => @node.auth[:password], :timeout => Oxidized.config.timeout,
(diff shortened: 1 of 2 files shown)


laf:
We're running this now, will report back how we get on over the next few days. Our install would lock up within 24-48 hours normally

laf:
So it doesn't help us with the issue where all threads become locked I'm afraid but in itself it hasn't broken anything from what I can tell.

PR #1280, opened 2018-04-12
Added Cisco Spark hook

rgnv:


laf:
@rgnv using the list rooms api, is it the ID from the response you use in the space config variable?

laf:
@wk has two PRs in the queue to clean up the slack and xmpp hooks. Probably would be worth updating this hook to use the same format: https://github.com/ytti/oxidized/pull/1283

wk:
@laf A brush-up of this PR to pass rubocop cleanly is available to @rgnv as PR rgnv/oxidized#1, awaiting review and pull. Once that is processed by @rgnv, this should pass tests.

laf:
Thanks @wk. I've tried testing this PR as it stands and can't get it to send the messages through so just need to know how to get the space ID.

ytti:
If @rgnv doesn't fix the complains, I'm ok if someone else does PR on this with those fixed. I think it's good to have this in.

laf:
I'm happy to fix this but so far I haven't been able to get it to work.

@rgnv if you can lend a hand with testing we can move this forward.

wk:
This now passes CI, only requires human testing to confirm that following the instructions results in a working configuration.

ytti:
I assume OP has Spark and that it works for him, which is good enough for me. Next Spark person can report brokenness and we can address that :>

rgnv:
@laf  sorry, been bogged down at work. You can retrieve the Space/Room ID the quickest way by enabling Test Mode at https://developer.webex.com/endpoint-rooms-get.html and plugging in the Bot/OAUTH API key (bearer token) and hitting Run. By default, test mode will dump all rooms that token has access to.

As of last week, Cisco announced a name change of Cisco Spark to WebEx Teams, which will result in the api/endpoint on their end changing. The ruby api I'm using appears to be abandoned, so whenever the old API endpoint shuts down, this module will break. I'll handle the rebranding later this week, and see about forking the ruby api and updating it with the new Teams endpoint.

laf:
> @laf sorry, been bogged down at work. You can retrieve the Space/Room ID the quickest way by enabling Test Mode at developer.webex.com/endpoint-rooms-get.html and plugging in the Bot/OAUTH API key (bearer token) and hitting Run. By default, test mode will dump all rooms that token has access to.

Got it, tested and works perfectly. Thanks for the pr :)

PR #1289, opened 2018-04-20
remove .ruby-version & add many ruby-version in .travis.yml

yuri-zubov:
**Purpose**
I think better if you remove .ruby-version from the project and make testing for all ruby via travis.yml.
What do you think about it?

ruby-2.1 has EOL(https://www.ruby-lang.org/en/downloads/branches/)
In my opinion, ruby-2.1 can be deleted from .travis.yml

**Known Compatibility Issues**
https://github.com/ytti/oxidized/pull/1006


wk:
Test updates from #1006 are already in master (addressed in #1267). 

Does Travis CI have a Ruby 2.0 target and does build occur cleanly under it today? If it does, this could be an opportunity to include it as well - since that is how far the backwards compatibility commitment goes in order to be installable on system Ruby of RHEL 7.

wk:
Reviewed this in #1290 - Travis does have Ruby 2.0 target, but Rubocop dropped support for Ruby < 2.1 at 0.50.0, and there isn't sufficient payoff to justify maintaining two parallel Rubocop rulesets for legacy.

PR #1290, opened 2018-04-20
add ruby 2.0 to travis build

wk:


wk:
Rubocop dropped support for Ruby 2.0 since 0.50.0. Closing this as it would be better to walk away from the 2.0 legacy over time.

PR #1296, opened 2018-04-21
the great makeover - standardize layout, alignment, indentation

wk:
This is a large PR that standardizes layout, alignment and indentation throughout the entire code base. Performed using `rubocop --auto-correct`, with additional artisan fix-ups where necessary.

Cosmetics only, no functionality modifying changes. Due to the size of the PR, testing is welcome nevertheless.

PR #1297, opened 2018-04-21
Do Not Merge: the great makeover part ii

wk:
Builds upon #1296 but also applies lint and style rules, standardizing the entire code base around rubocop defaults using `rubocop --auto-correct`, with the exception of a handful of cases that will require manual refactoring.

Believed to be correct, starts and runs, but is absolutely untested and not reviewed by a human.

@ytti this is a sneak peak into what can be expected as an end result of the rubocop cleanup process.


wk:
While #1296 deals primarily with cosmetics, the lint and style rules of rubocop enabled here do introduce code changes which *may* have side effects.

Unless there's appetite for an act of extreme bravery (commit first, fix what breaks later), it's unclear if this portion of the project would be easier to review as one large change set, or to implement as dozens of discrete but manually reviewed fix-ups.

jsynack:
Greetings @wk -

My opinions only below - take them for that :) 

1.
 Oxidized doesn't have a well-defined release cycle (that I know of) and PRs sometimes take a bit to make it in to where the end-users can get to it. This increases the chance that some other smaller PR will get rolled in and start to break apart the diff'ing/merging of this 'big giant fixes' PR - it would at least start to require more admin overhead of deciding on what the cherry-pick or keep from what PR etc. 

This is a byproduct of the "maintainer wanted" status and just the way it is (that is not a dig on anyone or oxidized in general - just the facts. If we can hit critical mass of a few people - it may start to shape up a bit more by spreading the workload.

I am in favor (FWIW) of a fast-track to get it in. That is really the only way to get it in there without creating more administrative overhead of diff hell (the longer a big PR waits).  Sure - you can always rerun `rubocop --auto-correct` - but there are better things for your time if you are working on fixes.

2.
There are some popular models that have quite a bit of churn, and changes since the 0.21 release, or pending PRs - pending items that could be helpful to the user community - but the user community doesn't know how to run from the git repo or add a custom model to the oxidized dir. 

This is a documentation issue that can be solved with some more writeups I think, but care needs to be taken that versioning stays within a sane threshold of supportability. E.g. - once a user has a working custom model - it could drift from the official model over time and introduce other problems.

3.
 Most projects tend to do the big giant fixes like this after a release milestone - so that there is plenty of bake-in time until the next release. This is for the brave that are willing to run from the git repo / HEAD / -current / whatever - they/we deserve whatever pain we get from it ;) 

4. 
My personal thoughts on this type of PR - there are 3 ways:
4a. A PR to fix the specific Rubocop violation family (e.g. a PR for fixing RE perlisms) 
4b. A PR based on the files / models. For example - I use about 6 models very heavily - and I would be able to spot problems on them quickly.
4c. Just get it in in like this PR - and then see what breaks running that branch (*this is my personal pref. given the size/scope*)


TL;DR (but you had to read all this way to get to it)
I suggest that a new release of oxidized take place (0.22?) *without* this PR - as there are several items between 0.21 and now that are probably worth a release (perhaps a few of the smaller / lower hanging fruit PRs can get implemented).

Immediately after the release - this PR gets integrated on a different branch.  This allows the folks that 'know' how to use git and grab it to run it - and then at some milestone it gets merged to master and the next release cut (0.23) as the 'rubocop sanity' release.

This provides the end-user community with a stable-ish newer release (0.22) with various fixes/updates and an ability to address any items that crop up.

 The new branch then provides a way to start addressing some of the architectural / code standards that this PR gets into. And then merge back to master when things are tested a little bit.

I was looking at some of the rubocop items - and some of it is clearly code style changes that would need to be adhered to to prevent relapses (hence the CI) - so this also means docs and enforcement for any new PRs or models (since right now it is copy n paste from models that have violations).

Again - just my thoughts. Happy to help where I can. I can commit to running any other test branch and testing models I have access to as it is easy enough to run in parallel to a stable instance of oxidized.

Thanks BTW for getting all the rubocop/CI stuff going - it is helpful and will certainly pay off once the initial hurdles are crossed. 👍

wk:
@jsynack thanks for the review! It is indeed rather difficult to review large thousand-line PRs, both due to Github not offering really great tools to do so - and due to the change velocity of the underlying code base.

Huge PRs aren't great. I am reasonably comfortable with #1296 since while the change set is still much larger than desirable, all the changes are purely cosmetics of how the same characters are arranged in files, and in my assessment the risk of pulling it is relatively low. It could still benefit from a couple extra eyeballs (and executions) to make absolutely sure.

This PR, on the other hand, is meant more as a demonstration of what the final outcome of the entire "rubocopization" process looks like, how much work there is still to do, and to get the kind of feedback you've just offered. 

If a few key contributors and a collaborator sponsor are willing to meaningfully deal with the fallout of a single act of bravery in finite time, it is possible to go ahead with (something like) this pull. Scheduling a release first before potentially intentionally breaking the code base is indeed be the responsible thing to do.

Alternatively, anyone can still take an item out of `.rubocop_todo.yml` at any time, on best effort basis, and clean it out over time. The advantage of this method is that the emergent rule set might be better aligned with the specifics of this project than wholesale application of rubocop defaults, as the people who allocate a block of time to clean out an issue are hopefully better positioned to give it some thought and consideration before starting to type commands into the terminal... 

Personally, the 4a/4b approach seems best to me as it (should be) the least disruptive, and I fully expect *this* PR to be closed un-merged.

wk:
Example of merged PR in the style of 4a #1260, in the style of 4b #1283

laf:
Little point in me reviewing the code :)

Testing on my home install though, will leave it running for 24 hours but so far it's gone through one cycle with no issues.

PR #1304, opened 2018-04-24
implement probot no-response

wk:
Configuration for the [Probot: no-response](https://github.com/probot/no-response) GitHub app that automatically closes issues tagged with `user_pending` after 21 days if they haven't received any feedback from the original author.

Requires a collaborator to activate the GitHub App at https://github.com/apps/no-response

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,0 +1,13 @@
+# Configuration for probot-no-response - https://github.com/probot/no-response
+
+# Number of days of inactivity before an Issue is closed for lack of response
+daysUntilClose: 14
+# Label requiring a response
+responseRequiredLabel: user_pending 
+# Comment to post when closing an Issue for lack of response. Set to `false` to disable
+closeComment: >
+  This issue has been automatically closed because there has been no response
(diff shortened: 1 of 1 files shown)


wk:
@laf since you perform the bulk of the community moderation activities, this one is primarily for your consideration.

laf:
I'm all for this and have considered it before (we need to do it for LibreNMS).

Would just mean we'd have to make sure we dealt with user_pending issues when the user has replied with the info needed but I'm sure we can cope :)

Would need @ytti to activate though as I can't do that.

wk:
`no-response` is very reasonable about that, in that if the user responds, the `user_pending` label will simply get removed, and the issue will linger in the queue, awaiting its turn. If the user replies after the issue has been closed for inactivity, the issue will additionally be re-opened at the same time.

laf:
> no-response is very reasonable about that, in that if the user responds, the user_pending label will simply get removed, and the issue will linger in the queue, awaiting its turn. If the user replies after the issue has been closed for inactivity, the issue will additionally be re-opened at the same time.

Nice :)

Issue #611, opened 2016-11-08
HP Comware xtd-cli-mode Password


adamboutcher:
Hi,

As all the commands required for Oxidized to work are being some stupid password wall xtd_cli_mode, the HP Comware 1950 doesn't get its configuration backed up.

Some kind soul on Reddit has found the password [here](https://www.reddit.com/r/networking/comments/3phmat/hp_195048gpoe/cy3eywi/) (foes-bent-pile-atom-ship), I was wondering if you guys could add support?

Thanks

ytti:
It looks like the password maybe changing per firmware?

adamboutcher:
This model of switch HP Comware 1950 has always had a different password
although it looks like it's changeable per firmware, could we have this as
a variable?

Thanks

ytti:
We could have hash mapping firmware to password. Then query for version, and set password from hash based on returned version. 

But I don' really understand the problem even, I don't have these devices. Maybe @sid3windr @erJasp @amarti2038 or @nickhilliard can chime in.

adamboutcher:
There's a section for the 1910/1920 I believe but the command it initially sends is incorrect for the later models...

adamboutcher:
The line
https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/comware.rb#L32

is incorrect for this model switch, it should be

> send "xtd-cli-mode\n"

ytti:
@sid3windr comments about the cli-mode command?

adamboutcher:
I know it's been a while but do we have any news/update on this issue?

laf:
@adamboutcher Can we send both commands to catch the different models or will that cause issues?

adamboutcher:
If you send _cmdline-mode on to my model which uses xtd-cli-mode just ends up saying "% Unrecognised command found at '^' potision.", so you could check for that and then provide the correct command.

adamboutcher:
I can't test how the models that don't support the xtd-cli-mode command would act but I cant see why not.

laf:
Try it for you, add the command to the model (see https://github.com/ytti/oxidized/blob/master/docs/Creating-Models.md on how to do it in your own install). See if it works and then feel free to submit as a pull request. I'm not sure it's the way the project would want it to be done but let's see.

semplonius6:
[...]

PR #1305, opened 2018-04-25
Updated HPE Comware - Issue #611 Fix

adamboutcher:
This change works for newer HPE Comware model switches and *shouldn't* cause any issues with the previous Comware models although past functionality hasn't been tested as I don't have an older model.

### DIFF
--- a/lib/oxidized/model/comware.rb
+++ b/lib/oxidized/model/comware.rb
@@ -38,6 +38,9 @@ class Comware < Oxidized::Model
         send "_cmdline-mode on\n"
         send "y\n"
         send vars(:comware_cmdline) + "\n"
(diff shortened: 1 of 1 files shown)


wk:
@udha @Ultra2D @amarti2038 @nickhilliard @roock @sid3windr @erJasp 
As you've all contributed to this model, could you validate that this change isn't breaking for older HPE Comware devices?

nickhilliard:
this doesn't break oxidized on comware 5900 devices, but they don't need this hack to start with.

Ultra2D:
No problems on a 5700 (Comware 7).

wk:
Thanks @nickhilliard and @Ultra2D. This looks ready to pull.

PR #1307, opened 2018-04-25
Capability to pipe diffs into post_store exec hooks, misc bug fixes

gs-kamnas:

* Adds the optional capability to pipe the config diff to standard input of handler scripts/binaries in the exec hook.

* Adds additional counter variables to allow the config download success and failure count to remain accurate beyond  the 11th run.

* Postprocessing of NXOS configs to have consistent Unix style line endings, remove lines containing exclusively a prompt, and comment all non-configuration lines. This allows the configuration taken from Oxidized to be directly supplied to a switch to revert to a specific version.

* Documentation of changes to the exec hook.

### DIFF
--- a/bin/oxidized
+++ b/bin/oxidized
@@ -8,6 +8,6 @@ begin
   Oxidized::CLI.new.run
 rescue => error
   warn "#{error}"
-  debug = Oxidied.config.debug rescue true
+  debug = Oxidized.config.debug rescue true
   raise if debug
 end
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -18,7 +18,7 @@ Following configuration keys need to be defined for all hooks:
 
 ## Hook type: exec
 
-The `exec` hook type allows users to run an arbitrary shell command or a binary when triggered.
+The `exec` hook type allows users to run an arbitrary shell command or a binary when triggered. The config diff may optionally be piped to standard input of such command.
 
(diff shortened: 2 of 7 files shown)


ytti:
The hook assumes output is to git, shouldn't it test it? Shouldn't it ask if output responds_to diff? And maybe of clarity talk about 'diff' or 'outputdiff' not 'gitdiff' as it could potentially be anything?

ytti:
Could you make this three commits

  1. stats
  2. nxos (maybe add the bin typo fix here, if you don't want separate for that)
  3. diff-to-pipe

Thanks!

gs-kamnas:
@ytti Refactored as requested

wk:
@gs-kamnas these are some great enhancements! I'm happy to see the nxos model fixes and the run counters.

I think @ytti may have misworded the request, as I'm certain the idea is to get this as three separate PRs (rather than commits) so that each can be processed separately. Feel free to also update CHANGELOG.md separately within each PR adding a line for the feature. 

This also applies to ytti/oxidized-web#163 which will need to be updated to reflect which of the three future PRs (presumably, only the counters one) it depends on. It will also require a minor conflict resolution.

Sorry for the extra hoops! There's a lot of work happening right now to streamline the contribution-to-release process to make oxidized move forward faster and with more guarantees of correctness, and this occasionally places additional burden on contributors.

ytti:
Quite, apologies for asking the wrong shit :)

gs-kamnas:
Refactored into #1332 , #1333 and #1334 as requested.

PR #1316, opened 2018-04-27
document resolve_dns knob

wk:


ytti:
The branch is like 2 lines changes, 2 commit, 1 with typo. I think I can just rewrite it, if the feature is even needed.

PR #1318, opened 2018-04-27
Introduce resolve_dns knob (closes #599)

### LINKED ISSUE #599: SSH Proxy - DNS Resolution
Hi,

I've been looking at connecting to multiple sites via SSH proxys, rather than having an oxidized VM at each site.

The initial testing looks great, but i've noticed that DNS resolution of switches happens before connecting to the SSH proxy, rather than after, meaning that the oxidized server must be able to resolve the switch, rather than relying on the SSH proxy to do the resolution.

Is this something we can look at changing?

Thanks!
Nic

wk:
Based off the `no-resolve` branch that was created to address #599 but never had a chance to be reviewed and tested. Includes documentation for the `resolve_dns` knob and an additional case in the ssh spec that validates its behavior.

Completely untested.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -214,3 +214,13 @@ next_adds_job: true
 ```
 
 This will allow for a more timely fetch of the device configuration.
+
+## Disabling DNS resolution
+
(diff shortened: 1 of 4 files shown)


ytti:
yolo

Noname2244:
Is the function "resolve_dns: false" enabled ? ie could I just add that line to the conf file?

PR #1322, opened 2018-04-28
Update boss.rb

ospfbgp:
Issues with some older boss switches that have the banner turned off.  They still display the banner but it is just scrolls and no pause.   The send "c" gets put at the prompt and you are now stuck.  An easy fix is to just send new line so you get a new prompt and the next command does not get corrupted and the backup of the configuration can continue.  Tested on about 100 BOSS switches and all working now.  Also Extreme now owns the BOSS platform after being purchased from Avaya.


ytti:
It might be safer to guard ```send``` with some expect.

Issue #1313, opened 2018-04-27
Release 0.22.0


ytti:
1. get PR's in, at least selected ones
 2. update changelog
 3. set docker to build tags
 4. any issues that should be solved?

jsynack:
Was @laf still suffering from a worker deadlock issue?

laf:
@jsynack The PR I put in that fixes that for us has been merged so would be in 0.20.0.

That hasn't resolved some other issues for slow downs that other see according to those who responded.

ytti:
For better or worse, done.

PR #1326, opened 2018-05-02
add ability to build oxidized-web in docker container

yuri-zubov:


yuri-zubov:
@wk ok, I got your arguments.
what do you think about 
```
 # web interface
WORKDIR /tmp
RUN git clone https://github.com/ytti/oxidized-web.git
WORKDIR /tmp/oxidized-web
RUN git fetch --unshallow || true
RUN rake install
```
oxidized is not published often and I think it will better if oxidized-web is built from github repo.

wk:
I understand the goal here, but there are still a few issues with this approach:
* The oxidized/oxidized docker container build is automatically triggered by commits to `oxidized`, which means that changes to `oxidized-web` will not trigger a container rebuild. Instead, however, changes to `oxidized` might end up pulling two different versions of `oxidized-web`, leading to confusion as to the source of issues when upgrading from one container to another.
* In the future (starting from the next release) there are plans to also tag the auto-built docker containers, so that users preferring a stable version can use `oxidized/oxidized:stable` for the last release, and `oxidized/oxidized:vX.Y.X` for a specific release, and the `oxidized-web` included there is expected to be a known and released version, not master.
* This does not solve the issue of velocity of change to oxidized-web, as changes still have to make their way into the oxidized-web `master`

A better approach might be to try and bump the release velocity of `oxidized-web`. Some work toward that is taking place, and you'll notice some momentum toward a release is building up; there should normally be a new release of `oxidized-web` together with the next release of `oxidized` (0.22.0) currently being finalized on #1313, and additional work is happening (CI pipeline, ...) toward automating some of the integration and build work toward enabling more frequent releases.

wk:
@ytti @laf this could benefit another pair of eyes. How do you see the separation between `oxidized` and `oxidized-web` working out in the context of CI, composable docker containers, ...?

ytti:
Maybe we can guard this on environment variable? I'm with @wk that we may have use-cases which do not require oxidized-web, but then again, probably 99/100 installations use oxidized-web.

What are our options? What is idiomatic in docker world?

wk:
The idiomatic docker approach to this would require two separate, composed, containers.

To realize that, there are essentially two obvious ways to go:

* Implement real loose coupling between `oxidized` and `oxidized-web`. That is, bring the RESTful API into `oxidized`, while keeping the user facing haml and js in `oxidized-web`. This will allow shipping the two as two separate loosely-coupled (over HTTP) containers, as `oxidized` will provide the RESTful API, and `oxidized-web` will be nothing but a web-frontend with no dependency on `oxidized` itself, but only on a configurable API endpoint.
* Merge the two completely into a single `oxidized` repository, acknowledging that both are typically used as a single product, and build a single web-enabled container whenever that single repository changes.

Other options include:
* Import `oxidized-web` into `oxidized` as a [git sub-module](https://blog.github.com/2016-02-01-working-with-submodules/) and ensure the docker hub build pulls it.
* Accept this PR and let the chips fall where they may
* Increase release velocity of both gems so that a build of `oxidized` will have a reasonably current `oxidized-web` gem included
* Do nothing

laf:
> Merge the two completely into a single oxidized repository, acknowledging that both are typically used as a single product, and build a single web-enabled container whenever that single repository changes.

I'd definitely go with that option. I don't see the benefit of shipping -web separately. If people don't want to use it then can just disable in the config.

PR #1327, opened 2018-05-02
Quick-fix: repair some logic for @jobs_done.

jsynack:
This small adjustment changes to only increment `@jobs_done` for a successful pull of a node or when the retries are exceeded and the node is abandoned for that cycle.

Previously `@jobs_done` was incremented as soon as `process()` was called.

The problem is that this incremented `@jobs_done` before knowing if the node completes OK or fails (And requires a retry).

During a retry - the node to be requeued for processing - which would increment `@jobs_done` multiple times per node (up to the configured retries count per node for a downed node).

This causes `@jobs_done` to become out of sync with reality. One of the main impacts of this is when the `:nodes_done` hook gets called. This could cause the hook to fire mid-cycle and then not fire at the 'real' end of the interval which is the intent of` :nodes_done`. The next time it fires would be when the `@jobs_done` catches back up (in the NEXT cycle) to the `@nodes.count`.

### DIFF
--- a/lib/oxidized/worker.rb
+++ b/lib/oxidized/worker.rb
@@ -42,9 +42,9 @@ module Oxidized
       node.stats.add job
       @jobs.duration job.time
       node.running = false
-      @jobs_done += 1 # needed for worker_done event
 
       if job.status == :success
+        @jobs_done += 1 # needed for :nodes_done hook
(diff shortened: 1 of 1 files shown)


ytti:
This is proooobably ok :)

PR #1329, opened 2018-05-03
Feature - Lockout by IP address

jsynack:
This implement a simple lockout by IP address methodology that the end-users can do when needed. 

This checks for a file in the lockout directory based on the IP address of the node about to be polled.
Only single IPs are locked, there is nothing fancy such as CIDR-based locking (yet).

If the file exists, the node is skipped for that polling interval/cycle. 

This allows for things to be locked out (not polled) in the event the user knows a node is going to be unreachable for some time - but doesn't want the polling and retries to build up on the offline nodes.
This also allows other things (such as external NMSes) to trigger a lockout etc as long as they have read/write to the lockout file directory (controlled by `lockout_directory:` configuration directive).

There were multiple ways to implement this - this seemed like the easiest/safest way for ease of use. There could be some impact to how job timing is calculated, since the job for a locked node is over instantly. 

Since this introduces a new status of `:locked` - I also will send in a PR for oxidized-web to see this status and render things properly in the UI.

Comments/feedback welcomed. I have had this sitting for a while - and trying to get it eye-ball checked by others or any different directions/thoughts.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -214,3 +214,35 @@ next_adds_job: true
 ```
 
 This will allow for a more timely fetch of the device configuration.
+
+
+## Node Lockout / Node Skip
+
+Sometimes it is helpful to temporarily prevent oxidized from polling a node without reloading the node source list or restarting the oxidized process.
+
+This can be accomplished by creating a lockfile with the node IP address as the filename in the `lockout_directory`. By default this is `~/.config/oxidized/lockout`. The contents of the file are not used by oxidized - just the existence of the file with the IP address of a node will prevent the node from being polled.
+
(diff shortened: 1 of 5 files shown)


jsynack:
I forgot to mention - this also implements the changes from PR #1327  . So that one could be closed if this one is taken.

ytti:
Can you do separate PR fo the jobs_done. I'm not sure about the lockout implementation. I think I prefer another Asetus instance (and consequently file) with array of IPs, names, but perhaps I misunderstand problem and current implementation may better address is.

ytti:
If it were instance of Asetus, we could POST/GET update it via oxidized-web, and then #save the Asetus instance to get it to disk too.

jsynack:
I havent looked much as the Web-i-fication aspects - but it does make sense to make it API-able and save the config out for large lockouts.

I will play around with some asetus-style methods and see what I can come up with.

mortzu:
any news on this?

Issue #561, opened 2016-09-08
Email notification on failed devices


sakthimaan:
Please send email notification to Network admins if backup is getting failed for any of the devices. 
Also add support for send device configuration changes in email.

ytti:
I'd like to say these are important to me, but they are not. 

I imagine vast majority of organisations already have stash, github, gitlab or equivalent installation, one which can already send diffs, reusing this reduces complexity for the organisation.

I also think emails are overused, email is not ticketing system, it's extremely poor guarantee that things are actually acted upon. We can get JSON data from node's configuration fetch information, we can easily inject this JSON data to any monitoring system, to get actual actionable alerts in your  normal alerting system.

Having said, that I'm not going to close this case. I maybe completely wrong, I may be missing obvious use cases.  So I'd like other people to participate and give their opinion on the matter.

I would definitely accept hooks for either of the problem. I might accept it in core. I'm not going to develop it.

rfdrake:
I like your idea of having a git hook send the change notice.  It would allow the email to be customized according to business logic, maybe highlight certain changed lines in the diff.

Here is something I found while searching "git post commit hook email diff"

http://stackoverflow.com/questions/552360/git-hook-to-send-email-notification-on-repo-changes

I think a good change to oxidized would be to add documentation with an example of how to setup the git email hook.  People coming from rancid might expect the emails.  Even if some people can be convinced the methodology is wrong, others might still miss the emails.

ytti:
We have this hook ```https://github.com/ytti/oxidized/blob/master/extra/oxidized-report-git-commits```

lobo5519:
Partial agreement - I am using Gitlab which already has the ability to send an email (or generate a ticket) on a change with the diff. I think the more important feature is to generate and email/ticket on failing to login to a device. I'm not sure how else to detect that. 

Oxidized-web isn't proactive and I actually just disabled it due to lack of authentication and the configs are available.

ytti:
You have access to JSON data of devices, I know some companies have very advanced metrics and graphs about device config backup performance based on the JSON stats of devices.

Regardless what your monitoring system is, it should be somewhat easy to digest the JSON data and produce custom alerts.

Issue #409, opened 2016-05-02
Email generation issue


BullfroG21:
Hi,

first of all thanks for your effort building oxidized.

We are replacing our Rancid/CVS using Oxidized/Git and using also the oxidized-web as interface, and we face a small issue that maybe it’s already solved but we do not find how. 
We have set the oxidized to send an email after a diff detection. The problem is that the email its launched per device and we would like to change this behavior to get an email per all router.db defined devices after all of them have been checked to avoid many emails in our network list.
We already thought of writing the commit ids in a file and then via cron read the commit ids file and compose an email hourly, but we think it’s not the best solution.

Do you have some idea of how to deal with this issue?

Best regards,
Florian

ytti:
Do you the email hook in extras directory? 

There are few things we could do, we could launch a hook after each full rotation (but that is not really guaranteed to be correct, as people can change the order of the rotation, so it would only be correct if order was not changed during given rotation), then you could send diff between previous rotation time and current rotation time.

Another option would be do this in the per commit hook, but only send like one diff per hour, 6 hours, day or something like that, and just return without sending email if it's already been done within that timeframe.
You'd need to store the last launch timestamp in some file, and when say 6 hours from that has gone, you'd ask git for diff between the previous timestamp and current timestamp and email that.

amarti2038:
Hi ytti,

I am a BullfroG21 colleague that works also in the same project.

Thanks for your response. We actually used our custom mail script but we are trying yours (oxidized-report-git-commits) just now :)

What would be the event to get it run after a full rotation? I just see node_success, post_store, and node_fail. If there is no we will go ussing the timestamps.

Best regards,
Albert M.

ytti:
There isn't such event, I'm open to accepting PR for such event added.

callumbnz:
I for one would also like this feature - all updates grouped into one email would be amaze.

BullfroG21:
Hi again,

After a pause we are again back xD

We think that @ytti in the comment from 2 May tells an interesting approach, because we can "grant" that no one will change the last device from the router.db. If I understand fine, you mean, to create a custom model for one of our devices, place it in the bottom of the router.db and then invoke from this custom model the mailer script. It’s that fine? 

The point here is if the execution order from the router.db its followed or not, taking into account the threads. Or should we limit the threads to 1 to force a sequential behavior?

With kind regards,
Florian and @amarti2038

amarti2038:
Hi @ytti,

[...]

PR #1332, opened 2018-05-07
Add capability to pipe diffs into post_store hooks

gs-kamnas:

* Adds the optional capability to pipe the config diff to standard input of handler scripts/binaries in the exec hook.

* Documentation of changes to the exec hook.

### DIFF
--- a/docs/Hooks.md
+++ b/docs/Hooks.md
@@ -18,7 +18,7 @@ Following configuration keys need to be defined for all hooks:
 
 ## Hook type: exec
 
-The `exec` hook type allows users to run an arbitrary shell command or a binary when triggered.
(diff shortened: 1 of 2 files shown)


wk:
@gs-kamnas could you explain in a few words on your use case for this functionality? Is this something you use to e-mail diffs out in the style of rancid as an alternative to the [e-mail hook](https://github.com/ytti/oxidized/blob/master/extra/oxidized-report-git-commits)?

gs-kamnas:
While I cannot discuss the details, the use case is to both generate emails in a manner very similar to the email hook, as well as deliver the diffs to a 3rd party NMS solution.

wk:
@ytti this looks mostly harmless, and potentially offers a building block toward a solution for #409 and maybe #561. What's your take on this?

mortzu:
Please process the checklist and fix the conflict

PR #1333, opened 2018-05-07
Add counters to correctly track the number of runs, successes and failures

gs-kamnas:

* Adds additional counter variables to allow the config download success and failure count to remain accurate beyond  the 11th run.


ytti:
Another approach could have been:

```
@stats[:counters] = Hash.new(0)
```
and

```
@stats[:counters][job.status] += 1
```

PR #1336, opened 2018-05-09
Create stoneos.rb

macaty:
Add new model of Hillstone Networks StoneOS software

### DIFF
--- a/lib/oxidized/model/stoneos.rb
+++ b/lib/oxidized/model/stoneos.rb
@@ -1,0 +1,38 @@
+class Stoneos < Oxidized::Model
+#Hillstone Networks StoneOS software
+
+  prompt /^\r?[\w.()-]+[#>](\s)?$/
+  comment '# '
+  
+  expect /^\s.*--More--.*$/ do |data, re|
+   send ' '
+   data.sub re, ''
+  end
+ 
+  cmd :all do |cfg|
+    cfg = cfg.lines.to_a[1..-2].join
+  end
+
+  cmd :secret do |cfg|
+    cfg
+  end
+
+  cmd 'show configuration running' do |cfg|
+    cfg
+  end
+
+  cmd 'show version' do |cfg|
+    comment cfg
+  end
+
+  cfg :telnet do
+    username(/^login:/)
+    password(/^Password:/)
+  end
+
+  cfg :telnet, :ssh do
+    post_login 'terminal length 256'
+    post_login 'terminal width 512'
+    pre_logout 'exit'
+  end
+end
--- a/lib/oxidized/model/stoneos.rb
+++ b/lib/oxidized/model/stoneos.rb
@@ -1,25 +1,19 @@
-class Stoneos < Oxidized::Model
-#Hillstone Networks StoneOS software
+class StoneOS < Oxidized::Model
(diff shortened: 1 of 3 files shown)


lib/oxidized/model/stoneos.rb
```diff
@@ -0,0 +1,38 @@
...
+  expect /^\s.*--More--.*$/ do |data, re|
+   send ' '
+   data.sub re, ''
+  end
+ 
+  cmd :all do |cfg|
+    cfg = cfg.lines.to_a[1..-2].join
+  end
+
+  cmd :secret do |cfg|
```
wk: If the `:secret` block does not modify the input in any way, it may be unnecessary.

lib/oxidized/model/stoneos.rb
```diff
@@ -0,0 +1,38 @@
...
+    cfg
+  end
+
+  cmd 'show configuration running' do |cfg|
```
wk: Since the input is not modified in any way, this block can instead be expressed as a one line `cmd 'show configuration running'`

lib/oxidized/model/stoneos.rb
```diff
@@ -0,0 +1,38 @@
...
+  prompt /^\r?[\w.()-]+[#>](\s)?$/
+  comment '# '
+  
+  expect /^\s.*--More--.*$/ do |data, re|
+   send ' '
+   data.sub re, ''
+  end
+ 
+  cmd :all do |cfg|
+    cfg = cfg.lines.to_a[1..-2].join
```
wk: There should be no need for assignment here, simply doing `cfg.lines.to_a[1..-2].join` should suffice.

lib/oxidized/model/stoneos.rb
```diff
@@ -0,0 +1,38 @@
+class Stoneos < Oxidized::Model
+#Hillstone Networks StoneOS software
+
+  prompt /^\r?[\w.()-]+[#>](\s)?$/
+  comment '# '
+  
+  expect /^\s.*--More--.*$/ do |data, re|
+   send ' '
```
wk: Indentation in this block is inconsistent, 1 character deep instead of 2 (vs rest of functions).

lib/oxidized/model/stoneos.rb
```diff
@@ -0,0 +1,38 @@
+class Stoneos < Oxidized::Model
+#Hillstone Networks StoneOS software
```
wk: This project enforces a style that requires the comments to be at the same nesting level as the rest of the block (thus, either the comment goes before the class definition, or should be 2 spaces deep), and a space between the comment character and the contents of the comment, thus `# Hillstone Networks StoneOS software`

lib/oxidized/model/stoneos.rb
wk: `each_line` rather than `lines` as the latter has been deprecated.

wk:
The suggested revisions have been made available as PR macaty/oxidized#1. If you choose to accept them as-is, please ensure they are first tested against the device.

wk:
Pulling this with cosmetic fixes, @macaty a confirmation that the changes are not functionality affecting would still be welcome!

Thank you for the new model!

PR #1338, opened 2018-05-09
reduce logspam - auth methods are debug, not info

wk:


wk:
Displaying just the auth methods at log level `info` does not appear to be beneficial, as the name of the device itself is not logged at this level, thus the output contains auth methods for unknown devices.

Moving auth methods logging to `debug`, where it is obvious what they pertain to.

Issue #1060, opened 2017-10-09
Telnet debug logs are appended to


ja-frog:
According to https://github.com/ytti/oxidized/blob/master/docs/Configuration.md#debugging, debugging should overwrite the existing logs. This works as expected with SSH. With telnet, however, it keeps appending to the file.

wk:
@ja-frog could you give a test run to [this branch](https://github.com/wk/oxidized/tree/refactor-telnet-log) and see if it solves the issue?

ja-frog:
@wk I should be able to do so tomorrow.

ja-frog:
@wk I am sorry for the delay. Tried it out and it started a new debug file with each iteration as expected.

wk:
@ja-frog thanks! I've opened #1328 to get this change in.

PR #1343, opened 2018-05-17
Fix telnet

Skinza:
input method telnet was not working. fixed it by these changes.


wk:
@Skinza thanks! This seems to have been introduced in #1328 and was reported as successfully tested in #1060,  but seeing your fix there must have been a miscommunication somewhere. Sorry for breaking this!

PR #1347, opened 2018-05-24
Allow build without git repo / .git directory

jaylik:
This allows to build oxidized with local sources (which has no .git directory).

### DIFF
--- a/oxidized.gemspec
+++ b/oxidized.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |s|
   s.summary           = 'feeble attempt at rancid'
   s.description       = 'software to fetch configuration from network devices and store them'
   s.rubyforge_project = s.name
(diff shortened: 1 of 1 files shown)


wk:
One issue that might come up with this approach is that unlike `git ls-files`, `find` will include files which aren't part of the repository (for example, the auto-generated `coverage` directory resulting from `rake test`), and not honor `.gitignore` - requiring the user to manually prune the contents of the directory before a gem build or edit the exclusion list in the gemspec. 

I see that in your commit you've excluded the previously built gems, but in practice, this might also pick up various editor state data, test products, build products, etc.

ytti:
I'd also like to understand workflow which yields to having oxidized without having git repo, to evaluate how typical it is and how complicated it would be to avoid that workflow.

PR #1350, opened 2018-05-25
Add Extreme Networks XOS model for legacy devices

lubiewiatr:


laf:
Can you also update https://github.com/ytti/oxidized/blob/master/CHANGELOG.md and https://github.com/ytti/oxidized/blob/master/docs/Supported-OS-Types.md @lubiewiatr

PR #1352, opened 2018-05-27
Removed unnecessary/illegal/unsupported escape sequence

### LINKED ISSUE #1321: Arris C4 CMTS - Temp gsub - Illegal/unsupported escape sequence
https://github.com/ytti/oxidized/blob/d3c5d49c2045bce7891f45259b58495fd690f3fd/lib/oxidized/model/c4cmts.rb#L20

`    cfg.gsub! /\s+[\-\d]+\s+C\s+[\(\s\d]+\s+\F\)/, ''	# remove temperature readings
`

RubyMine flags the \F as illegal/unsupported escape sequence.

Best guess is they wanted to specify the string will end with F (Fahrenheit) but this doesn't need to escaped.

I'd submit a pull request to correct this, but I don't have the hardware to validate the changes still function.

cppmonkey:
Resolved #1321 
Only tested on Cisco IOS devices thus far


cppmonkey:
Now tested against APC_Aos

wk:
These look sensible, and in some instances simplify the expressions. Without having a copy of the expected input, validating equivalence further would be difficult. @laf @ytti if this goes in, it could benefit from another eyeballing.

laf:
Looks good to me from a check through the changes @wk. I've been able to test against IOS and that's all though.

davama:
if it helps, 
tested on ciscosmb 
:+1:

PR #1356, opened 2018-05-30
Create openbsd.rb

amarti2038:
OpenBSD model implementation. We have it in production and works without troubles.


yeled:
Can you clean up the lint errors @amarti2038 and I will commit this

laf:
Can you also update https://github.com/ytti/oxidized/blob/master/docs/Supported-OS-Types.md and https://github.com/ytti/oxidized/blob/master/CHANGELOG.md.

amarti2038:
the checker fails just in one line (after some commits haha).. 
"lib/oxidized/model/openbsd.rb:63:5: W: Lint/UselessAssignment: Useless assignment to variable - cfg. Use + instead of +=.
    cfg += cmd('cat /etc/passwd')
    ^^^"
but if I use the + its just the same.. do u know why? :P

wk:
Thanks for the model!

The reason for the warning you're seeing is that this is a slightly quirky way (both here and in `cumulus.rb`) to use a combination of `pre do` (where the block appears to not return `cfg` at the end) and `cmd :all` which results in rubocop concluding that assigning values to `cfg` is pointless as it is never actually returned at the end of the block.

I've regenerated the rubocop exclusion list to allow this for now, but it'd be great to clean up both these models.

PR #1359, opened 2018-06-01
Remove more JunOS secrets

Zmegolaz:
Some JunOS secrets weren't removed, this should fix most of them.

We need to check for encrypted/hashed secrets with for example `$1$asdfasdf`, otherwise we'd remove public SSH keys and similar, which aren't really secret.


wk:
The correct and canonical manner of scrubbing secrets in JunOS is by creating an account on the device for the user as which Oxidized logs in with the `access` and `view` permissions, but not the `secret` permission, as described in the [JunOS documentation here](https://www.juniper.net/documentation/en_US/junos/topics/concept/access-privileges-levels-overview.html).

Such a user will see the configuration returned with sensitive values already masked by replacement with the string `/* SECRET-DATA */`.

Scrubbing based on the `## SECRET-DATA` hint makes sense to me, with two caveats:
* It appears to overlap with the existing defined scrubbers, as JunOS considers those a secret, so they should most likely be removed:

```
    cfg.gsub!(/encrypted-password (\S+).*/, 'encrypted-password <secret removed>')
    cfg.gsub!(/pre-shared-key ascii-text (\S+).*/, 'pre-shared-key ascii-text <secret removed>')
!!! cfg.gsub!(/pre-shared-key hexadecimal (\S+).*/, 'pre-shared-key hexadecimal <secret removed>') # Validate this
    cfg.gsub!(/authentication-key (\S+).*/, 'authentication-key <secret removed>')
!!! cfg.gsub!(/community (\S+) {/, 'community <hidden> {') # Validate this
```

* It seems like a fairly straight forward solution, which makes me wonder if there is an equally straight forward reason this is not how it has been implemented all along (is `## SECRET-DATA` inconsistent among JunOS versions, for example?). I see that the initial commit belongs to @nopedial [here](https://github.com/ytti/oxidized/commit/c19cc681e14977bf741359346fe18b8da5c51355#diff-02fe5fa894cd0474cf37454bc9a63ab0). Perhaps they (or @ytti) can shed some light into why this isn't how it's been done.

I am in favor of this change, but with the overlapping per-command scrubbing removed.

wk:
```
!!! cfg.gsub!(/community (\S+) {/, 'community <hidden> {') # Validate this
```
This is not considered SECRET-DATA by JunOS, and should remain in place.

Zmegolaz:
Yes, I know the "correct" way to hide secrets is to not give the user permission to view them, I just happened to see that only some were hidden by oxidized in a lab setup and thought I could fix it since the feature's already here.

I was also wondering why the `## SECRET-DATA` wasn't used. I looked through the history of the file to see if it was removed with some comment, but was never there.

Agree, the community replacement should be left intact. I'll verify the other replacements to see if they're needed.

wk:
If the `pre-shared-key hexadecimal` statement is validated, and the overlaps are removed, I think this can go ahead. Please also update `CHANGELOG.md` as part of this PR to include a line reflecting this change, as this is a rather widely used model.

Zmegolaz:
Turns out everything I could find, except community, has the `## SECRET-DATA` hint, is quoted, and doesn't include spaces. Even hexadecimal:

```
zmegolaz@mufasa# set security ike policy test pre-shared-key hexadecimal abcd 

[edit]
zmegolaz@mufasa# show security ike policy secrettest | grep pre-shared-key       
pre-shared-key hexadecimal "$9$pfZW0ORcSeW879A"; ## SECRET-DATA
```

wk:
Thanks for validating. This seems good now.

Issue #1348, opened 2018-05-25
Extreme Networks Summit400-48t 7.8e.1.1 -  xos model triggers false positive confich changes and changes config itself


sq9mev:
We have two problems with these legacy devices:

1. Regexp in https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/xos.rb#L30 is too restricteve, our device reports as "Summit400-48t", so regexp is missing dash.
2. Statement "disable clipaging" in https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/xos.rb#L44 modifies config itself before next statements, so we always end up with "*" in the prompt (which in itself  is quite misleading), and sets "Config Modified" field to new value, and it's another cause of false positive.

My teammate is preparing PR resolving this issue.

sq9mev:
Our Summit400-48t supports "disable clipaging session" statement, which does not modify config at all, but unfortunately this is not supported by newer devices… Our idea is just to implement xoslegacy model, but not sure if regexp (1) should be changed in xos.rb also.

wk:
Are there disadvantages to issuing a 2nd `post_login` statement in the base `xos.rb` model with `disable clipaging session`, which will result in both commands being executed at each run (one of the two generating a silent error), and thus keeping this as one single model?

Expanding the `Configuration generated` match in the base `xos.rb` model also seems to make sense.

sq9mev:
> Are there disadvantages to issuing a 2nd post_login statement in the base xos.rb model […]

Having this in one model would be nice, but…

Well, on newer devices this might make sense, but on legacy emitting `disable clipaging` marks the config as changed so this is not an option.

Another approach would be try to execute `disable paging session` and when it fails execute `disable paging` - the problem is we came from Python world and do not know Ruby at all ;(.

One another solution - emit `show version`, try to parse it, and issue `disable clipaging` or `disable clipaging session` basing on version, we did not find out which Extremeware version stopped to support `disable clipaging session`.

What do you think?

wk:
Based on ExtremeWare XOS documentation, the transition from `disable clipaging session` to `disable clipaging` for disabling paging in the current session took place in version 10.1.

We have a few ways to go here to keep this all within one model:

* introduce a `xos_legacy` knob (true or false) which will choose the `disable clipaging session` formulation over the `disable clipaging` one, letting the user configure which devices are legacy and requiring this formulation (see `comware.rb` for inspiration)
[...]

PR #1362, opened 2018-06-04
1348 falsepositives xoslegacy

sq9mev:
## Pre-Request Checklist

## Description

Add support for legacy Extreme devices, eg. Summit400-48t which marks their config as changed when `disable clipaging` is issued.

Closses issue #1348


wk:
This looks reasonable. Is `ExtremeWare` separate from `XOS` in Extreme Networks nomenclature to a point that it makes sense including both in the supported OS list? Can we get some testing for this on both old and new models?

@sq9mev How many such devices have you run this against, and what level of variability in XOS versions do they have? Should additional testing be done before this is pulled?

sq9mev:
> Is ExtremeWare separate from XOS in Extreme Networks nomenclature to a point that it makes sense including both in the supported OS list?
@wk I was wondering myself, see https://en.wikipedia.org/wiki/ExtremeXOS. Basing on this link, and discussion with @lubiewiatr i'm inclinded id deserves.

> Can we get some testing for this on both old and new models?
> How many such devices have you run this against, and what level of variability in XOS versions do they have? Should additional testing be done before this is pulled?

We have tens of Extremeware devices, most of them running Extremeware 7.8e.4, and few newer devices running xos 12.x - 15.x.

@lubiewiatr, could you provide us some feedback about how oxidized behaves with these devices?

lubiewiatr:
@sq9mev I've updated our router.rb with about 70 extreme devices and so far everything looks correctly.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Master
 
+* BUGFIX: xos model deos not modify config on legacy Extreme Networks devices (sq9mev)
```
wk: This looks like a typo (does?), also the sentence may be clearer if worded as 'no longer unintentionally modifies' or similar.

PR #1374, opened 2018-06-09
IOS-XR: formal running-config instead of normal running-config

rbeumer:
I have changed the IOS-XR model to do "show running-config formal" instead of "show running-config".

## Pre-Request Checklist

## Description
I have changed the IOS-XR model to do "show running-config formal" instead of "show running-config".

### DIFF
--- a/lib/oxidized/model/iosxr.rb
+++ b/lib/oxidized/model/iosxr.rb
@@ -22,7 +22,7 @@ class IOSXR < Oxidized::Model
     comment cfg
   end
 
-  cmd 'show running-config' do |cfg|
+  cmd 'show running-config formal' do |cfg|
     cfg = cfg.each_line.to_a[3..-1].join
     cfg
   end


rbeumer:
I don't really understand why there is a complaint about lib/oxidized/output/git.rb because I did not change that file. Can somebody who has more experience enlighten me if I need to fix anything?

wk:
@rbeumer the codecov check fail is a false positive and can be ignored.

wk:
In my understanding, `formal` in this context will order the configuration output differently - creating a large and unexpected change set upon first run of an upgraded version with this change for existing users. 

What benefit does the output of `formal` for full configuration backups offers to justify this?

rbeumer:
@wk the output is much more searchable if you have a large configuration. Let's take an example, `show run interface x`:

```
interface Bundle-Ether1
 description test
 mtu 9000
 load-interval 30
```

Here you see the interface statement and then the contextual configuration. If you have a lot of (sub)interfaces it is hard to find for example the MTU for each interface, unless you have a tool for parsing your configuration.

Now, if we use the formal configuration `show run formal interface x`:

```
interface Bundle-Ether1
interface Bundle-Ether1 test
interface Bundle-Ether1 mtu 9000
interface Bundle-Ether1 load-interval 30
```

If I wanted to search for all MTU sizes I could just grep the configuration file and also see the corresponding interface for which it has been applied.

sebasdoes:
I just saw this PR, and the lack of formal config has been the primary reason I was using my own scripts to backup Oxidized. 

The searchability of formal config (or display set in Juniper-land) is a major advantage, as well as that it makes it easier to restore small parts of the config, makes it easier to script with the config or parse it, etc.

I think the only reason against would be that it changes something. Maybe put it in a new model called iosxr-formal if you don't want to bother the current users. Formal config is a major boon to people who actually use it for all scenarios in my opinion.

ytti:
https://github.com/ytti/oxidized/blob/master/docs/Creating-Models.md

You could do something like this in ```~/.config/oxidized/model/iosxr.rb```

```
require 'oxidized/model/iosxr.rb'
class IOSXR
  cmd "show run formal" do |cfg|
    cfg.type = "iosxr-formal"
    cfg.name= "formal"
    cfg
end
```

To have separate file stored along side the normal config with the formal config.

sebasdoes:
That seems like a very reasonable solution for me personally, but I'm still a big supporter of making the abovementioned solution a standard thing in Oxidized. Having two files will hurt no one, and will make a lot of lives easier.

ytti:
I'm with you, I prefer set and formal, but I think you're making lot of assumption here about other people. People use rancid and rancid behaviour is the right behaviour, there is plethora of existing tooling in place assuming that rough configuration format. 

If we're going to talk how things should be, you shouldn't need configuration backup, your master data shouldn't be in the network. For every configuration change you should generate full new config and replace the entire device config, essentially network device config should be immutable and never be changed, only replaced.
And if so, then all the data is already in formal format somewhere in your backend.

If you do need configuration backups ,you're already doing something wrong, and then it's somewhat weak argument to say which is less wrong, I agree with you set/formal is less wrong, but I   don't expect to represent consensus opinion. 

Having two files will not hurt anyone is also massive assumption, lot of people may rely on ```ls|wc -l``` for device count and ```grep foo *``` to consider every device exactly once.
The multifile support is ```output``` specific, for git we support separate repo and separate file, which is right choice? Is it right as default for all?

rbeumer:
I tried the suggestion from @ytti and it worked flawlessly. It is very hard to find a solution that will fit everyone, but in the end you will have a different answer for everyone. Unless we want to discuss further on this topic we can close this PR if you ask me.

Issue #1385, opened 2018-06-13
Recent change to telnet.rb causes logins not to be detected


hunnymonster:
Hi all

The [recent change](https://github.com/ytti/oxidized/commit/194b6893bde341c4fc3ba4c8041d9a6b6775a97d#diff-c42fecf428133cca9b1884c510475dd0) to  /lib/oxidized/input/telnet.rb has caused all of my telnet only devices (currently a sizable but reducing list) to fail to login...

Each one shows that the telnet transport is working (log file shows the banner etc) but times out... reverting to the previous telnet.rb resolves the issue.

None of them is presenting anything unusual as the login prompt - all are "Username:"

ytti:
Can you try this
https://github.com/ytti/oxidized/tree/fix-1385_telnet

ytti:
Actually no need to test that, it won't work.

The new code also relies on #expect accepting array of expects, that needs to be addressed too.

hunnymonster:
I shall await further instructions :)

ytti:
ok now the telnet.rb has been changed, 100% untested :(

ytti:
I had bunch of NBSP characters, so if anyone tried, it was certainly broken. Might be still broken, but not sure anymore how.

hunnymonster:
Certainly appears to work on a sample size of 1...

ytti:
I added one more commit to the branch, it should be no-op, so if it already did work, it still should work. But yeah SHOULD is hardly a guarantee.

hunnymonster:
Confirming that it still works with the latest update.

ytti:
Hmm. I'm not actually sure why it would work :). I don't think the code is sending new line after user and password. Are you sure you're testing the right code?

hunnymonster:
Pretty sure I am - dug in to the commits and the one I took was 8 minutes old when I took it (just before my last reply)

ytti:
i don't get why it works, strange. Let me add the new line and see if it breaks it :>

ytti:
there is now one more commit in the branch, if you don't mind testing that as well.

Any chance you're actually running different oxidized via rvm or something?

ytti:
@Misairu does this work for you?

hunnymonster:
My oxidized is running on bare metal Ubuntu, Ruby 2.5.1p37... I'll not be able to test until the morning now though.

Misairu:
Just checked master branch. Prompt detection do not working.

[...]

PR #1413, opened 2018-06-20
added telnet password prompt detection for Netgears

cppmonkey:
Tested on Netgear M5300's running firmware 10.0.0.53, 11.0.0.23 & 11.0.0.31

## Pre-Request Checklist

## Description
Fixes issue with Netgear login via telnet.
The password prompt wasn't recognised after the latest update.

Closes issue #1385 for Netgear devices

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master
 
+* BUGFIX: netgear telnet password prompt not detected
 * BUGFIX: xos model should not modify config on legacy Extreme Networks devices (sq9mev)
 * BUGFIX: model edgecos, ciscosmb
 * MISC: bump Dockerfile phusion/baseimage:0.10.0 -> 0.10.1
--- a/lib/oxidized/model/netgear.rb
+++ b/lib/oxidized/model/netgear.rb
@@ -10,6 +10,7 @@ class Netgear < Oxidized::Model
 
   cfg :telnet do
     username /^(User:|Applying Interface configuration, please wait ...)/
+    password /^Password:/i
   end
 
   cfg :telnet, :ssh do


ytti:
Another issue in this model is, that it considers everything BUT TrueClass instance to mean that enable password is needed.
I don't understand why. To me:

```
if vars(:enable)
  cmd "enable", /expect prompt/
  cmd vars(:enable)
end
```
makes far more sense.

candlerb:
On my GS724Tv4, I have to type "enable" but there is no enable password prompt.

```
Escape character is '^]'.

User:admin
Password:********
(GS724Tv4) >ena

(GS724Tv4) #
```

So as I understand it there are three cases:

- If the enable var is nil/false, do nothing
- If the enable var is true, send "enable" and expect CLI prompt in response
- Otherwise: send "enable", wait for password prompt, and send the enable var as the enable password

ytti:
Is it always like this? That there is _no way_ to configure it to autoenable? 

If so, then we can do something like

```
if vars(:enable)
  cmd "enable", /expect prompt/
  cmd vars(:enable)
else
  cmd "enable"
end
```

Now if vars(:enable) is truthy value, we enable, expect enable prompt and then send password, expecting normal prompt.
If vars(:enable) is falsy value, we enable and expect normal prompt.

ytti:
If we also need NO enable password, perhaps we can consider 'nil' needing enable, but no actual password and 'false' no enable password?

```
if vars(:enable)
  cmd "enable", /expect prompt/
  cmd vars(:enable)
elsif vars(:enable) == nil
  cmd "enable"
end
``

**Comment by candlerb** [comment]
Some devices don't require enable at all: they go straight into enable mode after initial username/password login.

I did propose a while back (#926 ) that router model code could detect from the prompt string whether enable is needed or not - that would certainly make configuration easier.  It would have to be done at the model level, but the framework could provide some hooks to make it easier since the concept of "enable mode" and "enable password" is quite widely implemented.

**Comment by candlerb** [comment]
> If we also need NO enable password, perhaps we can consider 'nil' needing enable, but no actual password and 'false' no enable password?

Sure, as long as it's clearly documented and also it's clear what happens if you set the var in a CSV router.db file. In particular, does empty column in the CSV file mean "use the inherited default" or "send empty string" or "don't use enable" or something else?

Note: there are also devices where you must send empty string as the enable password!  An example is Netgear GS110TP:

```
Escape character is '^]'.

(Broadcom FASTPATH Switching)
Applying Interface configuration, please wait ...admin
Password:***********
(Broadcom FASTPATH Switching) >ena
Password:    << JUST HIT ENTER >>

(Broadcom FASTPATH Switching) #
```

This gives the CSV scenarios as:

1. inherit default
2. don't use enable
3. send "enable" but no password
4. send "enable", wait for prompt, send empty password
5. send "enable", wait for prompt, send non-empty password

This means that there may need to be special values of "enable" to cover each case, unless we either split into two fields (enable password / use enable), or we make the models automatically detect when an enable password needs to be sent, based on the prompts seen.

IMO, the auto-detection would be the most user-friendly approach, but would require the most changes to existing model drivers.

**Comment by ytti** [comment]
Empty column is considered empty password, i.e. just newline on password prompt, It is truthy value. 'nil' and 'false' are magic, and interpolated from string into ```nil``` and ```false``` values.

So the code example I gave

a. empty column, send enable, expect expect prompt, send new line
b. false in column, do not send enable
c. nil in column, send enable, expect normal prompt
d. anything else, send enable, expect enable prompt, send value

**Comment by candlerb** [comment]
> If we also need NO enable password, perhaps we can consider 'nil' needing enable, but no actual password and 'false' no enable password?

Just to check my understanding, do you mean that 'nil' would send the "enable" command without expecting a password prompt, and 'false' would send the "enable" command, wait for password prompt, and send the empty string as the password?

My inclination is that it would be more consistent like this:

- a ruby String is the actual enable password to send.  It could be the empty string.
- a ruby True means send the command "enable", but skip the password prompt entirely
- a ruby False or Nil means do not send the "enable" command at all

Then in the CSV[^1] file:

- empty column means inherit default value
- any double-quoted value is taken as string literal, so `""` is empty string and `"true"` is the string "true" [^2]
- the unquoted values 'nil', 'true', 'false' are converted into the corresponding Ruby values
- all other unquoted values are treated as strings.  These cannot contain the column separator character.

So:

```
router1:model:user:pass:          # just do the default from global or model config
router2:model:user:pass:false     # do not send "enable"
router3:model:user:pass:true      # send "enable", device goes straight into enable mode
router4:model:user:pass:""        # send "enable", wait for prompt, send empty string
router5:model:user:pass:"true"    # send "enable", wait for prompt, send "true"
router6:model:user:pass:weeble    # send "enable", wait for prompt, send "weeble"
router7:model:user:pass:"foo:bar" # send "enable", wait for prompt, send "foo:bar"
```

But I'm happy with any solution that covers all cases, is consistent between all device types, and is documented.

-----

[^1] Where CSV means "colon separated values" rather than "comma separated values"

[^2] Need to define mechanism for how to include double-quote inside a double-quoted string.  It could either be `""` like in a CSV file, or a traditional backslash escape `\"` (in which case backslash is `\\`)

**Comment by ytti** [comment]
All scenarios will work in my code, as far as I can see. But not like you propose, but as I explained.

**Comment by candlerb** [comment]
> Empty column is considered empty password

Ah, so once you've defined a vars enable column, you are required to enter a value for every device - you can't inherit default?

> So the code example I gave
> 
> a. empty column, send enable, expect expect prompt, send new line
> b. false in column, do not send enable
> c. nil in column, send enable, expect normal prompt
> d. anything else, send enable, expect enable prompt, send value

OK then that's pretty similar; the main difference is that the netgear driver currently uses "true" to mean "send enable and expect normal prompt", whereas you're proposing changing to "nil" for this purpose.

That'll work; it just doesn't feel right to me to use two "falsey" values to do very different things, or to use a false-like value to do a positive thing (yes, I *do* want you to send "enable" command).

false = "do not send enable", true = "send enable" is something I could remember more easily.

**Comment by ytti** [comment]
Your version is possible, I don't care which it is. To me ```nil``` sounds like yeah password is needed but it is ```nil```, but this is highly subjective.

Your version would be:

```
if vars(:enable) == true
 cmd "enable"
elsif vars(:enable)
  cmd "enable", /expect prompt/
  cmd vars(:enable)
end
```

I'm ok with that too.

**Comment by candlerb** [comment]
I believe your proposed code matches what the Netgear driver already does (backwards compatible, yay!), but your code is shorter and clearer.

I gave a quick test and it looks OK.  This is with my current device which requires enable without password; I have not tested any other scenarios.

```
--- netgear.rb.orig	2018-06-25 07:24:59.913846992 +0000
+++ netgear.rb	2018-06-25 10:51:16.975868610 +0000
@@ -13,15 +13,12 @@
   end

   cfg :telnet, :ssh do
-    if vars :enable
-      post_login do
-        send "enable\n"
-        # Interpret enable: true as meaning we won't be prompted for a password
-        unless vars(:enable).is_a? TrueClass
-          expect /[pP]assword:\s?$/
-          send vars(:enable) + "\n"
-        end
-        expect /^.+[#]$/
+    post_login do
+      if vars(:enable) == true
+       cmd "enable"
+      elsif vars(:enable)
+        cmd "enable", /[pP]assword:\s?$/
+        cmd vars(:enable)
       end
     end
```

And I've not looked for any other drivers which might want to be changed similarly.

ytti:
probably all of them would benefit from it, but let's just do this one first :). Can you do PR? Fix my alignment error on the first cmd "enable". Thanks!

PR #1421, opened 2018-06-26
Ignore SROS "Time Last Saved"

peter-:
From a device running version 15.0R8

### DIFF
--- a/lib/oxidized/model/sros.rb
+++ b/lib/oxidized/model/sros.rb
@@ -27,9 +27,10 @@ class SROS < Oxidized::Model
   #
   cmd 'show system information' do |cfg|
     #
-    # Strip uptime.
+    # Strip uptime & last save metadata
     #
     cfg.sub! /^System Up Time.*\n/, ''
+    cfg.sub! /Time Last Saved.*/, ''
     cfg.gsub! /# Finished .*/, ''
     cfg.gsub! /# Generated .*/, ''
     comment cfg


wk:
Could you comment on the use case for scrubbing the last saved time? Presumably, this field is updated only when a new configuration has been saved, which should generate a legitimate diff, containing the updated time and new configuration.

Does your use case constantly re-save unchanged configurations on these devices? Is this a common, or even rational use case which should be accommodated?

peter-:
Sorry for the delay, I was only relaying this here for my NOC colleages: Without that change we've had some 40k commits from few real changes, with most of those commits consisting of _nothing_ but:

```diff
-# Time Last Saved        : 2018/06/15 02:00:01
+# Time Last Saved        : 2018/06/16 02:00:01
 # Changes Since Last Save: No
```
In all of those cases "Changes Since Last Save" was "No", indicating that no change was made.

I guess even if "Changes Since Last Save" changed from "No" to "Yes" (and certainly not when it changes from "Yes" to "No", that's another thing I'd probably need to filter) I might not care about the "Time Last Saved" date in the diff (or the fact that the value of "Changes Since Last Save" itself changed since last time).

In short: A(ny) difference in device configuration itself should be sufficient to signal a difference in device configuration, no? Tracking that metadata too seems error prone and I don't see what value tracking it in git provide. But maybe I'm missing something?

rvandepu:
A use case could be a script that runs every x-hours to save the configuration. This prevents loss of configuration when users forget to trigger an 'admin save' manually. However, the 'time last saved' timestamp is altered, even when no configuration was changed in the meantime. This causes a false positive diff, because only the 'time last saved' timestamp was changed.

If a legit change was made to the config, it would create a diff.

rvandepu:
@wk, could you take another look at this and accept the PR if deemed OK?

mortzu:

PR #1442, opened 2018-07-09
New model for HPE Moonshot Switch

empi89:
## Description
Added new model for HPE Moonshot Switch. Tested with
* Moonshot-180G Switch, 2.0.0.14
* Moonshot-45G Switch, 2.0.0.14


lib/oxidized/model/hpvc.rb
```diff
@@ -0,0 +1,22 @@
+class Hpvc < Oxidized::Model
+  # HPE Moonshot Switch / HP Virtual Connect Linux
+  # sometimes the prompt might have a leading nul or trailing ASCII Bell (^G)
+  prompt /^\0*(\([\w.-]+\)).?[>#]$/
+  comment '# '
+
+  cfg :telnet, :ssh do
+    post_login 'enable'
```
wk: Do these devices never require enable password? Should this perhaps be wrapped around in a conditional block depending on if an enable variable is set, and if it contains a password string?

lib/oxidized/model/hpvc.rb
```diff
@@ -0,0 +1,22 @@
...
+    pre_logout "exit\nquit\n"
+  end
+
+  cmd 'show version' do |cfg|
+    comment cfg
+  end
+
+  cmd 'show running-config' do |cfg|
+    cfg.gsub! /^\!System Up Time .*$/, '\\1 <removed>'
+    cfg.gsub /^\s+/, ''
```
wk: Oxidized normally prefers to store configurations as-returned. If the goal is instead to compact the file, it may be unnecessary.

Is the goal here instead to remove a prefix or suffix empty line? Consider the .cut_head, .cut_tail, or .cut_both methods. 

For clarity, it may be useful to return 'cfg' explicitly as a last statement.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 * MISC: bump Dockerfile phusion/baseimage:0.10.0 -> 0.10.1
 * FEATURE: add viptela model (@bobthebutcher)
 * FEATURE: add ECI Telecom Appolo platform bij arien.vijn@linklight.nl
+* FEATURE: add hpvc model (@empi89)
+
```
wk: These seem to be a spurious empty line that snuck in here.

mortzu:
any news on this?

PR #1464, opened 2018-07-25
Remove 'Time Last Saved' line in sros.rb model

rvandepu:
## Description
The 'Time Last Saved' line in cmd 'show system information' is updated when a user executes 'admin save', even if no configuration was changed. This causes false positive diffs.
This change strips the line from the output.


rvandepu:
Seems like a duplicate of https://github.com/ytti/oxidized/pull/1421. Closing this PR.

PR #1470, opened 2018-07-28
Support for Calix AXOS Devices

djamp42:
I have tested this on a Calix AxOS E5-16F, backups config perfectly. The first line in the config file is the command to show the config, i don't know how to remove this, so if someone tells me i can update it and test again.

## Pre-Request Checklist

## Description
Add configuration backup support for Calix AXOS operating system, Tested with Calix E5-16F

### DIFF
--- a/lib/oxidized/model/axos.rb
+++ b/lib/oxidized/model/axos.rb
@@ -1,5 +1,5 @@
 class AxOS < Oxidized::Model
-  prompt /([\w.@()-]+[#]\s?)$/
+  prompt /(\x1b\[\?7h)?([\w.@()-]+[#]\s?)$/
   comment '! '
 
   cmd 'show running-config | nomore' do |cfg|
--- a/lib/oxidized/model/axos.rb
+++ b/lib/oxidized/model/axos.rb
@@ -1,0 +1,11 @@
+class AXOS < Oxidized::Model
+  prompt /([\w.@()-]+[#]\s?)$/
+  comment '! '
+  cmd 'show running-config | nomore' do |cfg|
+    cfg
+  end
+
+  cfg :ssh do
+    pre_logout 'exit'
+  end
+end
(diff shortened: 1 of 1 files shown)


ytti:
Most models have ```cmd :all``` which is ran to all models, which typically removes the first and last lines. Check that.

Also the prompt seems be odd, why is there group of letters containing just ```#```. Not that it matters, just bit of an implication that it may not be entirely thought out.

ytti:
Maybe there is \r at the start of the prompt?

```prompt /^\r?[\w.-]+#\s?$/```

But even more important, you need to get rid of the prompt etc in the the output, as instructed.

djamp42:
Okay, i've updated to take care of the first and last prompt.. I tried that Expression, but it didn't work, in fact i've tried quite a few that match in my reg-ex tester, but don't when i run oxidized, not sure what else i can try.

ytti:
The most probable example is that there is some unprintable characters at the beginning, which you should see by reviewing the output with appropriate editor.

ytti:
Are you sure #cut_tail is enough? That you don't need #cut_both (i.e. #cut_head and #cut_tail)

djamp42:
Ahh your right, just logged all output via PUTTY and i see this, i guess i'll play around some more and see if i can get a better match.

`ESC[?7hE5-01-22-2-comm# `

djamp42:
Okay well i guess i have to leave it as is. The initial prompt when you login has the extra characters at the beginning, but everytime the prompt comes up after that it doesn't.

ytti:
That is \e[7h for setting autowrap mode, you could match for that in regexp 0 or 1 time.

djamp42:
Thanks, found a regex that matches better now, tested, should be all good to go...

ytti:
I would have made the group non-capturing ```(?:foo)``` instead of ```(foo)``` but this is fine. Thank you for your contribution.

ytti:
Can you please list the model/device in the supported devices list, after that I'll merge this. Thanks!

Issue #1148, opened 2018-01-15
Cisco SG220 series not working with ciscosmb.rb


packetkicker:
Hi,

I have been using Oxidized for the last few months to backup configurations from various pfSense, Cisco, Cisco SMB and Ubiquiti devices which has been working really well so far.  Had an install last week with 5 x Cisco SG220-26P switches.  Have set these up with SSH access but there seems to be an issue that Oxidized will not get past the login.

I can connect to the device ok from the command line on the oxidized server:

```
[oxidized@oxidized ~]$ ssh cisco@1.2.3.4 -p 8082












        Welcome to Layer 2 Managed Switch












  Press <Enter> to continue...
Username: cisco
Password: ********
DATA-SW-01#
```

I do not have to press enter for the Username: prompt to show.

I see the same in the /home/oxidized/.config/oxidized/logs/1.2.3.4-ssh log.

Any ideas what I should try to troubleshoot this further?

On Cisco SG300 series switches I normally have to change the SSH settings with `ip ssh password-auth` but this is not supported on the SG220 series.

![screen shot 2018-01-15 at 15 08 02](https://user-images.githubusercontent.com/28621135/34948860-415b39cc-fa06-11e7-9b5e-52ff58e2c0f0.png)

Where on the SG300 and above there are more options:

![screen shot 2018-01-15 at 15 12 04](https://user-images.githubusercontent.com/28621135/34948943-82d1f8b4-fa06-11e7-9c77-299fbd44000c.png)

Oxidized is v0.21.0

Thanks,
PK

mattie47:
@packetkicker 

>Have set these up with SSH access but there seems to be an issue that Oxidized will not get past the login.

Please paste the output to /home/oxidized/.config/oxidized/logs/1.2.3.4-ssh to here or on pastebin so we can see at exactly what point it's failing. 

If you find it doing something like:

```
Trying 10.37.105.10...
Connected to 10.37.105.10.
c2-access1 login: cisco
Password:
Trying 10.37.105.10...
Connected to 10.37.105.10.
c2-access1 login: cisco
Password:
Trying 10.37.105.10...
Connected to 10.37.105.10.
c2-access1 login: cisco
Password:
```

Then there's probably something wrong with the oxidized config file for the vendor model. I guess another thing to confirm is ensure your oxidized model mapping is mapping your device to the correct model. 

You can check all the device models by (probably) looking in 
[...]

PR #1473, opened 2018-07-30
Added SG220 support to login prompt.

Atroskelis:
As per #1148 , SG220 has an unique login prompt, different than the rest.

This regex should cover both cases.

## Pre-Request Checklist


wk:
Hi there! As far as I can see, besides allowing the prompt to appear anywhere in the line rather than at the beginning - the modified username prompt regexp is actually stricter and does not introduce new cases. The modified password prompt also relaxes the match to include lowercase p ```password:```. 

Could you try if the following works for you instead, and if so, update your PR?

```ruby
username /^User ?[nN]ame:/
password /^\r?[pP]assword:$/
```

If it does not, it would be interesting to see a log of an interaction with an SG220 to understand what stray characters can be present before a prompt (```username /^\r?User ?[nN]ame:/``` might be a good guess).

Atroskelis:
Hello!

I have to admit i was not so good at the regex, so i just found a valid pattern that worked for both cases. I did not bother or think it was efficient, so i expected someone to find a potential more efficient pattern, many thanks for that!

I shall test and update the PR.

Atroskelis:
@wk 

Your code does not work.

SG220 still throws a 

` oxidized[24015]: 172.16.31.97 raised Oxidized::PromptUndetect with msg "#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012        Welcome to Layer 2 Managed Switch#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012#015#012  Press <Enter> to continue...#015#012#033[H#033[JUsername:  not matching configured prompt (?-mix:^\r?([\w.@()-]+[#>]\s?)$)"`

which definitely does not match /^User ?[nN]ame:/ due to characters before 'username'.

wk:
@Atroskelis thank you for testing! 

I see it appears that the different SG models may or may not include a newline before the `Username` prompt, and some seem to emit telnet control codes instead. Matching the beginning of a line won't work here - how about a simple `username /User ?[nN]ame:/` instead in your use case? (note the removed `^`)

Atroskelis:
@wk This regex pattern seems to work !

Alas, i'm not exactly sure how do do with the password, as i assumed it was the same with the username (just in case) . Nor the original nor the attempted altered version of yours /[Pp]assword:$/ worked. 

Debug showed like 

```
Connected to 172.16.31.209.
Username: cisco
cisco
Password: Trying 172.16.31.209...
Connected to 172.16.31.209.
Username: cisco
cisco
Password: Trying 172.16.31.209...
Connected to 172.16.31.209.
Username: cisco
cisco
```

My pushed variant for Password seems to work, though.

Atroskelis:
Hello, 

For me without /(?:(?:Username:)|(?:(?:User\ Name\:))|(?:(?:User\ name\:)))/ it doesn't work

ntokos:
Hello again.
I think the more generic username prompt that @wk proposed above and the removal of the dollar sign in the password prompt that I noted above should solve the login issues in SG220 switches.
@Atroskelis can you please try the following for login prompt in your ciscosmb.rb model file:
```
username /User ?[nN]ame:/
password /^\r?Password:/
```
This works perfectly well for my SG220 and SG550X Cisco SMB switches.

Atroskelis:
Hello,

The proposed prompt syntax by @ntokos  apparently works. We can commit that one instead of the one i proposed.

Atroskelis:
@wk  Fixed according to comments and tested functional, can be approved.

Thanks,

wk:
Pulled this one in, thanks!

PR #1483, opened 2018-08-07
added support for config mode to prompt regex

zylent:
## Pre-Request Checklist

## Description

added support for config mode to regex, this enable oxidized-script to make config mode changes to powerconnect devices without breaking existing support.


wk:
Looks good!

It might be beneficial to be as specific as possible to avoid catching things like MOTDs and banners.

Would you consider something along the lines of ```/^([\w\s.@-]+(\(\S*\))?[#>]\s?)$/``` here instead? Following the examples [here](https://www.dell.com/community/Networking-General/PowerConnect-Common-Example-Commands/td-p/3873249) it seems the config modes are represented as ```hostname(config-mode-type)``` on these devices.

zylent:
That looks excellent. I'll patch my local instance to test it out.

wk:
@zylent can you confirm if the proposed change works for you, or if you've reverted to your original change and instruct how we should proceed here?

zylent:
yep! works great. Sorry, I had added it to my instance and completely forgot. It's been fine in prod for quite some time. I've updated the PR with the new regex

PR #1501, opened 2018-08-30
capture ZebOS configuration from F5

yzguy:
## Pre-Request Checklist

## Description
When running an F5, you can configure Advanced Routing, this in generally BGP. This BGP configuration is done through `imish`, which drops you in ZebOS. This configuration is useful to capture/audit for changes as it affects traffic as much as any other configuration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * FEATURE: add ECI Telecom Appolo platform bij arien.vijn@linklight.nl
 * FEATURE: ssh keepalive now configurable per node with ssh_no_keepalive boolean
 * MISC: add gpgme and sequel gems to Dockerfile for sources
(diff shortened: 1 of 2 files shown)


wk:
Thanks! Pulling this as-is. Can you confirm that the ZebOS configuration file exists on the filesystem even when this functionality is not enabled / configured on the device, and thus users who do not make use of the feature will not be greeted with an unexpected error as a result?

yzguy:
ah yeah, that'll cause an issue. I spun up an AWS one quick and the zebos folder is not there. I'll get a PR up for a better solution.

wk:
Thanks! I'll keep this in for now so that you can commit a fix cleanly on top rather than reverting.

yzguy:
https://github.com/ytti/oxidized/pull/1528

PR #1509, opened 2018-09-07
Fix NDMS

yuri-zubov:
NDMS isn't work  after this commit 9db0c31f5656d1b689fd6261079730366f15591.


wk:
It seems this one inadvertently slipped through during a bulk conversion of models to use string convenience methods. Thank you for fixing it!

PR #1518, opened 2018-09-11
Update netscaler.rb

shepherdjay:
Removes values that change on every `show ns ns.conf` command

## Pre-Request Checklist

## Description

Removes the key lines in the netscaler model from the config save. We found in practice these values change on every pull of the show ns ns.conf command whether they change or not. As such I don't think its valuable to pollute the config history with these. Open to discussion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * FEATURE: add ECI Telecom Appolo platform bij arien.vijn@linklight.nl
 * FEATURE: ssh keepalive now configurable per node with ssh_no_keepalive boolean
 * MISC: add gpgme and sequel gems to Dockerfile for sources
+* MISC: add secret filtering to netscaler (@shepherdjay)
 
 ## 0.24.0
 
--- a/lib/oxidized/model/netscaler.rb
+++ b/lib/oxidized/model/netscaler.rb
@@ -14,6 +14,14 @@ class NetScaler < Oxidized::Model
     comment cfg
   end
 
+  cmd :secret do |cfg|
+    cfg.gsub! /(-password)\s\w+/, '\\1 <secret hidden>'
+    cfg.gsub! /(-keyValue)\s\w+/, '\\1 <secret hidden>'
+    cfg.gsub! /(-radKey)\s\w+/, '\\1 <secret hidden>'
+    cfg.gsub! /(-ldapBindDnPassword)\s\w+/, '\\1 <secret hidden>'
+    cfg
+  end
+
   cmd 'show ns ns.conf'
 
   cfg :ssh do


shepherdjay:
Internally we also have the following removed from on our netscaler config pull

```
  cmd 'show ns ns.conf' do |cfg|
    cfg = cfg.each_line
    cfg = cfg.reject { |line| line.match  /^# Last modified by/ }.join
```

This is because the Last modified by on the netscaler isn't a user name but the command `save ns config` and therefore we found it absent other changes to not be valuable. However I did not want to make it part of this pull request without some input from others that use this.

wk:
Seeing how these all seem to actually be "secrets" (for some definition thereof), perhaps this actually belongs under secret scrubbing? 

Is this behavior (of obfuscated secrets coming up different at every configuration poll) consistent between different versions of the device firmware?

shepherdjay:
I'm not certain how the secret scrubbing is different but I can certainly modify it to use that behavior if desired. Would just take a little direction.

As far as firmware goes I only have access to 11.0 and 11.1 versions for active testing. I'm not certain if this behavior changed or is different in 12.0 or pre 11.0 versions. I know 11.0  has been ended of software maintenance with Netscaler but I agree it would be good to test 12.0+ versions as well for the same behavior.

wk:
The thought behind classifying these additions as a secret would be that they would hide behind a well-known knob reserved for removing secret-like information (passwords, hashes, etc.), which would allow the original contributor (who presumably wasn't bothered by, or didn't experience, the behavior you report) to remain using it as-is, but allow you and other users to avoid this behavior by turning on secret scrubbing.

My interest in other versions stems from the idea that perhaps the original contributor never observed this behavior due to running a different version where it does not exist.

shepherdjay:
As far as the versions its hard to know, we've been using oxidized for some time and only recently realized the behavior with the netscaler due to turning on email notifications for our git repository. Once it was emailing us at every interval and the changes were just in the hash we moved to provide a quick patch internally.

I'll see if I can find any official documentation on the behavior, I also want to test if these hashes can even be copy pasted back into the netscaler. It could be that this is a built in obfuscation in which case backing them up provides no purpose and therefore I don't feel would make sense to put under the secrets module.

If I can't find any supporting documentation one way or another, or you feel it should still go under secrets regardless, I will amend this pull request to place under secrets

shepherdjay:
As an update in my testing I found that the hashes can be copy pasted back into the Netscaler. The key seems to be a `-encrypted -encryptmethod ENCMTHD_3` flag that is also part of the backed up config. However documentation wise both the interactive help reached with `?` and Citrix online documentation does not show these flags or how they are used. 

Unfortunately in my testing this undocumented `-encrypted` flag accepts any key value not just hex. Unlike Cisco it doesn't provide a "not valid encrypted secret" error. It will still throw a no surch argument error in 11.0 that I was testing with as follows:

````
> add authentication radiusAction RSA_RADIUS -serverIP 127.0.0.1 -radKey <secret hidden> -encrypted -encryptmethod ENCMTHD_3                                                                                                                                                            ^^^^^^^
ERROR: No such argument [hidden>]
````

Given this information I will edit this pull request to put it under secrets

shepherdjay:
Updated and squashed commits, utilized same format for secret as the ios model

wk:
Given that these all appear to effectively be secrets, this seems to be the most obvious manner to represent the intent without affecting behavior for existing users. Thanks for restructuring the PR as well as documenting the behavior! I've also bumped `CHANGELOG.md` to reflect this.

PR #1525, opened 2018-09-13
EOS: filter out BGP passwords and only user password hashes

yzguy:
## Pre-Request Checklist

## Description
I noticed that when removing secrets from EOS configuration, the regex was not account for there being a role assigned to the user, only a privilege, therefore not removing the hash. This fixes the regex to capture that and only remove the hash, previously it removed the whole line, which is unnecessary

In addition, the BGP password is removed for neighbors. It was not previously removed, and would be now.

Test

```
cfg = %{username example privilege 15 secret sha512 $6$fj1dmU/0CgLorUGB$ziKO4cK01GyzpVNulKa/T9qEEGSZfrzzj0Ol5qhQ2fKUptRVm.
router bgp 12345
   neighbor 10.10.10.10 password 7 aaBc123Bc123Bc123Bc123
}
cfg.gsub! /(secret \w+) (\S+).*/, '\\1 <secret hidden>'
cfg.gsub! /(password \d+) (\S+).*/, '\\1 <secret hidden>'
puts cfg
```

Will generate

```
username example privilege 15 secret sha512 <secret hidden>
router bgp 12345
   neighbor 10.10.10.10 password 7 <secret hidden>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,17 +2,18 @@
 
 ## Master
 
-* BUGFIX: netgear telnet password prompt not detected
-* BUGFIX: xos model should not modify config on legacy Extreme Networks devices (sq9mev)
-* BUGFIX: model edgecos, ciscosmb
-* MISC: bump Dockerfile phusion/baseimage:0.10.0 -> 0.10.1
-* MISC: Added verbiage to set OXIDIZED_HOME correctly under Debian 8.8 w/systemd
 * FEATURE: add viptela model (@bobthebutcher)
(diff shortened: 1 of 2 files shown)


zlesnr:
Thanks for the merge.  @wk Is there a timeline for the next release?  Currently using gem releases to deploy.

wk:
@ytti thoughts on next release timeline and maybe a semi regular (quarterly or twice-annual?) release schedule?

zlesnr:
@ytti May we request a release for this change please?

ytti:
I'll push new gem during the weekend.

PR #1528, opened 2018-09-17
tmos: ZebOS configuration should only be captured if it's configured

yzguy:
## Pre-Request Checklist

- [] Tests added or adapted (try `rake test`)
- [] Changes are reflected in the documentation

## Description
In #1501 ZebOS is captured, however ZebOS is used when the F5 has Dynamic Routing licensed + provisioned. This adds a check for the `/config/zebos` directory before it attempts to read the ZebOS configuration.

Makes #1501 more robust


wk:
Pragmatic. Let's get this in.

PR #1540, opened 2018-09-21
add model specs to edgeos.rb

systeembeheerder:
## Pre-Request Checklist


lib/oxidized/model/edgeos.rb
```diff
@@ -16,6 +16,11 @@ class Edgeos < Oxidized::Model
     cfg
   end
 
+  cmd 'show version | no-more' do |cfg|
+    cfg.gsub! /^Uptime:\s.+/, ''
```
wk: Could we use something like this instead of `gsub!` here?
```ruby
cfg = cfg.each_line.reject { |l| l.match /^Uptime:\s.+/ }.join
```
Currently, the use of `gsub!` and `reject` is highly inconsistent throughout the models, but there might be some clarity benefit to using `reject` when discarding whole lines, vs. using `gsub!` when doing actual substitutions.

PR #1545, opened 2018-09-21
Update Supported-OS-Types.md

STetzel:
Added HPE bladesystem virtual-connect modul


wk:
This doesn't seem like it needs to be a separate PR, but could (and should) be an additional commit under #1342. Closing this PR, please move these changes under #1342 instead.

PR #1550, opened 2018-09-24
added 'username: nil' in Advanced Config

Atroskelis:
## Pre-Request Checklist

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -186,6 +186,8 @@ map:
 
 For model specific credentials
 
+You can add 'username: nil' if the device only expects a Password at prompt. 
+
 ```yaml
 models:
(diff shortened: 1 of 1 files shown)


ytti:
This is not really specific for username.
https://github.com/ytti/oxidized/blob/master/lib/oxidized/source/source.rb#L14-L18

The whole point of this is, it is valid to have empty string as value, and lot of people have, for example requiring empty enable password is very typical.
Fro enable we need communicate

a) no enable needed (false or nil)
b) enable needed, not password asked (true)
c) enable needed, password asked, empty password needed ("")
d) enable needed, non-empty password needed ("foo")

ytti:
ehh, i guess it's fine

PR #1556, opened 2018-09-26
adding more gsubs for removing sensitive config on FTOS/DNOS

bathax:
## Pre-Request Checklist

## Description

Some additional sensitive data needed to be removed from ftos configs when using remove_secret: true.

Before:
```
radius-server host 192.168.1.1 key 7 somehashhere
snmp-server host 192.168.1.2 traps version 2c somecommunitystring udp-port 162
```

After:
```
radius-server host 192.168.1.1 key 7 <hash hidden> 
snmp-server host 192.168.1.2 traps version 2c <community removed> udp-port 162
```

### DIFF
--- a/lib/oxidized/model/ftos.rb
+++ b/lib/oxidized/model/ftos.rb
@@ -11,8 +11,8 @@ class FTOS < Oxidized::Model
     cfg.gsub! /^(snmp-server community).*/, '\\1 <configuration removed>'
     cfg.gsub! /(secret \d* {0,1})\S+(.*)/, '\\1<secret hidden>\\2'
     cfg.gsub! /(password \d+) \S+(.*)/, '\\1 <hash hidden>\\2'
-    cfg.gsub! /(snmp-server.*version \S+) \S+(.*)/, '\\1 <community removed>\\2'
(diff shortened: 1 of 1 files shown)


lib/oxidized/model/ftos.rb
```diff
@@ -11,6 +11,8 @@ class FTOS < Oxidized::Model
     cfg.gsub! /^(snmp-server community).*/, '\\1 <configuration removed>'
     cfg.gsub! /(secret \d* {0,1})\S+(.*)/, '\\1<secret hidden>\\2'
     cfg.gsub! /(password \d+) \S+(.*)/, '\\1 <hash hidden>\\2'
+    cfg.gsub! /(snmp-server.*version \S+) \S+(.*)/, '\\1 <community removed>\\2'
```
wk: Since these seem to be stand-alone commands (rather than elements of a command), could this perhaps be prefixed with a start of line match to reduce possibility of a false positive? For example, `/^(snmp-server.*version \S+) \S+(.*)/` and `/^(radius-server.*key \d )\S+/)?

bathax: Indeed, I've made these changes.

PR #1570, opened 2018-10-02
Update netscaler.rb with better secret handling

shepherdjay:
More generic regex for finding secrets. Testing shows all secrets are followed by `-encrypted`. Makes up for some additional secrets found on new production deployments from my previous commit.

### DIFF
--- a/lib/oxidized/model/netscaler.rb
+++ b/lib/oxidized/model/netscaler.rb
@@ -15,10 +15,7 @@ class NetScaler < Oxidized::Model
   end
 
   cmd :secret do |cfg|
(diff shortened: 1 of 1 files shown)


wk:
`\h` might be a better match than `\w` here, as the string preceding `-encrypted` appears to be always represented as hexdigit.

wk:
If you're willing to test this and validate that `\h` is sufficient, a further PR to narrow this to the exact constraint would be appreciated!

shepherdjay:
Can do so, should be able to validate this across our own environment. Honestly not used to ruby regex so I didn't even know `\h` was a thing.

Issue #1574, opened 2018-10-03
model/tplink.rb: remove password in clear text for older TPLink models


lwillek:
Some of the older TPLink models ([TL-SG3210 V1](https://www.tp-link.com/de/download/TL-SG3210_V1.html) in my case) are storing the passwords in clear text within the configuration.
The password(s) will not removed if the current model/tplink.rb is used and `remove_secret=true` is set.

I will write a small patch to fix that.

lwillek:
https://github.com/ytti/oxidized/pull/1575

PR #1575, opened 2018-10-03
model/tplink.rb: remove password in clear text for older TPLink models (#1574)

lwillek:
https://github.com/ytti/oxidized/issues/1574

### DIFF
--- a/lib/oxidized/model/tplink.rb
+++ b/lib/oxidized/model/tplink.rb
@@ -27,6 +27,8 @@ class TPLink < Oxidized::Model
   end
 
   cmd :secret do |cfg|
+    cfg.gsub! /enable password (\S+)/, 'enable password <secret hidden>'
+    cfg.gsub! /user (\S+) password (\S+) (.*)/, 'user \1 password <secret hidden> \3'
     cfg.gsub! /^(snmp-server community).*/, '\\1 <configuration removed>'
     cfg.gsub! /secret (\d+) (\S+).*/, '<secret hidden>'
     cfg
--- a/lib/oxidized/model/tplink.rb
+++ b/lib/oxidized/model/tplink.rb
@@ -19,7 +19,7 @@ class TPLink < Oxidized::Model
 
   cmd :all do |cfg|
     # remove unwanted paging line
-    cfg.gsub! /Press any key to contine.*/, ''
+    cfg.gsub! /^Press any key to contin.*/, ''
     # normalize linefeeds
     cfg.gsub! /(\r|\r\n|\n\r)/, "\n"
     # remove empty lines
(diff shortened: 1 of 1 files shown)


lib/oxidized/model/tplink.rb
```diff
@@ -18,13 +18,17 @@ class TPLink < Oxidized::Model
...
+    # remove unwanted paging line
+    cfg.gsub! /Press any key to contine.*/, ''
     # normalize linefeeds
     cfg.gsub! /(\r|\r\n|\n\r)/, "\n"
     # remove empty lines
     cfg.each_line.reject { |line| line.match /^[\r\n\s\u0000#]+$/ }.join
   end
 
   cmd :secret do |cfg|
+    cfg.gsub! /enable password (\S+)/, 'enable password <secret hidden>'
```
wk: Do these statements always appear the the beginning of the line, or are they part of other commands?
Could we prefix them with ^ to ensure we don't catch these words appearing elsewhere?

lib/oxidized/model/tplink.rb
```diff
@@ -18,13 +18,17 @@ class TPLink < Oxidized::Model
   end
 
   cmd :all do |cfg|
+    # remove unwanted paging line
+    cfg.gsub! /Press any key to contine.*/, ''
```
wk: Normally, we'd expect the pager handler at L8 to discard this after handling the pager. If the pager handler didn't match, then no key would've normally been pressed by the pager handler and the model should hang - whereas if it was, the prompt should've been removed. 

What situation leads to this being needed?

lib/oxidized/model/tplink.rb
lwillek: I cant say this for all models, that's why I initially did not wrote a better regex, i.e using '^'. At least for the tree different models I own, the term 'enable password (\S+)' is not part of other commands.
However, the regex `/^enable password (\S+)/` should work too and is more specific, so I will try it.

lib/oxidized/model/tplink.rb
lwillek: I can't answer exactly. I experienced this behave for all the different tplink switches I own, and this line (sometimes followed by one or more special symbols) was part of the output, and the output did not hung at his point. (i.e. the config was always complete)
As you, I think the pager handler did something wrong, but I don't have the knowledge to fix this.
That was the reason for me just to remove the line in the output. I think this should be safe.

lwillek: However, I just realized that the current regex does not match all situations. So I will review and adapt.

lib/oxidized/model/tplink.rb
lwillek: After review: yes, its possible to use this regex. Other firmware versions use a different approach (`enable secret ....` in Firmware from 2014, encrypted output) (`enale secret` in firmware starting from 2015, encrypted output), so its fine to use the better regex.

PR #1579, opened 2018-10-05
update dnos.rb to take account of EMC branding

kslawrence:
newer DNOS switches are branded EMC as well - this means the regex for excluding the uptime from config backups is no longer effective on these switches. A sample output on the newer switches is "Dell EMC Networking OS uptime is 6 day(s), 20 hour(s), 33 minute(s)"

I have added a seperate line for the new format.
I imagine its also possible to do this more gracefully by editing the regex but im not all that skilled in it.

## Pre-Request Checklist


wk:
I've updated the regexp to include both variants in a single statement, as that seems to have been the original intention of your change. Thanks!

PR #1585, opened 2018-10-09
3com 4200 support

dra:
## Pre-Request Checklist

## Description
Comware needs some modifications to support 3Com 4200 series.


wk:
Could you include the output of the login banner and `show version` or equivalent on this device that gives a hint about the software version at play?

It would be interesting to understand if this device and OS type have anything in common with the rest of the Comware product line, or if it runs a different OS which might be better served by another model (e.g. vrp.rb, which seems very similar).

raunz:
I would suggest not to add the following commands to model post_login section, as they are needed only once to get 3COM 4200 switch working with Oxidized. Enter the following commands once via CLI and there's no need to update the model.
Still updating the Model-Notes/Comware.md is crucial. It took me a while to get rid of paging at "Switch 4200G 24-Port Software Version 3Com OS V3.02.04s56" 
```
system-view
info-center source default channel 1 log state off debug state off
user-interface vty 0 4
screen-length 0
quit
save
```

mortzu:

PR #1612, opened 2018-10-30
Add logic to check :enable to see if it is password String or a Boolean.

joshland:
## Pre-Request Checklist

- [X ] Passes rubocop code analysis (try `rubocop --auto-correct`)

## Description
Powerconnect.rb didn't work in my case because an enable password is not required.  

I updated the model to allow enable to be a true or a password.

### DIFF
--- a/lib/oxidized/model/powerconnect.rb
+++ b/lib/oxidized/model/powerconnect.rb
@@ -43,7 +43,11 @@ class PowerConnect < Oxidized::Model
     if vars :enable
       post_login do
         send "enable\n"
-        cmd vars(:enable)
+        if :enable.instance_of?(String) then
+          cmd vars(:enable)
+        else
+          Oxidized.logger.debug "lib/oxidized/model/powerconnect: No enable password."
+        end
       end
     end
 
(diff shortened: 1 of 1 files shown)


ytti:
Can enable password be newline? Can it be auto-enabled after writing enable?

ios.rb model handles all these cases:

```
    post_login do
      if vars(:enable) == true
        cmd "enable"
      elsif vars(:enable)
        cmd "enable", /^[pP]assword:/
        cmd vars(:enable)
      end
    end
```

a) if enable is truthty value, enable, but do not expect enable password prompt, expect normal prompt
b) if enable is specified (even as an empty string) enable, expect enable password prompt, send enable password (might be just newline)
c) otherwise do not enable


In your case it is pilot error, you are specifying empty-string enable, which is truthy value and implies  just new line on password prompt. If you'd instead say ```nil``` or ```false``` you would not try to enable.

What you are doing now, breaks the entire model for every user who needs enable, you as are asking 'is symbol :enable a string' which is never true. It's entirely static comparison and would be better written as 'if false'.

joshland:
I am not deeply rubified.  You solution seems decent.  I need to issue the "enable", but there is no enable password.  I don't want to send bad text to the command line.

I didn't have a good use case for testing with other circumstances.  Thank you for the feedback.  I will rewrite this.

PR #1636, opened 2018-11-19
Improved secret filtering for ios

hexdump0x0200:
## Pre-Request Checklist

## Description

Improved secret filtering for ios

### DIFF
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -26,15 +26,21 @@ class IOS < Oxidized::Model
   cmd :secret do |cfg|
     cfg.gsub! /^(snmp-server community).*/, '\\1 <configuration removed>'
     cfg.gsub! /^(snmp-server host \S+( vrf \S+)?( version (1|2c|3))?)\s+\S+((\s+\S*)*)\s*/, '\\1 <secret hidden> \\5'
-    cfg.gsub! /^(username \S+ privilege \d+) (\S+).*/, '\\1 <secret hidden>'
-    cfg.gsub! /^(username \S+ password \d) (\S+)/, '\\1 <secret hidden>'
-    cfg.gsub! /^(username \S+ secret \d) (\S+)/, '\\1 <secret hidden>'
-    cfg.gsub! /^(enable (password|secret) \d) (\S+)/, '\\1 <secret hidden>'
+    cfg.gsub! /^(username .+ (password|secret) \d) .+/, '\\1 <secret hidden>'
+    cfg.gsub! /^(enable (password|secret)( level \d+)? \d) .+/, '\\1 <secret hidden>'
     cfg.gsub! /^(\s+(?:password|secret)) (?:\d )?\S+/, '\\1 <secret hidden>'
     cfg.gsub! /^(.*wpa-psk ascii \d) (\S+)/, '\\1 <secret hidden>'
     cfg.gsub! /^(.*key 7) (\d.+)/, '\\1 <secret hidden>'
-    cfg.gsub! /^(tacacs-server key \d) (\S+)/, '\\1 <secret hidden>'
+    cfg.gsub! /^(tacacs-server (.+ )?key) .+/, '\\1 <secret hidden>'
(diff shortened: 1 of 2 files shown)


lib/oxidized/model/ios.rb
```diff
@@ -26,15 +26,21 @@ class IOS < Oxidized::Model
   cmd :secret do |cfg|
     cfg.gsub! /^(snmp-server community).*/, '\\1 <configuration removed>'
     cfg.gsub! /^(snmp-server host \S+( vrf \S+)?( version (1|2c|3))?)\s+\S+((\s+\S*)*)\s*/, '\\1 <secret hidden> \\5'
-    cfg.gsub! /^(username \S+ privilege \d+) (\S+).*/, '\\1 <secret hidden>'
-    cfg.gsub! /^(username \S+ password \d) (\S+)/, '\\1 <secret hidden>'
-    cfg.gsub! /^(username \S+ secret \d) (\S+)/, '\\1 <secret hidden>'
-    cfg.gsub! /^(enable (password|secret) \d) (\S+)/, '\\1 <secret hidden>'
+    cfg.gsub! /^(username .+ (password|secret) \d) .+/, '\\1 <secret hidden>'
```
hexdump0x0200: The "username" command may include additional options. Previously, with their presence secrets were not cut.
before:
```
username TEST2 access-class 2 privilege 0 noescape secret 5 $1$qDrb$WZiqdvbykLL.x0fj59fT10
```

after:
```
username TEST2 access-class 2 privilege 0 noescape secret 5 <secret hidden>
```

lib/oxidized/model/ios.rb
```diff
@@ -26,15 +26,21 @@ class IOS < Oxidized::Model
...
+    cfg.gsub! /^(enable (password|secret)( level \d+)? \d) .+/, '\\1 <secret hidden>'
     cfg.gsub! /^(\s+(?:password|secret)) (?:\d )?\S+/, '\\1 <secret hidden>'
     cfg.gsub! /^(.*wpa-psk ascii \d) (\S+)/, '\\1 <secret hidden>'
     cfg.gsub! /^(.*key 7) (\d.+)/, '\\1 <secret hidden>'
-    cfg.gsub! /^(tacacs-server key \d) (\S+)/, '\\1 <secret hidden>'
+    cfg.gsub! /^(tacacs-server (.+ )?key) .+/, '\\1 <secret hidden>'
```
hexdump0x0200: The "tacacs-server" command may include additional options and a key with spaces. Previously, with their presence secrets were not cut.
before:
```
tacacs-server host 192.168.100.1 nat single-connection port 5 key qqq www
```
after:
```
tacacs-server host 192.168.100.1 nat single-connection port 5 key <secret hidden>
```

lib/oxidized/model/ios.rb
wk: Why was the match for `enable` also removed here? That case does not appear to be handled by the suggested replacement regexp.

hexdump0x0200: The match for ``enable`` is not deleted, but changed. Disable the display of comments on the code, then you will see the change completely, it includes:
```
- cfg.gsub! /^(enable (password|secret) \d) (\S+)/, '\\1 <secret hidden>'
+ cfg.gsub! /^(enable (password|secret)( level \d+)? \d) .+/, '\\1 <secret hidden>'
```
before:
```
enable secret level 3 5 $1$iCV/$xgCDHtF4OhgjnFhd/ff1G.
enable secret 5 <secret hidden>
```
after:
```
enable secret level 3 5 <secret hidden>
enable secret 5 <secret hidden>
```

Commenting on lines in patch code was a bad idea. It was necessary to describe all the changes in the pull-request body.

wk: The mistake is mine, glanced over this too quickly instead of looking at all the change set. The comments have in fact been helpful!

wk:
This looks good to go in. I've added a note to the `CHANGELOG.md` file as part of this change set, as the `ios` model is one of the most widely used, and users might want to be informed about any expected behavior changes (such as new secrets being filtered out).

PR #1654, opened 2018-12-12
added model AudioCodes Mediant

ecm75:
## Description
Added new model for AudioCodes Mendiant SBC. (model/mediant.rb)


wk:
Could you comment how this model relates to the `audiocodes.rb` and `audiocodesmp.rb` models? It seems quite similar, and potentially an extension of `audiocodes.rb`, judging from the commands executed; could the two be unified? Do they both target the same device(s)?

wk:
@ecm75 are you still working on this? Could you provide some input on the questions above?

mortzu:
why you've deleted other file?

PR #1479, opened 2018-08-01
Create huawei_smartax.rb

ldevantier:
Adds support to obtain the configuration from a Huawei Smart AX device - will also handle/minimise pagination prompts from the device.

You will need to have the 'enable' property for the host set to a value that evaluates to true in the router DB, but no password is actually needed to escalate into the administrative mode to access the configuration on the device.

Can also assist with some of the questions raised in https://github.com/ytti/oxidized/issues/1295 as it will disable the 'interactive' command behaviour on the device CLI during the session.

Issue #1295, opened 2018-04-20
Huawei SmartAX Custom Model


jfosnight:
We have a few devices that are running Huawei's SmartAX Operating System.  I have tried to create a custom model for these, but I've ran into some issues.

The main issue is that when you enter a command in the CLI, if the command has any additional options, it prompts the user for further input.

For example, running: `display version`
the device outputs: `{ <cr>|backplane<K>|frameid/slotid<S><Length 1-15>||<K> }:`

At this point, if you hit enter again, it will go ahead and output all the version information.  A similar thing happens with `display current-configuration`  (Also you have to run `enable` before you can run display `current-config`, which changes the prompt from '>' to '#', but I think I fixed that with some extra regex for the prompt.
 
Here was my attempt at a model after messing around with it for awhile.
```ruby
class SmartAX < Oxidized::Model
  # Huawei SmartAX

  prompt /^([\w.-]+[>#])$/
  comment '# '

  expect /^\{ <cr>.* \}\:$/ do
    send "\r\n"
  end

  cmd :secret do |cfg|
    cfg.gsub! /(pin verify (?:auto|)).*/, '\\1 <PIN hidden>'
    cfg.gsub! /(%\^%#.*%\^%#)/, '<secret hidden>'
    cfg
  end

  cmd :all do |cfg|
    cfg.each_line.to_a[1..-2].join
  end

  cfg :telnet do
    username /^login as: $/
    password /^User password:$/
  end

  cfg :telnet, :ssh do
#    post_login 'screen-length 0 temporary'
    pre_logout 'quit'
  end

   cmd 'display version' do |cfg|
     comment cfg
   end

end
```

Any guidance would be helpful.  I'm not familiar at all with Ruby, so the syntax is a bit confusing.  

Thanks

jsynack:
Hello @jfosnight ,

(Full disclosure - I do not have access to one of those devices - these are just generic thoughts)

For your `display version` command - is there an option such as `all` or anything that would proceed with the command.  I admit that I find the Huawei site a little troublesome to find simple command reference docs.

If there is not a 'one shot' way of running the command - There are a few different ways to send extra information (such as a CR).  Some information can be found on the [Ruby API page](https://github.com/ytti/oxidized/blob/master/docs/Ruby-API.md) .

It is also good to see examples of other models that require some of this behavior as well.
[...]

PR #1686, opened 2019-01-21
Huawei SmartAX support

nyash:
This PR adds support for Huawei SmartAX series devices. 

I have tested this extensively and it works flawlessly for both SSH and Telnet for configuration retrieval.

For proper SSH config retrieval it is necessary to disable SSH keepalives, otherwise net-ssh crashes randomly.

Some explaination for the commands used in the model as well as the order since it is important.

```
post_login "enable" #This enters the privileged mode, it is not necessary to provide any password. It is mandatory to be in privileged mode in order to fetch the config.
post_login "undo interactive" #This disables confirmation prompts, e.g. when entering "quit" it is otherwise necessary to input "Y" for confirmation.
post_login "undo smart" #This disables command help prompts. Normally once a valid command is entered it first displays the available parameters to which you need to input enter to proceed further or type the additional parameters.
post_login "scroll" #This disables paging. This is not obvious since it's not mentioned in the documentation and only works after entering the privileged mode by "enable".
```
 

This addresses issue #1295 and #1479

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * FEATURE: add privilege escalation to the cumulus model (@user4574)
 * FEATURE: add adtran model (@CFUJoshWeepie)
 * FEATURE: add firebrick model (@lewisvive)
+* FEATURE: add huawei smartax model (@nyash)
 * BUGFIX: netgear telnet password prompt not detected
 * BUGFIX: xos model should not modify config on legacy Extreme Networks devices (@sq9mev)
 * BUGFIX: model dlink, edgecos, ciscosmb, openbsd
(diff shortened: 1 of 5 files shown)


wk:
Thanks; I've made a few cosmetic cleanups to pass RuboCop tests and link the markdown document, but pulling this in as-is otherwise.

Closes #1295
Closes #1479

wk:
Could you perhaps add a line to the `CHANGELOG.md` file to let users know of this model and take credit for it?

nyash:
Added a new device with old firmware for cfg retrieval and noticed that the output is a bit different mandating the use of `cut_both` for the output.

PR #1688, opened 2019-01-21
New model for AudioCodes MediaPack MP1xx and Mediant 1000 devices (firmware v4.xx, v5.xx, v6.xx)

pedjaj:
## Pre-Request Checklist

### DIFF
--- a/lib/oxidized/model/audiocodesmp.rb
+++ b/lib/oxidized/model/audiocodesmp.rb
@@ -1,0 +1,37 @@
+class AudioCodesMP < Oxidized::Model
+  # AudioCodes MediaPack MP1xx and Mediant 1000 devices (firmware v4.xx, v5.xx, v6.xx)
+  # Created by pedjajks@gmail.com
+  # 2019 v3.0
+
+
+#   prompt /(\/.*?>)/
+   prompt /\/>|\/CONFiguration>/
+
+  comment ';'
+
+  cmd 'cf get' do |cfg|
+    # remove Unnecessary Lines
+    # cfg.gsub! /^AutoUPDate SaveAndReset*/, ''
+    cfg.gsub! /^cf get*/, ''
+    cfg.gsub! /^SIP\/ SECurity.*/, ''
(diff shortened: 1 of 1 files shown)


wk:
This looks good! Before proceeding with this PR, please add a line to `CHANGELOG.md` describing this pull, and add the model to the `Supported-OS-Types.md` file as part of this change set so that users can discover this model and you get credit for your contribution.

Finally, some of the scrubbing done in 'cf get' isn't immediately apparent; is this all volatile output that changes between each run? Decorative elements? Secrets which shouldn't be logged into configuration? If any of these lines are secrets (`cfg.gsub! /^SIP\/ SECurity.*/, ''` seems like a candidate), it would be more idiomatic to scrub them in a `:secret` block.

pedjaj:
Thanks, this is my first : ). I added lines to CHANGELOG.md and Supported-OS-Types.md
Unfortunately this devices has variable output and prompt that depending on executed command like in attached telnet session.
[telnet session mp104.txt](https://github.com/ytti/oxidized/files/2790521/telnet.session.mp104.txt)

It may have been better to remove all lines before 
`;**************`
`;** Ini File **`
`;**************`

and after 

`; End of INI file.`

pedjaj:
Need help or example for deleting lines in cfg before `;**************` and after `; End of INI file.`

wk:
A few things you might want to try here:
* Define the prompt as `/^\/\w*>/` or similar, to catch all mutations of `/word>` as well as `/>`, as the prompt appears to be contextual.
* try `cfg.cut_tail` to remove the last returned line (the available commands line) inside the command block, or `cfg.cut_both` to also clean away the prompt at the beginning.
* Remove the rest of the `gsub` substitutions, they normally should not get logged as only the output of `cf get` will.

Finally, I see you've acknowledged adding modifications to `CHANGELOG.md` and `Supported-OS-Types.md`, but I don't see them as part of this PR. Did you perhaps forget to commit and push these additional changes to Github? :smile_cat:

PR #1689, opened 2019-01-23
Add parsing of configurations

rbns:
## Pre-Request Checklist

## Description

this adds the possibility to add a parse method to models.
it is used when serializing a node, extracting additional information
about CPU cores, RAM, etc. which is then stored into
a new field ":info" containing a hash of these values.

NB: i'm not sure if accessing the configuration requires locking as
it is done in yield_node_output in nodes.rb.


ytti:
I think we should support this.

But some thoughts.

 * may new method, rather than serialise, as serialise is called on web and we probably don't need this to exists there always in same page
 * maybe return new object which contains the parsed information, instead of hash. This would allow us to add complexity and keep complexity in separate class
 * should the class be platform agnostic? I think this would be massive value proposal, if the methods implemented are generic, perhaps the class can contain method 'vendor' which returns another instance of same class, which contains 'vendored' methods, instead of standard sets.
 * perhaps the structure should be built during collection, so we have access to 'cmd' etc, not post fact. So maybe here: https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/model.rb#L169 we call some method, while we are still in scope of talking to the device and we pass tho the method outputs, so it has access to everything already collected. 
 * maybe model/x.rb should require+include model/x-formal.rb or something when it exists, where the method to build the formal class and vendor class exist, to avoid making models too messy?

rbns:
> I think we should support this.

Great :)

> may new method, rather than serialise, as serialise is called on web and we probably don't need this to exists there always in same page

I don't see a use-case where the data is useful in the web-ui, so this is fine for me. I only need the data to be available as JSON from the API.

> maybe return new object which contains the parsed information, instead of hash. This would allow us to add complexity and keep complexity in separate class

This is also fine, hash was just the quickest way to add this. 

> should the class be platform agnostic? I think this would be massive value proposal, if the methods implemented are generic, perhaps the class can contain method 'vendor' which returns another instance of same class, which contains 'vendored' methods, instead of standard sets.

Do you mean agnostic with regard to different switch models and vendors? If yes: This is what I've had in mind, otherwise this feature is rather useless. There needs to be a standard format for the information so that it can be parsed automatically. I'd welcome input for the field/method names, some of the ones in this PR aren't really good "hdd" for example.

> perhaps the structure should be built during collection, so we have access to 'cmd' etc, not post fact. So maybe here: https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/model.rb#L169 we call some method, while we are still in scope of talking to the device and we pass tho the method outputs, so it has access to everything already collected.

Yes, the current approach isn't the best. I will look into this :)

> maybe model/x.rb should require+include model/x-formal.rb or something when it exists, where the method to build the formal class and vendor class exist, to avoid making models too messy?

Sounds good, will look into this as well.

ytti:
Not sure about word formal or what ever, but https://github.com/ytti/oxidized/blob/master/lib/oxidized/manager.rb#L47 could check for model/formal/x.rb and require it, if it exists.

And there would be also model/formal/formal.rb which defines the container for the formal information.

ytti:
i'm taking this in to another branch and make some, untested, changes, to show direction i'd take

ytti:
Ok I refactored it a bit, it doesn't work now as 

1. I'm not calling #formalise in model.rb
2. I'm not loading formal/name.rb in manager.rb

And dunno if it works even after doing those.

I'm still thinking how it should look like. Like should we subclass the formal or should it always be the same, probably like it is now, always the same. 

I ran out of time working on this.

PR #1694, opened 2019-01-24
make regex for enable on powerconnect less specific to fix potential …

n3wtype:
…problems with badly formatted string on some platforms

## Pre-Request Checklist

## Description
Trivial change to fix some issues encountered with few old Powerconnect 5548 models over ssh transport.

### DIFF
--- a/lib/oxidized/model/powerconnect.rb
+++ b/lib/oxidized/model/powerconnect.rb
@@ -45,7 +45,7 @@ class PowerConnect < Oxidized::Model
       if vars(:enable) == true
         cmd "enable"
       elsif vars(:enable)
-        cmd "enable", /^[pP]assword:/
+        cmd "enable", /[pP]assword:/
         cmd vars(:enable)
       end
     end


n3wtype:
These are probably some control characters as they are not reflected in output (switches which worked and the ones which didn't gave the same user readable prompt). I cannot switch to telnet due to the company policy, so I can't figure out what exactly these characters are. However, as the meaningful part of string is "[Pp]assword:" not including these within the regex shouldn't cause issues.

wk:
Thanks for looking into this. While not a risk here specifically, the benefit of trying to be as specific as possible in regexp matching is the sheer diversity of use cases. What might seem to be unambiguously a part of a password prompt to one, might in fact be part of a pre-login banner saying "Number to call if you forget your Password: +1 555 0101" to another.

Pulling this in, but please monitor the issue queue if any fellow `powerconnect.rb` users run into login trouble as a consequence; unfortunately, interacting with the PowerConnect CLI has proven to be quite flaky and inconsistent across versions.

PR #1698, opened 2019-01-28
Rebranded (hardware compatibile) devices

pedjaj:
## Pre-Request Checklist

## Description

* Allied Telesis AT-8000s and AT-8000GS series switches are  RADLAN switches, same as powerconnect and can use powerconnect.rb model.
* TP-link switch TL-SL5428 (not JetStream) with Marvell 98DX106-B0, 88E6095[F] same as EdgeCore - ES352x or SMC TigerSwitch 10/100 SMC6128L2 Manager and can use edgecos.rb model
* TP-link switch TL-SL3428 can use powerconnect.rb model.

### DIFF
--- a/docs/Supported-OS-Types.md
+++ b/docs/Supported-OS-Types.md
@@ -174,6 +174,7 @@
 * TPLink
   * [TPLink](/lib/oxidized/model/tplink.rb)
   * [TL-SL5428](/lib/oxidized/model/edgecos.rb)
+  * [TL-SL3428](/lib/oxidized/model/powerconnect.rb)
 * Ubiquiti
   * [AirOS](/lib/oxidized/model/airos.rb)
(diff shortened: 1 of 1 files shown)


wk:
Thanks for sending this in; I'll pull it as-is, but when abstracting away models, it can be quite valuable to try and determine the actual name of the underlying OS type. 

For example, if the 'Dell PowerConnect' devices actually run 'RADLAN OS', it might be worth the effort to rename the `powerconnect.rb` model to `radlan.rb`, create a temporary shim for the 'powerconnect.rb' model (see [here](https://github.com/ytti/oxidized/blob/23603127362e6f4a7606325501a3ce26ef8317a1/lib/oxidized/model/supermicro.rb) for an example), and document this change. 

This allows a clearer and more correct view of which OS types are supported, and eases maintenance and matching of new devices.

PR #1733, opened 2019-03-11
Add new model for A10 Networks TPS

alexandrecorso:
## Description
Add a new model for A10 Networks TPS (https://www.a10networks.com/products/ddos-protection-thunder-tps)

### DIFF
--- a/docs/Supported-OS-Types.md
+++ b/docs/Supported-OS-Types.md
@@ -4,6 +4,7 @@
   * OS model
 * A10 Networks
   * [ACOS](/lib/oxidized/model/acos.rb)
+  * [TPS](/lib/oxidized/model/a10tps.rb)
 * Accedian Performance Elements (NIDs)
   * [AEN](/lib/oxidized/model/aen.rb)
 * Adtran
--- a/lib/oxidized/model/a10tps.rb
+++ b/lib/oxidized/model/a10tps.rb
@@ -37,7 +37,14 @@ class A10TPS < Oxidized::Model
     new_cfg << cfg
   end
 
-  cmd 'show running-config'
+  ## Enhance configuration by removing timestamp 
+  cmd 'show running-config partition-config all' do |cfg|
+    cfg.gsub! /(Current configuration).*/, '\\1 <removed>'
+    cfg.gsub! /(Configuration last updated at).*/, '\\1 <removed>'
+    cfg.gsub! /(Configuration last saved at).*/, '\\1 <removed>'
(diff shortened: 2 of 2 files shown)


ytti:
There isn't telnet support in this model, as there insn't username/password prompt. I'd remove the :telnet tag.
The prelogout seems fragile, I'd recommend replacing it with multiple ```cmd``` which each expect some regexp, either default (prompt) or defined regexp.

wk:
Hi! Thanks for submitting this.

I notice that this model is largely similar to the `acos.rb` model, with support for partitions removed and a number of additional / different strings cleaned away as secrets and from `show version`. This suggests that both devices run the same OS type (`ACOS`).

What is the motivation for splitting this specific device into its own dedicated model instead of integrating the changes into `acos.rb`?

alexandrecorso:
Hello
A10 Networks TPS is using ACOS software. 
```
tps03# show version
Thunder Series Unified Application Service Gateway TH7445
    ...
     64-bit Advanced Core OS (ACOS) version 3.2.3-P3, build 3 (Mar-03-2019,20:04)
```

TPS is different too (it is not a load balancer). Some commands are not good on TPS.
```
tps03#show partition-config all
                            ^

tps03#show running-config all-partitions
                                   ^
```
Alexandre

wk:
@alexandrecorso could you take a look at #1760 and see if the changes suggested there also work in your use case, and how can they be reconciled into supporting all of these devices under one model?

alexandrecorso:
From my point of view, we need 2 templates. Some commands like this one, 
```
tps01.ams9.ams#show partition-config all
                             ^
```
doesn't works as in #1760

edielson:
i have an A10 th1040 CGNAT

when using configure it in oxidized I have the following error:

Apr 12 11:06:19 host.myrouter.com.br oxidized [15551]: 10.200.0.1 raised Timeout :: Error with msg "execution expired"
Apr 12 11:06:39 host.myrouter.com.br oxidized [15551]: 10.200.0.1 raised Oxidized :: PromptUndetect with msg "unable to detect prompt: (? -Mix: ^ ([- \ w. \ /: ? \ [\] ()] + [#>] \ s?) $) "
Apr 12 11:06:39 host.myrouter.com.br oxidized [15551]: a10tps / QNETFACIL-A10 status no_connection, retry attempt 1
Apr 12 11:07:05 host.myrouter.com.br oxidized [15551]: 10.200.0.1 raised Timeout :: Error with msg "execution expired"



in file router.db

QNETFACIL-A10:10.200.0.1:a10tps:a10tps

mortzu:
Please process the checklist

Issue #1742, opened 2019-03-15
Missing dependencies for net-ssh


aneagoe:
Running docker container version 0.25.1 fails with following error
```oxidized_1  | *** Running /etc/my_init.d/00_regen_ssh_host_keys.sh...
oxidized_1  | *** Running /etc/my_init.d/10_syslog-ng.init...
oxidized_1  | Mar 15 13:32:11 4136e72fe801 syslog-ng[15]: syslog-ng starting up; version='3.13.2'
oxidized_1  | *** Booting runit daemon...
oxidized_1  | *** Runit started as PID 24
oxidized_1  | Mar 15 13:32:12 4136e72fe801 cron[33]: (CRON) INFO (pidfile fd = 3)
oxidized_1  | Mar 15 13:32:12 4136e72fe801 cron[33]: (CRON) INFO (Running @reboot jobs)
oxidized_1  | I, [2019-03-15T13:32:13.080593 #31]  INFO -- : Oxidized starting, running as pid 31
oxidized_1  | I, [2019-03-15T13:32:13.082018 #31]  INFO -- : lib/oxidized/nodes.rb: Loading nodes
oxidized_1  | I, [2019-03-15T13:32:13.173443 #31]  INFO -- : lib/oxidized/nodes.rb: Loaded 2 nodes
oxidized_1  | Puma starting in single mode...
oxidized_1  | * Version 3.11.4 (ruby 2.5.1-p57), codename: Love Song
oxidized_1  | * Min threads: 0, max threads: 16
oxidized_1  | * Environment: development
oxidized_1  | * Listening on tcp://127.0.0.1:8888
oxidized_1  | Use Ctrl-C to stop
oxidized_1  | #<Oxidized::Job:0x000055deb4830db8@/var/lib/gems/2.5.0/gems/oxidized-0.26.3/lib/oxidized/job.rb:7 run> terminated with exception (report_on_exception is true):
oxidized_1  | /var/lib/gems/2.5.0/gems/net-ssh-5.2.0/lib/net/ssh/authentication/ed25519_loader.rb:21:in `raiseUnlessLoaded': OpenSSH keys only supported if ED25519 is available (NotImplementedError)
oxidized_1  | net-ssh requires the following gems for ed25519 support:
oxidized_1  |  * ed25519 (>= 1.2, < 2.0)
oxidized_1  |  * bcrypt_pbkdf (>= 1.0, < 2.0)
[...]

PR #1743, opened 2019-03-15
Fixed dependencies for net-ssh (#1742)

aneagoe:
## Description
Fixes dependencies for net-ssh when using ssh keys.
Closes issue #1742

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * MISC: add pgsql support, mechanized and net-tftp to Dockerfile
 * MISC: upgrade slop, net-telnet and rugged
+* FIX: add dependencies for net-ssh
 
 ## 0.26.3
 
--- a/oxidized.gemspec
+++ b/oxidized.gemspec
@@ -19,8 +19,8 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version =           '>= 2.3'
   s.add_runtime_dependency 'asetus',  '~> 0.1'
-  s.add_runtime_dependency 'bcrypt_pbkdf', '>= 1', '< 2'
-  s.add_runtime_dependency 'ed25519', '>= 1.2', '< 1.3'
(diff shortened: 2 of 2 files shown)


wk:
This looks good; surprising ed25519 adoption is sufficiently low that it took a few days since the update of the `net-ssh` dependency from 4.1 to 5.x for this to surface.

Could you add a change line to the `CHANGELOG.md` file as part of this PR before it's pulled in?

aneagoe:
Will do it shortly. Also, it doesn't matter whether the key is ed25519 or plain rsa, this will still trigger.

ytti:
The version specification seems too conservative compared to rest of our version specs.

wk:
Per [net-ssh](https://rubygems.org/gems/net-ssh/versions/5.2.0), the actual dependencies are:

```
bcrypt_pbkdf ~> 1.0 
ed25519 ~> 1.2
```

@aneagoe any reason you've decided to narrow them further than this in your PR?

aneagoe:
@wk I thought I took dependencies from the error output, but now I noticed a mismatch. I'll update this, sorry for the inconvenience.
```
oxidized_1  | net-ssh requires the following gems for ed25519 support:
oxidized_1  |  * ed25519 (>= 1.2, < 2.0)
oxidized_1  |  * bcrypt_pbkdf (>= 1.0, < 2.0)
```

wk:
In this case, these are just two different ways to express the same thing, with the `~>` operator allowing to do so slightly more concisely.

aneagoe:
@wk Right. Had no idea what twiddle:wakka was until now. I've fixed and pushed.

wk:
New container with these changes in has been built at `oxidized/oxidized`, please confirm if this has resolved the issue in your case.

aneagoe:
I just pulled and it all looks good. Would be great if version tags were added to dockerhub so I don't use latest in my configs. Thanks a lot for such a quick turnaround!

ytti:
Docker does have versions, we've just not done release with this fix.

Issue #1740, opened 2019-03-14
Cambium Canopy backups causes constant diff


cchance27:
When performing backups from cambium canopy devices every time the backup runs it creates a new diff this seems to be due to a line in the config that shows the timestamp the config was pulled.

The cfgUtcTimestamp changes constantly, if we can exclude that from the config backup it would solve the issue.

```
@@ -498,7 +498,7 @@
    "priorityPrecedence": 0
   }
  },
+ "cfgUtcTimestamp": "2019-03-14T18:33:33Z",
  "configFileParameters": {
   "rebootIfRequired": false,
   "setToDefaults": false
```

ytti:
Copy the model into ~/.config/oxidized/model/ directory and modify it to  change the timestamp to say NNNNN.

something like

```ruby
cfg = JSON.parse(cfg)
cfg["cfgUtcTimestamp"] = "0000-00-00T00:00:00Z" if cfg.has_key?("cfgUtcTimestamp")
JSON.pretty_generate(cfg)
```

or 

```ruby
cfg.each_line.map do |line|
   key, value = line.split(": ")
  if key == '"cfgUtcTimestamp"'
    key + ": " + "0000-00-00T00:00:00Z"
  else
    line
  end
end
```

cchance27:
hmmm odd I tried that but it seemed to somehow break the backups completely like the model stopped updating even though I made a change to its configuration, removed the cambium.rb from the /model directory and restarted the docker instance and it started backing up everything again (literally everything because all the timestamps changed)

ytti:
I wrote those as inspiration, not as tested working code. So it wouldn't seem odd to me if there is problem in the  code or if it was misapplied.

PR #1746, opened 2019-03-18
Cambium timestamp correction

cchance27:
## Pre-Request Checklist

## Description
Just added a gsub to drop the timestamp from configuration as it changes every backup

Closes issue #1740


ytti:
I would have probably just changed the timestamp to all zeroes. Removing the key from the config may mean that some parsers which want the key to exist break.

But ultimately, you use the model, I don't.

cchance27:
Ya that was an option, but in reality recommended restoration is actually to remove timestamp and even some other fields that can break restore, so decided to just drop it.

PR #1750, opened 2019-03-19
Add support for BOSS Extreme switches to have spaces in sys name

ospfbgp:
## Pre-Request Checklist

### DIFF
--- a/lib/oxidized/model/boss.rb
+++ b/lib/oxidized/model/boss.rb
@@ -7,7 +7,7 @@ class Boss < Oxidized::Model
 
   comment  '! '
 
-  prompt /^[^\s#>]+[#>]$/
+  prompt /^\(?.+\)?(\s?)[#>]/
 
   # Handle the banner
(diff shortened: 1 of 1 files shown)


ospfbgp:
needs more testing do not approve this pull request thanks.  Found bug. :-)

wk:
Does the prompt in these models appear multiple times or at unexpected intervals within the output of `show running-config`? If the intention is just to remove the prompt at the end of the config, perhaps a `cut_tail` would be more suitable?

ospfbgp:
I need to do some more testing.  I was traveling this week.  I will look into this more.  Thank you for the suggestions.  I will do some testing and debugging and see what I get.

wk:
@ospfbgp did you have a chance to see if this can be cleaned up? I'd like to pull this in, but lacking access to a similar device or a sample of the unfiltered output, can't confirm if the suggestion above is feasible.

mortzu:
Please process the checklist

PR #1756, opened 2019-03-22
List unsaved changes on the Cisco switch.

dubbelj:
This will list if there is any diff to running-config and startup-config.
But will ignore some diff that are normal, like "ntp clock-period" and certificate as they are read from file in startup and shown in running.

## Pre-Request Checklist

### DIFF
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -103,6 +103,21 @@ class IOS < Oxidized::Model
     comment cfg
   end
 
+  # Show if there is any unsaved changes on the switch.
+  cmd 'show archive config diff' do |cfg|
+    # Print diff unless ntp period change or ssl-cert read from file
+    cfg.gsub! /^\n/, '' # Remove empty line
+    cfg.gsub! /^!\n/, '' # Remove line with only !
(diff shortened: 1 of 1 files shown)


wk:
Thanks for this! 

This is a behavior modification that affects one of the most used models in `oxidized`. Typically, oxidized models archive only the current running configuration. For many operators, unsaved config may be irrelevant for archiving and considered ephemeral (as in the event of a device reboot it will be gone for good). It's also quite easy for a regular process (e.g. `oxidized`) to catch a human operator mid-action, triggering a set of changes as the human operator is working on them.

Furthermore, it is unclear if this particular command is indeed available and consistent across the entire range of devices and versions this model supports as-is.

While I see the utility (identifying if changes have been made and not saved, and perhaps triggering follow-up action resulting from the diff to rectify this situation; or perhaps in a lab environment, to save 'snapshots' of on-going work) - it is unclear if this is a functionality that's directly applicable and desirable for the entire user base.

Would you consider instead to document this as a possible model extension within a model notes document for IOS? That would allow anyone who is interested in this functionality to quickly add it by extending the `ios.rb` model without modifying the default behavior.

dubbelj:
Thanks for answer. I find it quite usefull myself. So I will look into documenting that in model specific docs. And do I understand that you are quite careful with changes in ios.rb.

PR #1759, opened 2019-03-26
Documentation to add feature to list unsaved changes to a Cisco switch

dubbelj:
## Pre-Request Checklist

### DIFF
--- a/docs/Model-Notes/README.md
+++ b/docs/Model-Notes/README.md
@@ -13,6 +13,7 @@ Arista|[EOS](EOS.md)|05 Feb 2018
 Cumulus|[Cumulus](Cumulus.md)|11 Jun 2018
 Huawei|[VRP](VRP-Huawei.md)|17 Nov 2017
 Huawei|[SmartAX series](SmartAX-Huawei.md)|21 Jan 2019
+Cisco IOS|[IOS](ios.md)|26 Mar 2019
 Juniper|[MX/QFX/EX/SRX/J Series](JunOS.md)|18 Jan 2018
 Netgear|[Netgear](Netgear.md)|11 Apr 2018
 Nokia|[Nokia ISAM](Nokia.md)|22 Aug 2018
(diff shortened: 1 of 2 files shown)


wk:
This looks great! Did you know that [extending existing modules](https://github.com/ytti/oxidized/blob/master/docs/Creating-Models.md) can actually be even simpler? There's no need to override the upstream module that's part of the gem! Simply create the file `~/.config/oxidized/model/ios.rb` as part of your Oxidized configuration, and extend it by first loading the existing model, and then introducing the new command; in your case, something like this:

```ruby
require 'oxidized/model/ios.rb'

class IOS

  cmd 'show archive config diff' do |cfg|
    # Print diff unless ntp period change or ssl-cert read from file
    cfg.gsub! /^\n/, '' # Remove empty line
    cfg.gsub! /^!\n/, '' # Remove line with only !
    cfg.gsub! /.*ntp clock-period \d+\n/, '' # Remove line with only "ntp clock-period blabla"
    cfg.gsub! /\n/, "\\n" # Escape newline
    cfg.gsub! /crypto pki certificate chain.*certificate .*\.cer\\n/, '' # Remove ssl-cert in start config, as it is read from file, this always differ in running if used.
    cfg.gsub! /crypto pki certificate chain.*-\s*quit\\n/, '' # Remove ssl-cert from running
    cfg.gsub! /\\n/, "\n" # Set newline back
    unless cfg == "!Contextual Config Diffs:\n" # Do not print if only something above was changed
      comment cfg
    end
  end

end
```

Reload Oxidized, et voilà!

wk:
Mentioning this because the PR makes a reference to editing `/var/lib/gems/RUBY-VERSION/gems/oxidized-VERSION/lib/oxidized/model/ios.rb`, which would be overwritten upon upgrade, whereas the alternative way would not. 

If this piece can be updated, this is good to go in!

dubbelj:
Thanks! This was a great tip that I did not know about and will help me as I had this problem myself that it is written over every upgrade. I will update the docs.

wk:
Merged, thanks!

PR #1760, opened 2019-03-26
Update acos.rb

GeneralFox:
modified for error on ACOS 4.x
! COMMAND: show running-config all-partitions
                                    ^

## Pre-Request Checklist

### DIFF
--- a/lib/oxidized/model/acos.rb
+++ b/lib/oxidized/model/acos.rb
@@ -37,7 +37,7 @@ class ACOS < Oxidized::Model
     cfg
   end
 
-  #FOR ACOS 4.x
+  # change for ACOS 4.x
   cmd 'show running-config partition-config all' do |cfg|
     cfg.gsub! /(Current configuration).*/, '\\1 <removed>'
     cfg.gsub! /(Configuration last updated at).*/, '\\1 <removed>'
--- a/lib/oxidized/model/acos.rb
+++ b/lib/oxidized/model/acos.rb
@@ -37,7 +37,8 @@ class ACOS < Oxidized::Model
     cfg
   end
 
-  cmd 'show running-config all-partitions' do |cfg|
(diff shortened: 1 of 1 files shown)


wk:
Is this change backwards compatible with older ACOS? Presumably the command has worked for some device / version of ACOS. There's also #1733 where a user suggests removing the same command completely (rather than replacing it) in order to make this work with a different ACOS device / model / firmware. 

Perhaps you could, together with @alexandrecorso figure out which exact devices/versions no longer support the old command and require the new version and see if the best approach here is to introduce a new `acos.rb`, transition the current `acos.rb` into `acos-legacy.rb` or similar and override it, patch detection to execute the correct command into the model, or some other solution?

alexandrecorso:
Hello,
I checked with my device, almost every command works except:
```
tps01.ams9.ams#show partition-config all
                             ^
```
On the other hand, ```show aflex ....``` commands is not for Thunder TPS product. It works because A10 Networks use mutual features for different product.

We will definitely need two templates.
Alexandre

wk:
@alexandrecorso thank you for the additional detail!
@GeneralFox can you confirm if the changes here also work on versions under 4.x, or if the change is breaking?

alexandrecorso:
Can we merge the branch?

GeneralFox:
> @alexandrecorso thank you for the additional detail!
> @GeneralFox can you confirm if the changes here also work on versions under 4.x, or if the change is breaking?

The change works on Acos 4 in fact it was done specifically for that release.
Maybe in future we need to do have different template for A10 products.
@alexandrecorso for me yes :)

wk:
@cppmonkey @broxio, anyone who uses this model with the older (3.x ?) version who can confirm this works both for old *and* new version, and not only the new one?

mortzu:
Please process the checklist

PR #1764, opened 2019-03-29
Explicit stringio to string type conversion in nagios check

Kegeruneku:
## Pre-Request Checklist

## Description
The script crashes at least on Ruby 2.3 due to a denied implicit conversion during JSON parsing:
```
/usr/lib/ruby/2.3.0/json/common.rb:156:in `initialize': no implicit conversion of StringIO into String (TypeError)
	from /usr/lib/ruby/2.3.0/json/common.rb:156:in `new'
	from /usr/lib/ruby/2.3.0/json/common.rb:156:in `parse'
	from ./check_oxidized_failing_nodes.rb:13:in `<main>'
```

I added a `.string` explicit conversion to prevent the error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * BUGFIX: dlink model didn't support prompts with spaces in the model type (Extreme EAS 200-24p) (@cchance27)
 * BUGFIX: routeros model does not collect configuration via telnet input (@hexdump0x0200)
 * BUGFIX: add dependencies for net-ssh
+* BUGFIX: crash on some recent Ruby versions in the nagios check (@Kegeruneku)
 * MISC: add pgsql support, mechanized and net-tftp to Dockerfile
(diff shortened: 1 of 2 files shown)


wk:
Thanks for this! Is there any reason you've opted for the `.string` conversion instead of the `.read` method to resolve this?

Kegeruneku:
Hello @wk , to be honest I am rather unexperienced at Ruby so I chose the most straightforward approach after scraping some related issues on other repos.

I'll gladly change the PR if you prefer another way! :-)

Kegeruneku:
Updated to use `.read` instead, as it seems to work just as fine.

wk:
`.string` and `.read` should, in this instance, be essentially identical but `.read` is more contextually suitable in this case and makes it easier to understand the intent. Thanks for validating it and addressing the bug!

PR #1793, opened 2019-04-30
Added support for reloading just a single node

laf:
## Pre-Request Checklist

## Description
This will accept a single node and then reload just the source for that single node where the source backend supports it (http and sql will).

We have been using this code for 6 months now to allow us to add / update a single node without reloading the entire source. This cuts reload times down from minutes to seconds when you have 1,000's of nodes in your source list.

My Ruby is very limited so please feel free to judge me, please make some recommendations though and I'll try and work through them :)

Requires: https://github.com/ytti/oxidized-web/pull/189

### DIFF
--- a/lib/oxidized/nodes.rb
+++ b/lib/oxidized/nodes.rb
@@ -20,10 +20,10 @@ module Oxidized
           begin
             node_obj = Node.new node
             new.push node_obj
-          rescue ModelNotFound => err
-            Oxidized.logger.error "node %s raised %s with message '%s'" % [node, err.class, err.message]
-          rescue Resolv::ResolvError => err
-            Oxidized.logger.error "node %s is not resolvable, raised %s with message '%s'" % [node, err.class, err.message]
+          rescue ModelNotFound => e
+            Oxidized.logger.error "node %s raised %s with message '%s'" % [node, e.class, e.message]
+          rescue Resolv::ResolvError => e
+            Oxidized.logger.error "node %s is not resolvable, raised %s with message '%s'" % [node, e.class, e.message]
           end
         end
         size.zero? ? replace(new) : update_nodes(new)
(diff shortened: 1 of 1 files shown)


laf:
I've made some changes to resolve the travis issue but I could do with help on the others if possible?

wk:
Hi @laf 

I'm a bit confused at what is happening here; as far as I can tell, the existing load() function already has support for loading a single node - if the node is passed an argument to do so. The new function seems to be largely replicating the same functionality exactly, but makes the argument required. 

Why is this necessary, and why didn't simply passing the name of the desired node to load() work for you?

miff2000:
Hi @wk

I'm going to try and pick up on this PR :)

I think I see what you're referring to in your comment above. I've excluded the code in this PR, but made the modifications in the other PR (https://github.com/ytti/oxidized-web/pull/189), making it send the reload to `node.load()` instead of `node.load_node()`. If seems to work for me.

Is that what you were referring to? If so, if just amend the other PR to use the `node.load()` method and close down this PR, would you be happy to accept that?

wk:
Indeed, since the intent behind existing node.load() with node_want argument and the proposed node.node_load() appears to be identical, this PR does not seem essential to  facilitate the desired functionality, and thus can be closed.

The web side changes should then indeed be rebased on the existing functionality, and I see no reason why they couldn't go in after testing.

miff2000:
Perfect, sounds good to me. Thanks for your help!

miff2000:
New PR created for the changes on the oxidized-web side (<https://github.com/ytti/oxidized-web/pull/206>), as I don't have access to amend or close the other PR. I'll pick up on that tomorrow with laf.

wk:
Closing this as identical functionality has been added in ytti/oxidized-web#206.

Issue #1785, opened 2019-04-25
Cisco SF350 (ciscosmb): remove "Unit Temperature" and "Fans Status" from the output


igloo777:
Hi!
Output of _show system_ in software 'Version: 2.4.5.71' return the next data, that can change quite often, doesn't relate to config and shows the useless diff between the number of config backups:

```
Unit                Fans Status             
----   -------------------------------------
 1           Fans status is N/A          

Unit Temperature (Celsius)   Status   
---- --------------------- ---------- 
 1            41               OK    
```

I think it should be removed from the output of ciscosmb.rb module.

wk:
Thanks for the report! Should be resolved by #1799.

Issue #1500, opened 2018-08-30
CiscoSMB false positives from “show bootvar”


quacktacular:
Recently I’ve been seeing an issue where the actual command “show bootvar” will be present in the diff instead of the actual output. This is almost always resolved on the next scan but produces a lot of notifications.

Happening on multiple SG500 switches. Short of removing the command from the model how can I troubleshoot further?

davama:
could you provide a diff?
More info would be helpful

This is how my `sh bootvar` looks: (cisco sf300/302)
```
sh bootvar 
Image  Filename   Version     Date                    Status
-----  ---------  ---------   ---------------------   -----------
1      image-1    1.3.7.18    12-Jan-2014  18:02:59   Active* 
2      image-2    1.3.7.18    12-Jan-2014  18:02:59   Not active

"*" designates that the image was selected for the next boot
```

quacktacular:
Sure no problem:
```
Version 155 (3 days 4 hours ago)
diff --git a/sw1.me.acme.com b/sw1.me.acme.com
index bc9b76f..bb923a0 100644
--- a/sw1.me.acme.com
@@ -4,8 +4,7 @@
 !          1                1.4.8.6            1.4.0.02               V05         
 !          2                1.4.8.6            1.4.0.02               V05         
 ! 
+! show bootvar
+! 
Unit  Image  Filename   Version    Date                    Status
 ! ----  -----  ---------  ---------  ---------------------   -----------
 ! 1     1      image-1    1.4.5.02   20-Apr-2016  12:24:28   Not active
 ! 1     2      image-2    1.4.8.6    10-Jul-2017  17:07:52   Active* 
```
Also do you know wether there's a way to suppress port status updates like this? Also causing notifications sometimes. 
```
Version 145 (1 hours 50 min ago)
diff --git a/sw5.me.acme.com b/sw5.me.acme.com
index 0e439c4..854e667 100644
--- a/sw5.me.acme.com
@@ -1,4 +1,3 @@
+! 30-Aug-2018 05:55:52 %LINK-W-Down:  gi1/1/44
 ! 
 !        Unit             SW version         Boot version         HW version      
 ! ------------------- ------------------- ------------------- ------------------- 
```

davama:
That's interesting.

So this gets added/removed on every `sh bootvar` run?
```
+! show bootvar
+! 
```

> Also do you know wether there's a way to suppress port status updates like this? Also causing notifications sometimes.

[...]

Issue #1398, opened 2018-06-14
CiscoSMB false positive diffs / notifications based on uptime


quacktacular:
Hey! Updated oxidized today (0.24.0) and I now get notifications every time the check runs.

Looks like its not filtering out uptime:
```
-!  1    175,18:24:38   
-!  2    175,18:27:43   
+!  1    175,18:26:21   
+!  2    175,18:29:26   
```
How can I sort it out?

quacktacular:
This is still an issue on multi unit swithces:
```
! Unit     Up time     
! ---- --------------- 
-!  1    175,20:32:50   
-!  2    175,20:35:24   
-!  3    175,20:37:30   
-!  4    175,20:36:58   
+!  1    175,20:45:31   
+!  2    175,20:48:04   
+!  3    175,20:50:10   
+!  4    175,20:49:38   
```

quacktacular:
@wk also still include temperature information. So it will also notify / diff on temp change.
```
! Unit   Temperature (Celsius)   Temperature Sensor Status 
! ---- ------------------------- ------------------------- 
!  1              47                        OK             
!  2                                        N/A            
!  3              46                        OK             
!  4                                        N/A            
! 
! 
! Unit     Up time     
! ---- --------------- 
!  1    175,19:25:32   
!  2    175,19:28:05   
!  3    175,19:30:11   
!  4    175,19:29:39   
```
I think the "show system" is unnecessary. It doesn't provide much utility (for backups anyway), and it wasn't included before 0.22.0.

wk:
Based on your included output extract, it's apparent this isn't the style output that the regexp has been written to catch, as it is dissimilar from the output of devices provided by users based on which it was written. 

For system uptime, in #1364, the output looks like so:
`!  System Up Time:          49 days, 20 hours, 57 minutes, and 4.47 seconds`

For temperature, in #1394, the output looks like so:
```
!
! System Temperature:
! Unit 1
! Temperature 1: 35 degrees
```

The output above appears to be tabulated and differently styled (as opposed to my initial guess at a partial match). Could you specify the exact model and configuration on which you observe this behavior via e.g. a full output of `show version` and `show system`?

wk:
Earlier reply contained incorrect information based on conflation of two similar but unrelated issues (now removed).

[...]

PR #1799, opened 2019-05-03
Ciscosmb: remove 'sh system' output

Exordian:
## Pre-Request Checklist

## Description
'show system' on a Cisco SG500 emits the following output:
```
System Description:                       SG500-52P 52-Port Gigabit PoE Stackable Managed Switch
System Up Time (days,hour:min:sec):       102,20:25:15
System Contact:                           X
System Name:                              X
System Location:                          X
System MAC Address:                       XX:XX:XX:XX:XX:XX
System Object ID:                         1.3.6.1.4.1.9.6.1.81.52.2

Fans Status:                              OK


          Unit            Temperature (Celsius)            Status          
------------------------ ------------------------ ------------------------ 
           1                        32                       OK            
```

None of these informations are relevant for a backup system IMO. The ever-changing temperature triggers unneeded 'config' changes on oxidized.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,13 @@
 * FEATURE: add Ubiquiti Airfiber model support (@cchance27)
 * BUGFIX: voss model
 * BUGFIX: cambium model should not consider timestamp for backup as unneeded, and causes diffs (@cchance27)
-* BUGFIX: remove 'sh system' from ciscosmb model
+* BUGFIX: remove 'sh system' from ciscosmb model (@Exordian)
 * BUGFIX: dlink model didn't support prompts with spaces in the model type (Extreme EAS 200-24p) (@cchance27)
 * BUGFIX: routeros model does not collect configuration via telnet input (@hexdump0x0200)
 * BUGFIX: add dependencies for net-ssh
 * BUGFIX: crash on some recent Ruby versions in the nagios check (@Kegeruneku)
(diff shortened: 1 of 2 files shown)


davama:
Could similar argument (monkey wrench) be made here about need for `show system` and other cmds?

Exordian:
> Could similar argument (monkey wrench) be made here about need for `show system` and other cmds?

I agree. I removed the whole command. Location, contact and so can be found in the config as well. I don't think the model type and mac address is needed in oxidized.

wk:
As this is causing a lot of users considerable issues (#1785, #1398, #1500) - and as the output is highly variable across different models and configurations (stacked / unstacked) and fairly complex, removing `show system` and `show bootvar` seems to be the way forward here.

@phurrelmann, I notice you've thumbed down this PR - do you have an alternative approach or any specific concerns?

@Exordian, @davama - do we have consensus on this being the way forward for this model?

wk:
On a deeper review, it seems #1500 had only a single affected user which found a workaround, so `show bootvar` can stay for now.

davama:
I agree 👍

menvol3:
Hi

Could you please add to the following file support for enable command, like in ios.rb

davama:
That is outside this PR.
You could create one with those specifications, it’s not that challenging.
Give it a try, ppl here will let you know if all is well 👍

PR #1807, opened 2019-05-15
Strip line "# received packet from x.x.x.x bad format" from RouterOS

candlerb:
## Pre-Request Checklist

Minimal change

## Description
Strip line `# received packet from x.x.x.x bad format` from RouterOS config.

This message intermittently appears and disappears when RouterOS is running VRRP and there is also CARP traffic on the same network.

The two protocols work side-by-side without problem if using different group IDs, but share the [same IP protocol number](https://en.wikipedia.org/wiki/Common_Address_Redundancy_Protocol#Incompatibility_with_IANA_standards) and this causes Mikrotik to (sometimes) include this warning in configs.

### DIFF
--- a/lib/oxidized/model/routeros.rb
+++ b/lib/oxidized/model/routeros.rb
@@ -29,6 +29,7 @@ class RouterOS < Oxidized::Model
     cmd run_cmd do |cfg|
       cfg.gsub! /\\\r?\n\s+/, '' # strip new line
       cfg.gsub! /# inactive time\r\n/, '' # Remove time based system comment
+      cfg.gsub! /# received packet from \S+ bad format\r\n/, '' # Remove intermittent VRRP/CARP collision comment
       cfg = cfg.split("\n").reject { |line| line[/^#\s\w{3}\/\d{2}\/\d{4}.*$/] }
       cfg.join("\n") + "\n"
     end


wk:
Are there any legitimate comments (lines beginning with `#`) which are part of RouterOS config output? Could perhaps all lines starting with `#` be stripped out?

Alternatively, is there perhaps a console setting command present in RouterOS which Oxidized could run after login in order to disable log output to the current terminal / console?

candlerb:
> Are there any legitimate comments (lines beginning with #) which are part of RouterOS config output? Could perhaps all lines starting with # be stripped out?

There are some legitimate comments at the top of `/export`, e.g.

```
# may/16/2019 08:02:54 by RouterOS 6.43.16
# software id = XXXX-XXXX
#
# model = CCR1036-8G-2S+
# serial number = XXXXXXXXXXXX
```

> Alternatively, is there perhaps a console setting command present in RouterOS which Oxidized could run after login in order to disable log output to the current terminal / console?

It's actually part of the generated config - it's not just random log output mixed in with the config.  When you do `/export`, the comment (when it appears) is always in exactly the same place.

The problem is that with this VRRP/CARP issue, when the router is acting in VRRP backup role, the comment appears and disappears randomly - if you repeatedly do `/export` then sometimes it's there and sometimes it isn't.  As a result, we get E-mailed every few hours.

(When the router is VRRP master role, then the comment is there permanently, so it doesn't cause an issue)

wk:
Thank you for the additional detail. From the initial description, my guess was that the VRRP/CARP related comments are situational output by the VRRP/CARP process that is logged into the console (while the configuration is being dumped) which is included in the output - and thus the idea of categorically cleaning any (other) log messages as well. From your explanation, this does not appear to be the case here.

Pulling this as-is!

PR #1808, opened 2019-05-16
Create fastiron.rb

ZacharyPuls:
Add support for Ruckus (Brocade) ICX-series switches.

Tested on an ICX7750-48F running FastIron 08.0.30mT203

## Pre-Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * FEATURE: include licensing information in aos model (@pozar)
 * FEATURE: add firelinuxos (FirePOWER) model (@rgnv)
 * FEATURE: add sonicos model (@rgnv)
+* FEATURE: add FastIron model (@ZacharyPuls)
 * BUGFIX: voss model
 * BUGFIX: cambium model should not consider timestamp for backup as unneeded, and causes diffs (@cchance27)
 * BUGFIX: dlink model didn't support prompts with spaces in the model type (Extreme EAS 200-24p) (@cchance27)
--- a/docs/Supported-OS-Types.md
+++ b/docs/Supported-OS-Types.md
@@ -38,7 +38,7 @@
 * Brocade
   * [FabricOS](/lib/oxidized/model/fabricos.rb)
   * [FastIron] (/lib/oxidized/model/fastiron.rb)
-  * [Ironware](/lib/oxidized/model/ironware.rb)
+  * [IronWare](/lib/oxidized/model/ironware.rb)
   * [NOS (Network Operating System)](/lib/oxidized/model/nos.rb)
   * [Vyatta](/lib/oxidized/model/vyatta.rb)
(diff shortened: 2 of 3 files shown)


wk:
Looks good! What is the reason for processing `show version` in this model, instead of either keeping it commented as-is, or removing known volatile fields from it?

ZacharyPuls:
> 
> 
> Looks good! What is the reason for processing `show version` in this model, instead of either keeping it commented as-is, or removing known volatile fields from it?

Just keeping it in line with the ios.rb version processing, as well as the output being a bit more verbose than I'd like. So it looks like this:

! Version: 08.0.30mT203 
! Boot-Monitor Version: 10.1.05T205 (swz10105)
! Serial: CRH3350M0CF

As opposed to this:

  Copyright (c) 1996-2016 Brocade Communications Systems, Inc. All rights reserved.
    UNIT 1: compiled on Jan 24 2017 at 23:48:59 labeled as SWR08030m
      (21156985 bytes) from Secondary SWR08030m.bin
        SW: Version 08.0.30mT203 
      Compressed Boot-Monitor Image size = 1835008, Version:10.1.05T205 (swz10105)
  HW: Stackable ICX7750-48F
  Internal USB: Serial #: 40016BD285FC0D89
      Vendor: UB90PTC, Total size = 1910 MB
==========================================================================
UNIT 1: SL 1: ICX7750-48XGF 48-port Management Module
      Serial  #:CRH3350M0CF
      License: ICX7750_L3_SOFT_PACKAGE   (LID: etjIIKFoFeh)
      P-ASIC  0: type B852, rev 03  Chip BCM56852_A2
==========================================================================
UNIT 1: SL 2: ICX7750-QSFP 6-port QSFP 240G Module
==========================================================================
 1500 MHz Power PC processor (version 8023/0022) 88 MHz bus
 8192 KB boot flash memory
 2048 MB code flash memory
 3840 MB DRAM
STACKID 1  system uptime is 781 day(s) 30 minute(s) 28 second(s) 
The system started at 00:07:24 Central Fri Mar 31 2017

The system : started=cold start                                   
My stack unit ID = 1, bootup role = alone

Now that I look at it, it also includes the uptime in the 'show version' output, which is not ideal.

wk:
Thanks for the detail! 

I'm not absolutely sure this is the desired behavior in either case, since information is being unnecessarily lost after taking the effort to run the command. This information might be useful to someone building an inventory (and thus having an interest serial numbers), or attempting to repurchase hardware to meet a specific configuration's requirements.

Generally, scrubbing output should be reserved for cases where the information might be volatile between two executions, in which case, only the volatile information should typically be removed (i.e. uptime in this instance).

Pulling this as-is, but should possibly be reconsidered in both `ios` and here.

ZacharyPuls:
Thank you for the pull!

Let me know if you need help working the show version output in `ios`, I have a wide array of IOS/IOS-XE devices to test with.

PR #1809, opened 2019-05-16
HP MSM Wireless Controller support

timwsuqld:
## Pre-Request Checklist

## Description
Added support for the HP MSM Wireless Controller. Tested on a HP MSM720

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * FEATURE: include licensing information in aos model (@pozar)
 * FEATURE: add firelinuxos (FirePOWER) model (@rgnv)
 * FEATURE: add sonicos model (@rgnv)
+* FEATURE: add hpmsm model (@timwsuqld)
 * BUGFIX: voss model
 * BUGFIX: cambium model should not consider timestamp for backup as unneeded, and causes diffs (@cchance27)
 * BUGFIX: remove 'sh system' from ciscosmb model (@Exordian)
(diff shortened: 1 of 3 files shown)


wk:
Looks good! I've added an attribution line for the change to `CHANGELOG.md`, and applied the recommended Rubocop fix that prevented this PR from passing tests.

Could you add some detail as to why the `show all config` is being pre-processed instead of taken as-is? Is there any reason to discard the additional information? Presumably `Who` and `When` lines don't change unless the configuration itself does...

timwsuqld:
Thanks for looking at it @wk 

What was kept/discarded for `show all config` was based on our script we wrote for rancid.

The Who/When lines are from the user that runs the `show all config` and so the `When` changes each run as it's a timestamp, and the `Who` changes each time as it includes the port you're connecting from.
```
# Who:  admin@192.168.0.2 60364 22
# When: Mon May 20 07:55:32 2019
```

The igmp proxy lines contain funny characters that also change between runs. I don't know why, but we excluded it in rancid, and in oxidized it also changes between each run, so I added an exclusion for it.

Did you want me to add comments to the class file as to why it's excluded?

wk:
Thanks for the detail! I'd suggest reworking the cleanup section for ease of comprehension along these lines (pesudocode, untested):

```ruby
cmd 'show all config' do |cfg|
  cfg = cfg.each_line.reject { |line| line.match /^running configuration:/ }.join
  cfg = cfg.each_line.reject { |line| line.match /(^#\s+Who:)|(^#\s+When:)/ }.join
  cfg = cfg.each_line.reject { |line| line.match /^[ \t]*igmp proxy (upstream|downstream)/ }.join
  cfg
end
```

A comment around the igmp line explaining why it is removed might also benefit users, as it's generally expected that a user can recover from an archived configuration back to the original state - so if some elements are omitted breaking that promise, it's great to document them.

timwsuqld:
@wk Thanks for that. I've made updates suggested.

wk:
Great, thanks for the touch-ups!

Issue #1818, opened 2019-05-29
Propose Feature Request: Generic Linux Module for Configs


davama:
Hello,

One of my backburner projects is to backup all critical configurations for our different linux servers. I was planning to write up a simple bash script to do it but then i wondered, "does oxidized already do this??" :thinking: 

Not that they change a lot but if something it did, we can see the change and restore if need be.

Was looking at the [culumus](https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/cumulus.rb) module which has something similar to what i was thinking.

Would it be ok for me to submit something similar but for Linux in general? I could call it `linuxgeneric`

Of course my org needs differ so ill make it generic to start with. 

The details could be hashed out later but just wanted some opinions.

Thoughts??

Thanks,
dave

wk:
There is a considerable breadth to what falls under "Linux server", and may include hundreds or even thousands of packages and configuration files which further differ in availability and names across different distributions and versions.

Network equipment and appliance vendors typically hide much of this complexity behind a vendor provided distribution that includes an opinionated, default, configuration for the operating system - and then expose a small purpose-built user-serviceable configuration aperture limited to the scope of its supported features.

You might want to consider inverting the model and embracing declarative configuration management via a tool such as ansible, salt, puppet or a declarative distribution such as nixos. These tools include a broad range of high and low level abstractions to deal with common problems in this domain, and would result in there being a single repository of authoritative configurations. Nearly all of these tools also offer some functionality to validate the "compliance" of a specific server to its declared configuration in order to identify and rectify configuration drift (i.e., "diffs" in the style of Oxidized).

Alternatively, you might want to consider traditional backup tools for these Linux servers which will include both configuration and data (or even complete partitions, volumes, and disks) allowing for a complete recovery of all data.

davama:
@wk thank you for the reply

> There is a considerable breadth to what falls under "Linux server", and may include hundreds or even thousands of packages and configuration files which further differ in availability and names across different distributions and versions.

Absolutely agree

>You might want to consider inverting the model and embracing declarative configuration management via a tool such as ansible, salt, puppet or a declarative distribution such as nixos. These tools include a broad range of high and low level abstractions to deal with common problems in this domain, and would result in there being a single repository of authoritative configurations. Nearly all of these tools also offer some functionality to validate the "compliance" of a specific server to its declared configuration in order to identify and rectify configuration drift (i.e., "diffs" in the style of Oxidized).

[...]

PR #1820, opened 2019-06-03
Create linuxgeneric.rb model

davama:
See: https://github.com/ytti/oxidized/issues/1818

View [culumus.rb](https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/cumulus.rb) for differences

I have tested it with CentOS 7. It worked fine. :smile: 

I am not too sure how to feel about the `prompt`. 

To to reiterate here, for more specific needs, users could extend the model like so:

```
require 'oxidized/model/linuxgeneric.rb'

class LinuxGeneric
  
  cmd :secret, clear: true do |cfg|
    cfg.gsub! /^(username (\S+) privilege (\d+) (\S+) (\d+)).*/, '\\1 <redacted>'
    cfg
  end
   
  post do
    cfg = add_comment 'THE MONKEY PATCH'
    cfg += cmd 'firewall-cmd --list-all --zone=public'
  end
end

```

Any input is much appreciated!

Thank you,
Dave

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * FEATURE: add sonicos model (@rgnv)
 * FEATURE: add hpmsm model (@timwsuqld)
 * FEATURE: add FastIron model (@ZacharyPuls)
+* FEATURE: add Linuxgeneric model (@davama)
 * BUGFIX: voss model
 * BUGFIX: cambium model should not consider timestamp for backup as unneeded, and causes diffs (@cchance27)
 * BUGFIX: remove 'sh system' from ciscosmb model (@Exordian)
--- a/docs/Model-Notes/LinuxGeneric.md
+++ b/docs/Model-Notes/LinuxGeneric.md
@@ -1,0 +1,23 @@
+# LinuxGeneric model notes
+
+To expand the usage of this model for more specific needs you can create a file in `~/.config/oxidized/model/linuxgeneric.rb`
+
+```
+require 'oxidized/model/linuxgeneric.rb'
+
+class LinuxGeneric
+  
+  cmd :secret, clear: true do |cfg|
+    cfg.gsub! /^(default (\S+).* (expires) ).*/, '\\1 <redacted>' 
+    cfg
+  end
+   
+  post do
+    cfg = add_comment 'THE MONKEY PATCH'
+    cfg += cmd 'firewall-cmd --list-all --zone=public'
+  end
+end
+```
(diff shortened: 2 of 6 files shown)


wk:
This looks interesting; much like the dnsdig model in #1669, it's equal parts interesting and unexpected to see something like this find a use case, but the context seems to be addressed in the related issue.

This PR could benefit from:
* A model notes file, which includes at least the comment above
* CHANGELOG.md entry
* Supported-OS-Types.md entry

davama:
Thank you for the input @wk 

That is an interesting PR; i would say that it is much like this one. although that one looks more complicated and specific; i guess to capture the whole ZONE.

I will do as you noted.

Thank you

davama:
I have done as suggested.

I still plan to add more notes and test with Ubuntu and Debian for completeness
I'll update when done.

Thank you

wk:
This looks like a considerable amount of progress, @davama! It looks basically ready to go in. 

While I maintain there are better-suited tools for this particular task (and recovery from a single text blob will be, at best, inconvenient) - it seems to fill two niches nicely:
* Oxidized testing without real equipment (this model could be used as part of unit tests and ran against a Linux docker container or similar).
* Troubleshooting local installations (by targeting, for example, the Oxidized host itself as the only node with the expectation that it's a good known state host which will succeed).

@ytti, what are your thoughts here?

wk:
@davama - I've discussed this in short with @ytti out of the issue, and there is no objections to this being pulled. If you wish, we can pull it in as-is today. With that said:
* `iptables` is not a package that's installed or available out of the box in all distributions
* `/etc/ntp.conf` relies on a specific version of ntpd being installed, not a common default
* `/etc/snmp/snmpd.conf` relies on snmpd being installed, not a common default

It might be helpful to document the assumptions being made (in terms of distribution, packages present, ...) for the model to behave in a predictable manner.

wk:
Please let us know how you'd like to proceed here.

davama:
Points noted!
I want this to be a basic as possible.
I can simply remove them if need be.
Also, havent gotten around to trying with Unbunt/Debian. 

So ill report when this is ready to go.

Thank you,
Dave

davama:
@wk 
It has been quite busy for me, unfortunately I still have not been able to have time to test with other OSes...

I dont want my slowness to stop the progress of this.
I think you can merge as is. If complaints then we can address then.

Please let me know if there are any questions.

Thank you,
Dave

wk:
Pulling this in as requested; please be aware that the `SNMP settings` block also likely relies on a package that's not present by default in a minimal install of the distribution. A PR for this would be welcome.

PR #1832, opened 2019-06-18
Added "password cipher" to comware secret removal pattern.

bengels00:
## Description
Comware let's you store passwords with the "password cipher" command that was not recognized as secret pattern, so the cipher string was stored in git without removing it.

Added "password cipher" to secret detection patterns.


wk:
Looks good, pulling this in. Thanks!

PR #1840, opened 2019-07-02
Fix location of systemd service file on Debian systems

vincentbernat:
Administrators are not expected to put files in `/lib`. The correct
location for the systemd service file is in
`/etc/systemd/system` (like for RHEL).

This is only a change in documentation. Also, I think we should not use `SIGKILL` as the kill. Is the issue still present in recent versions of oxidized? Just using `SIGTERM` works for me. Cc @mattie47

### DIFF
--- a/extra/oxidized.service
+++ b/extra/oxidized.service
@@ -1,12 +1,9 @@
-#For debian 8 put it in /lib/systemd/system/
-#To set OXIDIZED_HOME instead of the default:
-# ~${oxidized_user}/.config/oxidized in debian 8, then uncomment
-#(and modify as required) the "Environment" variable below so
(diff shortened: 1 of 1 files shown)


mattie47:
@vincentbernat I'm not sure. I used oxidized for around a year on two different machines and the sigkill was always needed. Otherwise I always had to find the pid and kill -9 it. 

I haven't used oxidized now in about a year so I don't know if anything has changed (in a different job).

wk:
Pulling this as-is, since the recommendation to use `/etc/systemd/system` is consistent with distribution guidelines. If SIGKILL vs SIGTERM is relevant and should be fixed, don't hesitate to open a separate PR!

Thank you!

PR #1843, opened 2019-07-09
Change default routing daemon

devicenull:
The default routing stack on modern versions of Cumulus is FRR.  Use this as the default instead of Quagga


wk:
Hi! Changing the default is probably fine, but please update [the documentation](https://github.com/ytti/oxidized/blob/master/docs/Model-Notes/Cumulus.md) for consistency with an explanation on how to set this to Quagga and for which versions instead, i.e. the inverse of what's there currently - and add an attribution line to `CHANGELOG.md` as part of this PR, as this change will likely catch some users of the existing model by surprise.

mortzu:
Please process the checklist

PR #1862, opened 2019-07-30
Nagios

rverchere:
## Description

Enhance nagios script to check if only one host is correctly backuped instead of all.

### DIFF
--- a/extra/nagios_check_failing_nodes.rb
+++ b/extra/nagios_check_failing_nodes.rb
@@ -12,6 +12,11 @@ pending_nodes = []
 
 json = JSON.parse(open("http://localhost:8888/nodes.json").read)
 json.each do |node|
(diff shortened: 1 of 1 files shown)


wk:
Looks good, pulling this in.

acederlund:
Just a quick input on this, if you pass an unknown node it will still say it backed up OK:
[root@XXXXX custom]# ./check_oxidized_backup.rb guasytrgy6uatrgyuiasgtiuawshguis
[OK] Backup of node guasytrgy6uatrgyuiasgtiuawshguis completed successfully.

Perhaps this needs some error handling to check if the node exists at all? That way if running this as a check on switches with the switch name as argument, there will be no error at all if the node is removed from oxidized and this script will still say backup went OK.

Issue #1804, opened 2019-05-09
Fetching configurations not working (running post_command step)


XaaT:
The log is here : https://gist.github.com/XaaT/35f3250e3283c617e8f990cffa2609a4

It seems, after the login, to have something wrong to fetch (or to send command used for) configuration.
More information here : https://community.librenms.org/t/oxidized-only-working-issues-hp-procurve/8245

* Gem install (ubuntu 18.04 and Debian 9.6)
* Simple router.db (1 or 3 switchs)
* HP Procurve 2810-24G, 2510G-48 or FS.com S3900-24T4S
  * 2810-24G firmware : N.11.52
* Using telnet
* Management network (full access from ubuntu 18.04 box)

Confs are in the Librenms community post.

Thanks in advance !

XaaT:
I think the issue is related to telnet.rb as it is not functionnal through at least 2 manufacturers

Issue #1705, opened 2019-02-19
Failed getting to prompt on HP Procurve switches


mincebert:
[Branching from thread 1607.]

Some of my HP ProCurve switches are failing to backup...

> What do the ~/oxidized/logs/ip-ssh and/or ip-telnet file(or whatever your folder structure is) show?

They show the login banner saying "HP J..." ending with "ESC[1;15rESC[1;1H".  (I'm using 'less' to view this so the escape sequences show up rather than being interpreted by my terminal.)

The working one continues on from there saying "ESC[24;1HPress any key to continueESC[15;1H..." and then the rest of the session.  So I think it's failing before it gets to the prompt.

> Have you tried to ssh or telnet from the server to the device does that work correctly?

Yup - it works fine.

> Within oxidized Is debug enabled? If so in the router.db disable all devices by putting a # in front of them except for one of the ones with an issue. Run oxidized manually from command prompt from the user you are running it as, what does the log say?

W, [2019-02-13T19:16:39.489158 #16914]  WARN -- : 172.30.66.15 raised IOError with msg "closed stream"
W, [2019-02-13T19:16:45.733439 #16914]  WARN -- : 172.30.66.16 raised Oxidized::PromptUndetect with msg "unable to detect prompt: (?-mix:^\r?([\w\s.-]+# )$)"

... this message is probably slightly misleading as it's really saying "I didn't get to the prompt" when it actually failed somewhere above that.

It may be worth [as a suggestion for a general improvement] if we could dump out the last 100 or so bytes read on the SSH session to show where it failed, to help work out which bit it's stuck on.

I've had a look at the escape sequences around this and they look the same across the working and not working versions.

I've attached the logs for the two switches and also the output of 'ssh SWITCH | tee -o ...', then taken that and put a newline before the 'ESC[' and a '$' at the end of the lines to highlight those and it looks the same to me.

> Some of the hp switches are comware and not procurve and require a slightly different set of commands/prompts.

Indeed, but I don't have any of those.

> I've tried that 
The key to troubleshooting with oxidized is to watch the logs both ssh/telnet and oxidized itself. Pretty straightforward errors from what I have seen/experienced.

Yes.  I think the HP might be more complicated than some other platforms due to its extensive use of VT100 escape sequences and the need to strip them out with lots of 'expect' substitutions.

[...]

Issue #1607, opened 2018-10-26
Procurve oxidized


NextOne86:
Hi guys, I got this issue: 
 WARN -- : 0.0.0.0 raised Oxidized::PromptUndetect with msg "Your previous successful login (as manager) was on 2018-10-26 16:47:15
 from 0.0.0.0
switchName#  not matching configured prompt (?-mix:^\r?([\w.-]+# )$)"

It seems I get access to the switch but then nothing....
------------------------------------------------------------------------------
My configuration:

username: name
password: pass
model: Procurve
interval: 3600
use_syslog: false
debug: false
threads: 30
timeout: 20
retries: 3
prompt: !ruby/regexp /^([\w.@-]+[#>]\s?)$/
rest: 127.0.0.1:8888
next_adds_job: false
vars: {}
groups: {}
models: {}
pid: "/home/oxidized/.config/oxidized/pid"
input:
  default: ssh
  debug: true
  ssh:
    secure: false
output:
  file:
    directory: "/home/oxidized/deviceconfigs"
source:
  default: csv
  csv:
    file: "/home/oxidized/.config/oxidized/router.db"
    delimiter: !ruby/regexp /:/
    map:
      ip: 0
      model: 1
      username: 2
      password: 3
model_map:
  hp: Procurve
--------------------------------------------------------------------------------
router.db:

#Switches
1.1.1.1:procurve:user:password

NextOne86:
This is what I got from the debug:
W, [2018-10-29T15:28:46.168267 #34901]  WARN -- : 10.4.0.36 raised Oxidized::PromptUndetect with msg "Your previous successful login (as manager) was on 2018-10-29 15:26:03
 from 10.4.96.21
FD_PS/D#  not matching configured prompt (?-mix:^\r?([\w.-]+# )$)"
D, [2018-10-29T15:28:46.168319 #34901] DEBUG -- : lib/oxidized/node.rb: Oxidized::SSH failed for
D, [2018-10-29T15:28:46.168351 #34901] DEBUG -- : lib/oxidized/job.rb: Config fetched for  at 2018-10-29 14:28:46 UTC
W, [2018-10-29T15:28:46.473552 #34901]  WARN -- :  status no_connection, retry attempt 2
D, [2018-10-29T15:28:46.473607 #34901] DEBUG -- : lib/oxidized/worker.rb: Jobs running: 0 of 1 - ended: 0 of 1
[...]

PR #1866, opened 2019-07-31
Procurve compatibility enhancement

deajan:
## Pre-Request Checklist

## Description

Fixes an issue with detecting prompt on older procurve switches (J4899B and J4813A tested), where VT100 control chars get send to scroll, resulting in a prompt which starts with a control character instead of a newline.

Also tested for regression against the following switches (which still work):
- J9776A, J9138A, J9775A, J9783A, J9728A, J9586A

Should fix https://github.com/ytti/oxidized/issues/1804 and https://github.com/ytti/oxidized/issues/1705 and maybe https://github.com/ytti/oxidized/issues/1607

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * FEATURE: add hpmsm model (@timwsuqld)
 * FEATURE: add FastIron model (@ZacharyPuls)
 * FEATURE: add Linuxgeneric model (@davama)
+* BUGFIX: improve procurve telnet support for older switches (@deajan)
 * BUGFIX: voss model
 * BUGFIX: cambium model should not consider timestamp for backup as unneeded, and causes diffs (@cchance27)
 * BUGFIX: remove 'sh system' from ciscosmb model (@Exordian)
--- a/lib/oxidized/model/procurve.rb
+++ b/lib/oxidized/model/procurve.rb
@@ -1,7 +1,6 @@
 class Procurve < Oxidized::Model
   # some models start lines with \r
   # previous command is repeated followed by "\eE", which sometimes ends up on last line
-  
   # additional prompt regex \e\[24;[0-9][hH]([\w\s.-]+# ) catches prompts on telnet with preceding vt100 control chars, where prompt does not start on a new line
   # tested on J4899B HP ProCurve 2650 Switch and J4813A HP ProCurve 2524 Switch
   prompt /^\r?([\w\s.-]+# )$|\e\[24;[0-9][hH]([\w\s.-]+# )/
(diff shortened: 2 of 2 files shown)


deajan:
No idea why the codecov/project check failed :(

wk:
Looks good. Pulling this in.

wk:
Already in, but on a second inspection - why isn't the alternative prompt terminated as last character, i.e. `prompt /^\r?([\w\s.-]+# )$|\e\[24;[0-9][hH]([\w\s.-]+# )$/` (commit misses last `$`)?

wk:
In fact, a more concise way to express this might be `prompt /^(\r|\e\[24;[0-9][hH])?([\w\s.-]+# )$/`.
@deajan Could you validate if this still meets your use case?

deajan:
Let me check this again.

deajan:
@wk Your latest PR https://github.com/ytti/oxidized/pull/1869 broke my fixes.
I didn't do a single regex because on those nasty telnet switches, prompt doesn't begin and end on a single line, hence the separation of my regex to keep "^" and "$" on the regular switches:
```
              [Nice SSH switches][Ugly telnet switches           ]
prompt /^[\r]?([\w\s.-]+# )$|\e\[24;[0-9][hH]([\w\s.-]+# )/
```

Would you mind reconsidering using my form which I was able to validate against all the switches stated above ?

wk:
Normally, the two variants should be equivalent unless there's also *trailing* junk in the prompt, which is quite unusual. Without a pcap or detailed log from interaction with both affected and unaffected devices, it's hard to determine what's going on.

The concern here is that matching such a broad and un-terminated string could easily catch a banner, motd or other output which is not the prompt.

I've reverted the simplification to your original version in #1876 (will be committed when the CI run completes), but it'd be great if this could be narrowed to as specific of a match as practicable.

deajan:
There are indeed trailing VT100 chars. The log files I've got show that quite well.
I have tried to narrow it down a bit more today, trying to find a good workaround for the ending VT100 chars.
Let me see if I can enhance the regex to something less error prone. I'll post back.

deajan:
Okay, I've tried to have something more restrictive.
Provinig that the telnet prompt has 3 trailing VT100 control chars (followed by other variable VT100 chars), I came up with this regex (tested, working):

```
prompt /(^\r|\e\[24;[0-9][hH])?([\w\s.-]+# )($|(\e\[24;[0-9][0-9]?[hH]){3})/
```

@wk Do you think this should be sufficient for the motd / banner not to be detected ?

deajan:
Shall I make a PR for this ? (works with all the switches stated above).

wk:
This indeed looks more conservative and should reduce false positive potential, but the intent becomes more complex to understand; if you do PR this, please include a short comment in the model explaining what's going on for the benefit of future adventurers.

Issue #1703, opened 2019-02-15
Zyxel GS1900 telnet - possible expect issue


inc0gnitus:
I am having an issue backing up a Zyxel GS1900 switch via telnet. The issue is that there is pagination when doing a "show running-config". You can not shut off pagination and there is no other method of grabbing the config that I've found from these units. As far as I can tell, the "expect" block in my model file is being completely ignored. I have tried many different expect prompts, and used a regex checker to verify they all work. Tcpdump doesn't show any whacky characters, just "--More--".

Here is the zyxelgs1900.rb model:

```
class ZyxelGS1900 < Oxidized::Model

  prompt /^.*# $/
  comment  '! '

  expect /--More--/ do |data, re|
    send ' '
    data.sub re, ''
  end

  #cfg << cmd('show run')
  cmd 'show run' do |cfg|
    cfg
  end

  cfg :telnet do
    username /^(User name|.*Username):/
    password /^\r?Password:/
  end

  cfg :telnet do
    pre_logout do
      send "exit\r"
    end
  end
end
```

I wrote a small expect script, separate from Oxidized:

```
#!/usr/bin/expect -f
set user username
set password password
set timeout 30
spawn telnet switchname
expect "Username: "
sleep 1
send "$user\r"
sleep 1
expect "Password:"
sleep 1
send "$password\r"
expect "switchname# "
sleep 1
send "show run\r"
expect {
    "More" {send -- " "; exp_continue}
}
sleep 5
expect "switchname# "
send "exit\rexit\r"
```

The expect script works just fine. The Oxidized model does not. The debug log for a specific device doesn't show anything:

```
Username: =username
Password: *********
switchname# 
```

In /var/log/messages, Oxidized logs a timeout:

`[2019-02-15T11:24:59.883280 #19733]  WARN -- : ###.###.###.### raised Net::ReadTimeout (rescued Timeout::Error) with msg "timed out while waiting for more data"`

A tcpdump shows login and executing "show run", and then output stopping at "--More--". The expect block isn't firing off. 

In my model file, if I change the cmd from "show run" to anything that doesn't paginate (eg, "show run int 1", "show version", etc), it all works just fine. So the issue is that the expect block is not being used during the "show run" command. I can't for the life of me figure out why.

oxidized (0.25.1)
oxidized-script (0.6.0)
oxidized-web (0.12.1)

ruby --version
ruby 2.4.5p335 (2018-10-18 revision 65137) [x86_64-linux]

[...]

Issue #1404, opened 2018-06-17
Model by group


dracmic:
Hello,

Is any possibility to do a model map by group? I have two types of Zyxel Switches and one is not working on ssh only on telnet.

Best regards.

ytti:
Models are mapped at load time, we can't at this time remap them before fetch. But I'm not sure I understand what the problem is. If you set input default to ```ssh, telnet``` it'll try ssh first then telnet, and you should be fine.

dracmic:
Hello ytti,

Actually i hvae GS1900-8 and GS1900-24, the GS1900-24 is almost as CISCO IOS and GS1900-8 is acting like a old CATOS. The GS1900-8 doesn't even login it needs some pre_login stuff.

dracmic:
Yeah ytti, but guess what bot are seen as ZYNOS, that why I asked if it is possible model by group.

ytti:
Well if they are different models, they should be different models. If they are entirely different UI, like you say, IOS and CatOS, they certainly are not same OS and shouldn't be same model.

So from my POV, we need to make new model and you should in your source state that it's different model.

dracmic:
Base model is the same GS1900.

deajan:
@dracmic Would you mind testing the following https://github.com/deajan/oxidized/blob/zynos-gs-compat/lib/oxidized/model/zynos.rb ?
Also, do you use zynos as backend for GS-1900-24 since the OS is totally different than GS-1900-8 ?

PR #1867, opened 2019-07-31
Zynos gs compat

deajan:
## Pre-Request Checklist

## Description

Adds support for Zyxel GS1900-8 switch to ZynOS, might also work with bigger GS1900 switches.

Fixes https://github.com/ytti/oxidized/issues/1703
Fixes part of https://github.com/ytti/oxidized/issues/1404

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master
 
+* FEATURE: add zyxel GS1900 support to ZynOS model (@deajan) 
 * FEATURE: add PurityOS model support (@elliot64)
 * FEATURE: add Ubiquiti Airfiber model support (@cchance27)
 * FEATURE: add Icotera support (@funzoneq)
--- a/lib/oxidized/model/zynos.rb
+++ b/lib/oxidized/model/zynos.rb
@@ -23,7 +23,7 @@ class ZyNOS < Oxidized::Model
     cfg.gsub! /[\b]|\e\[A|\e\[2K/, ''
     cfg
   end
-  
+
   cfg :ftp do
   end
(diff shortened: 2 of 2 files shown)


wk:
Thanks for the PR; breaking functionality for existing users of this model probably not an ideal way to add support for a new variant. 

What about the prompt of SAM1316 and the prompt of GS1900 that is different so that a single prompt regexp can't support both? Could you paste copies of prompts from both types of equipment, or debug logs for both?

deajan:
Sorry I don't own the SAM1316 equipment, I Can only post the logs for mine.
From what I've read, ZynOS is a generic name for multiple different behaving OSes from Zyxel.
Any Idea on how I can make a model for a specific ZynOS without breaking another one ?
Maybe I could just create a zynosgs.rb file instead ans document that ?

wk:
If there are multiple sub-versions (or different OSes available under the same brand), you can indeed create a new model with a different name. 

Renaming the (modified) model from ZynOS to e.g. ZynOSGS (both class name and file name) seems to be the way to go here. You will also likely want to remove the `config-0` and the `ftp` blocks as they won't be relevant.

On a side note, have you attempted the `zynoscli` model against your equipment? It seems to use similar commands, appears to also scrub what seems like control characters, but does not have a 'More' prompt handler (perhaps there's a configuration that eliminates the need for it?).

deajan:
I did test thé zynoscli model, but it seems that thé GS séries are a différent kind of beasts. I'll redo a new PR shorty.

deajan:
Replaced by https://github.com/ytti/oxidized/pull/1875

PR #1868, opened 2019-08-02
Watchguard FirewareOS name fix

deajan:
## Pre-Request Checklist

## Description

LibreNMS returns "fireware" instead of "firewareos".
Checked with snmpwalk over a Watchguard, it's not the firewall itself that gives that name, but librenms.
I'm not sure whether the name should be changed in oxidized or LibreNMS, but but well here we are.
Open to any suggestiongs.

This PR also fixes `undefined method `new' for nil:NilClass` because of the class naming.

No issue open I think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * FEATURE: add hpmsm model (@timwsuqld)
 * FEATURE: add FastIron model (@ZacharyPuls)
 * FEATURE: add Linuxgeneric model (@davama)
+* BUGFIX: fireware model made oxidized crash (@deajan)
 * BUGFIX: voss model
 * BUGFIX: cambium model should not consider timestamp for backup as unneeded, and causes diffs (@cchance27)
 * BUGFIX: remove 'sh system' from ciscosmb model (@Exordian)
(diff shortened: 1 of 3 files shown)


laf:
If the name is FirewareOS then LibreNMS should be changed.

We can automatically adapt the name in the LibreNMS api: https://github.com/librenms/librenms/blob/master/includes/html/api_functions.inc.php#L1211

Add what you need there and submit the pull request.

wk:
Thanks for the PR!

The name of the OS appears to be [Fireware OS](https://www.watchguard.com/wgrd-products
/fireware-os), but WatchGuard seems to use the two interchangeably at times.

If this was go to in it'd have to go in with a shim (similar to `supermicro.rb`) to avoid breaking at upgrade - and would then need to be phased out several versions later.

Inconsistency in product branding, however, seems like an insufficient reason to introduce this change into Oxidized, especially if there's lack of strong and obvious justification to go with one over the other. Unless there are additional arguments besides the mismatch with LibreNMS, there doesn't appears to be a great reason to pull this in.

deajan:
Tested and created LibreNMS side PR. Thanks.

PR #1873, opened 2019-08-05
Remove snmpd lines from linuxgeneric.rb 

davama:
See: 

https://github.com/ytti/oxidized/pull/1820#issuecomment-515677075


wk:
Thank you!

PR #1882, opened 2019-08-14
Fix edgeswitch model

JeanFrancoisMoule:
## Pre-Request Checklist

### DIFF
--- a/lib/oxidized/model/edgeswitch.rb
+++ b/lib/oxidized/model/edgeswitch.rb
@@ -20,11 +20,13 @@ class EdgeSwitch < Oxidized::Model
         send "enable\n"
         cmd vars(:enable)
       else
-        cmd 'enable'
+        cmd 'enable', /^[pP]assword:/
+	      cmd (@node.auth[:password])
       end
       cmd 'terminal length 0'
     end
     pre_logout 'quit'
     pre_logout 'n'
   end
-end
+end  
+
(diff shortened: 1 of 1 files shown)


wk:
It seems that the goal of this change is to facilitate sending the user's own password as an `enable` password if `enable` isn't configured; this results in a rather unexpected behavior:
* If the user did not specify enable, enable is used anyway (but with user's own password)
* If the user did specify enable, the enable command is sent, but no password is sent afterwards.

A more consistent way to handle this could be:

```ruby
    # preferred way to handle additional passwords
    post_login do
      if vars(:enable) == true
        cmd "enable"
      elsif vars(:enable)
        cmd "enable", /^[pP]assword:/
        cmd vars(:enable)
      end
    end
```

The configuration will then need to specify the user's password both as password, and as the value of the enable field. This will result in the same outcome, but a more predictable behavior.

wk:
@JeanFrancoisMoule @fadenb could you perhaps review if this can be done in a manner more similar to the standard way proposed above? We'd like to have a fix in, but for the right problem (not the lack of a correctly set 'enable' value for the node, but for the underlying behavior).

Cormoran96:
With my EdgeSwitch 24 Lite
Os version 1.9.2. It doesn't work. I have message:

```
% Invalid input detected at &apos;^&apos; marker.
```

mortzu:
Duplicate of #2234

PR #1891, opened 2019-08-29
EdgeCOS: include transceiver details

freddy36:
## Pre-Request Checklist

## Description
add support for transceiver details in the EdgeCOS model

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master
 
+* FEATURE: include transceiver information in EdgeCOS model (@freddy36)
 * FEATURE: add ZynOS GS1900 specific model support (@deajan)
 * FEATURE: add PurityOS model support (@elliot64)
 * FEATURE: add Ubiquiti Airfiber model support (@cchance27)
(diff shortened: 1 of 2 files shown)


wk:
Seeing that much of the information produced by the command appears to be scrubbed, it appears to be that the intention is to identify the type of optics in place - does EgdeCOS perhaps offer some form of 'show hardware' or 'show inventory' command that would produce this information as part of its output, but without generating the volatile information often produced by inspecting DOM optics?

freddy36:
@wk there are no such EdgeCOS commands.
There's no other command of getting the transceiver details.

wk:
This looks okay as-is, although for clarity, it might be better to remove the values rather than the entire line (as a user simply viewing the output collected by oxidized may not immediately understand why only some of the output has been collected).

Let me know if you are willing to work on this or if you'd like it to go in in it's current version.

freddy36:
@wk updated the patch to hide the values as suggested

wk:
Great! Thank you. Let's pull this in.

Issue #1893, opened 2019-09-02
Config::Vars more specific value False gets overridden by less specific non-False


raunz:
There's an issue with lib/oxidized/config/vars.rb, when trying to set specific value for Node intentionally to False and there's a less specific variable set to True. Result is True, not False as expected. 
Found this when trying to set remove_secret to False for some specific scenarios, while the model default is true.

```
/etc/oxidized/config
models:
  fortios:
    vars:
      remove_secret: true

node = Oxidized::Node.new(name: "samplenode", ip: "127.0.0.1", group: nil, model: "fortios", output: 'git', vars: { remove_secret: false })

node.vars :remove_secret  returns true.
```
Also if a group variable is intentionally False, it will be overridden by Model variable True value.

ytti:
You are right.
https://github.com/ytti/oxidized/blob/master/lib/oxidized/config/vars.rb#L5

This line would set it to false. Then we continue to set it if it's false or nil.

Probably refactored one would use returns. Something like

```ruby
def vars(name)
   return @node.vars[name] if @node.vars && @node.vars.has_key?(name)
   return Oxidized.config.groups[@node.group].vars[name] if Oxidized.config.groups.has_key?(@node.group) && Oxidized.config.groups[@node.group].vars.has_key?(name)
    return Oxidized.config.models[@node.model.class.name.to_s.downcase].vars[name] if Oxidized.config.models.has_key(@node.model.class.name.to_s.downcase) && Oxidized.config.models[@node.model.class.name.to_s.downcase].vars.has_key?(name)
    return Oxidized.config.vars[name] if Oxidized.config.vars.has_key?(name)
  end
```

Didn't check, even for typos.

raunz:
Resolved with commit 42aa7d17b755e536cc4d948b7a525acf020fe5b5

PR #1894, opened 2019-09-03
Fix issue #1893

raunz:
## Description
Change makes possible to use more specific 'false' values on top of less specific 'true' values.

Closes issue 1893

### DIFF
--- a/lib/oxidized/config/vars.rb
+++ b/lib/oxidized/config/vars.rb
@@ -1,15 +1,14 @@
 module Oxidized::Config::Vars
   # convenience method for accessing node, group or global level user variables
-  # nil values will be ignored
   def vars(name)
-    r = @node.vars[name] unless @node.vars.nil?
-    if Oxidized.config.groups.has_key?(@node.group)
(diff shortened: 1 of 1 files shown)


ytti:
You are double checking the first condition

raunz:
Indeed. Missed that. I'll correct the code, rerun tests and make a new PR later. Apologies.

ytti:
Perhaps we want to ignore nil values? So that false is used as falsy terminating and nil is used to revert to less specific value? I could imagine use-case, when you can't avoid returning value, but you don't want to actually override less specific?

raunz:
The scenario, when the key exists, but the value is nil, can be achieved with a bad configuration. I assume, that we shouldn't specially check for nil? to catch that corner case.
If the return value of `vars` is used by if conditional, then it doesn't make difference, is it nil or false.

ytti:
It does matter, consumer of variable can test if it's false or nil. So against my initial thought, we must terminate on both, so the code is fine.

Issue #1616, opened 2018-11-05
Slow version REST API with git backend


empi89:
We are using oxidized with librenms integration and git output. The load time of the config page in librenms increased drastically over time and we found that it is was due to the oxidized node/version API with a load time of about 15 seconds at about 20k git commits. I also tried to git gc the repo and store the repo in a ram disk, but as the CPU spikes to 100% the problem seems not to be io bound.

I checked the code and found that the implementation at https://github.com/ytti/oxidized/blob/master/lib/oxidized/output/git.rb#L73 (def version node, group), especially the commit.diff slows down the API. I am wondering if it might be possible to add a per device caching there to reduce the queries to the whole git repo. I have already checked if rugged suggests another way of implementing the functionality but it seems that it is already implemented in the suggested fashion. What do you think regarding a cache at this part of the code?

As a workaround we have disabled the versioning feature of the librenms extension to reduce load time again.

zombah:
Check ```git gc``` for you repos they may require housekeeping

PR #1904, opened 2019-09-18
ASA: remove last change time

seros1521:
## Pre-Request Checklist

## Description
The date of the last configuration change is changed even if the user enters the configuration mode and exits. It creates additional (useless) entries in the history, which is inconvenient and slows down the work with a long history (useful for #1616).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 * BUGFIX: updated aosw.rb prompt. addresses issue #1254
 * BUGFIX: update comware model to fix telnet login/password for HPE MSR954 and HPE5130. Issue #1886
 * BUGFIX: filter out IOS configuration/NVRAM modified/changed timestamps to keep output persistent
+* BUGFIX: filter out ASA configuration modified/changed timestamps to keep output persistent
 * BUGFIX: update screenos model to reduce the amount of lines being stripped from beginning of cfg output
 * MISC: add pgsql support, mechanized and net-tftp to Dockerfile
 * MISC: upgrade slop, net-telnet and rugged
--- a/lib/oxidized/model/asa.rb
+++ b/lib/oxidized/model/asa.rb
@@ -34,6 +34,8 @@ class ASA < Oxidized::Model
     # avoid commits due to uptime / ixo-router01 up 2 mins 28 secs / ixo-router01 up 1 days 2 hours
     cfg = cfg.each_line.reject { |line| line.match /(\s+up\s+\d+\s+)|(.*days.*)/ }
     cfg = cfg.join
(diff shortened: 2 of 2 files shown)


raunz:
I'll share here my version of IOS model - I haven't made a PR yet, but the target is the same.
Also vtp status output contains last modified information on some devices.

```
diff --git a/lib/oxidized/model/ios.rb b/lib/oxidized/model/ios.rb
index 1135c6b..3e14531 100644
--- a/lib/oxidized/model/ios.rb
+++ b/lib/oxidized/model/ios.rb
@@ -95,6 +95,7 @@ class IOS < Oxidized::Model

   cmd 'show vtp status' do |cfg|
     cfg.gsub! /^$\n/, ''
+    cfg.gsub! /Configuration last modified by.*\n/, ''
     cfg.gsub! /^/, 'VTP: ' unless cfg.empty?
     comment "#{cfg}\n"
   end
@@ -107,6 +108,8 @@ class IOS < Oxidized::Model
     cfg = cfg.each_line.to_a[3..-1]
     cfg = cfg.reject { |line| line.match /^ntp clock-period / }.join
     cfg.gsub! /^Current configuration : [^\n]*\n/, ''
+    cfg.gsub! /^! (Last|No) configuration change (at|since).*\n/, ''
+    cfg.gsub! /^! NVRAM config last updated.*\n/, ''
     cfg.gsub! /^ tunnel mpls traffic-eng bandwidth[^\n]*\n*(
                   (?: [^\n]*\n*)*
                   tunnel mpls traffic-eng auto-bw)/mx, '\1'
```

wk:
Could you mention shortly what workflow necessitates entering and exiting configuration mode on regular basis without inflicting any changes? 

It seems that under a *normal* use pattern, this should be unnecessary, at it will simply report no changes have taken place, unless the configuration was edited (in which case it will be part of the diff for that edit). What leads to this occurring regularly in your environment?

seros1521:
Investigation of the functionality of devices, checking the conformity of the documentation on the functionality and syntax of commands with real ones. I often just go into configuration mode and look at commands for commands, since the documentation from the vendor is not always accurate and comprehensive, there is not always information about what functionality is available in which version of the software. These checks have to be done not only on test equipment, but also on working equipment (there is no full test copy of the network - it's expensive). As a result, there is no actual change, but the backup version is changing. The result is #1616, although our number of commits is much less than 20k.

RANCID also removes modification time from backups, and in other models this is done. A quick search found code in acos, acsw, ios, ipos, voss models for removing modification time.

mortzu:
Please fix the conflicts

seros1521:
Reopen: https://github.com/ytti/oxidized/pull/2484

PR #1905, opened 2019-09-18
Moved show configuration output the end of cfg

raunz:
## Pre-Request Checklist

## Description
Moved 'show configuration' to the end of commands list, so the cfg result will start with comments and end with configuration. Most of the models execute configuration retrieval command the last one, so changing junos model to the same behaviour would be nice.


wk:
This seems reasonable, but will result in a one-time unexpected diff upon upgrading for all users of this model. It's unclear if this is justified. @ytti - thoughts?

wk:
Pulling this in after short discussion with @ytti on rationale that following standard RANCID behavior here for consistency's sake is better than not - until a meaningful reason to break this convention comes up.

Issue #1754, opened 2019-03-22
(Comware) "Current terminal monitor is off" Flapping issue


berilslv:
Hi Oxidized enthusiasts,

Im facing problem with git version flapping. Exactly the same problem as [#1447 ](https://github.com/ytti/oxidized/issues/1447) .
Problem screenshot: [Here](https://ibb.co/xqtfVGM)

I tried to solve issue by adding vars_map: 

>  #IGNORE_DISPLAY_MY
    if vars :ignore_display
       cfg.gsub! /"! Info: Current terminal monitor is off. .*/, ''
       Oxidized.logger.debug "Ignoring terminal off"
    end  

But without success. 
I hope for some advices.

Have a nice friday guys! ;)

berilslv:
Can somebody help me with this issue? Im still getting this current info flapping.

raunz:
I ran into the same problem with current master comware model and HP 1910/1920 switches.
First issue with the same problem seems to be #1447 
These 19x0 units are tricky and cause very funky stuff with using 'send'
Here a snippet of writeshark telnet tcp stream with current master comware.rb and it explains how 'send' commands get input/output/prompts totally messed up:
```
Username:oxidizeduser
oxidizeduser
Password:oxidizedpass

<hp1910switch>_cmdline-mode on
_y
512900
xtd-cli-mode on
y
512900
screen-length disable
.cmdline-mode on
All commands can be displayed and executed. Continue? [Y/N]y
Please input password:******
Warning: Now you enter an all-command mode for developer's testing, some commands may affect operation by wrong use, please carefully use it with our engineer's direction.
<hp1910switch>xundo terminal monitor
.td-cli-mode on
                     ^
 % Unrecognized command found at '^' position.
<hp1910switch>ydisplay version
.
                     ^
 % Unrecognized command found at '^' position.
<hp1910switch>display device
.512900
                     ^
 % Unrecognized command found at '^' position.
<hp1910switch>sdisplay device manuinfo
.creen-length disableundo terminal monitor
                                   ^
 % Unrecognized command found at '^' position.
<hp1910switch>ddisplay current-configuration
.isplay version
HPE Comware Platform Software
Comware Software, Version 5.20.99, Release 1113
Copyright (c) 2010-2016 Hewlett Packard Enterprise Development LP
HPE 1910-24 Switch uptime is 127 weeks, 5 days, 11 hours, 55 minutes

HPE 1910-24 Switch
128M    bytes DRAM
[...]

Issue #1447, opened 2018-07-12
Comware: Output contains prompt and command (flapping)


empi89:
I have a huge amount (>500) of HPE switches and especially the 5120 which sometimes output the prompt and command. This leads to flapping in configuration versions and flooding of hooks. e.g.:

```
-<SW-INT-XXX >display current-configuration
 #
  version 5.20.99, Release 2221P15
+<SW-INT-XXX>display current-configuration
 #
  version 5.20.99, Release 2221P15
```

I have observed that this is especially the case when the devices have some cpu load due to snmp polling. I have already changed some parts of the comware-Model to have it more reliable:
1. removed "display version" - block (because I had the same behavior as with display current-configuration)
2. removed "display device" - block (because I had the same behavior as with display current-configuration)
3. replaced
```
    post_login 'screen-length disable'
    post_login 'undo terminal monitor'
```
with 
```
    post_login do
      send "screen-length disable\n"
      send "undo terminal monitor\n"
      expect /[cC]urrent terminal/
    end
```
Because I often had the message "Info: Current terminal monitor is off." contained in the output.

For the display current-configuration I would like to implement some kind of late-filter which removes the lines starting with a prompt. However I have not found any examples how to do this in the other models. Could you please provide a hint on this? Thanks in advance!

ytti:
The device obviously always shows prompt and command, right? It doesn't print them sometimes and not print sometimes? 
So they should be taken away by this:
https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/comware.rb#L18

Why isn't it working? I think it's race condition because broken model. We rely on screen-scraping prompt to know when output has ended. If we do not eat prompt when it happens, another command may consume it early and we go out of sync and cause all kind of funky issues.

Your correction makes it more broken. ```send``` doesn't expect anything, if those commands produce prompt, you have prompts in buffer which are unconsumed.

In the original model, there is same problem:
https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/comware.rb#L37-L43

I assume bunch of these actually produce command line, which is not consumed. 

It would be safer to refactor all ```send``` away. Replace them with ```cmd``` which expects to see prompt, or if another argument is specifies, expect to see that regexp. This way we retain lock-step with output. 
I can't tell how to refactor the original, as I don't have the platform, but once it works with all ```send``` replaced with ```cmd``` it'll be far more robust.

berilslv:
[...]

PR #1917, opened 2019-09-30
comware fix 19x0 cli mode and add device manuinfo

raunz:
## Pre-Request Checklist

## Description
BUGFIX: comware improvement for requesting HP 19x0 switches hidden CLI.
FEATURE: comware added device manuinfo to include serial number
Fixes issues #1754 and #1447


wk:
Using 'cmd' instead of 'send' here is definitely a good way forward. Looking at its history, this model has been quite challenging to make compatible with all the variations of equipment and firmware versions deployed in the wild. 

@empi89 @berilslv some reports on if this works out well in your environments would be beneficial.

Pulling this in in hope of additional feedback before it ships as a release...

PR #1930, opened 2019-10-09
Procurve exclude power consumption

deajan:
## Pre-Request Checklist

## Description
I recently worked on HPe 5400 series switches which work great with current oxidized, but create new versions of the configuration because of power consumption output in the configuration.

This fix replaces the actual wattage with `<removed>` so the config doesn't change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * FEATURE: add FastIron model (@ZacharyPuls)
 * FEATURE: add Linuxgeneric model (@davama)
 * FEATURE: add SpeedTouch model (@raunz)
+* BUGFIX: prevent versionning on procurve switches by removing power usage output (@deajan)
 * BUGFIX: improve procurve telnet support for older switches (@deajan)
 * BUGFIX: voss model
 * BUGFIX: cambium model should not consider timestamp for backup as unneeded, and causes diffs (@cchance27)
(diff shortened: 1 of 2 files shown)


wk:
This looks good, but should move from the `:all` section into the specific section that produces the relevant output which is to be filtered.

deajan:
Since we have both a `system-information` and a `system information` section, I thought of putting this into `:all` section in order to avoid duplicating that block for both cases.

wk:
Are the two commands different variants to request the same information which produces the same output from different versions of devices / firmware?

deajan:
To be honest, I don't know which devices use the `system-information` variant.
I only own devices that use `system information`.
It is my wild guess that the power consumption might also exist for the other variant.

davama:
The power one produces the UPSs hardware info, similar to sts info cmd

deajan:
@davama sorry what ? I dont get what you meant.

davama:
i thought i read something else.
My bad

wk:
This should be mostly harmless as-is. Pulling the PR. @Ultra2D  - if you could shed some light as to which of the commands is supported by which device or firmware, it might be used for those wishing to develop this model further.

Ultra2D:
2510 and 2610 use `system-information`. 
The other models I can access do not: 2900, 2910, 3800 (and 5400).

The output is the same.

PR #1950, opened 2019-10-30
remove tacacs, bgp, and ospf secrets

robwilkes:
## Description
Mask TACACS, BGP, and OSPF secrets on NXOS


mortzu:
Please process the checklist

Issue #1959, opened 2019-11-05
Issue with AOS7 specific type


optimuscream:
Hello,

For this device :

Alcatel-Lucent Enterprise OS6900-X72 8.5.196.R04 GA, February 25, 2019 ---> aos7 model

I have to make adjustment to 

`lib/oxidized/model/aos7.rb`

in line 

`username /^login : /   `

to 

`username /^([\w -])*login: /`

Because login prompt is like this, machine name example (OS6900X72-MACHINE)

`OS6900X72-MACHINE login:`

Just for your information. 
And then it works.
This is an issue and also the solution.

Thanks

davama:
You can submit a PR

PR #1961, opened 2019-11-06
Update aos7.rb

optimuscream:
## Pre-Request Checklist

## Description

Login prompt for device type :
Alcatel-Lucent Enterprise OS6900-X72 8.5.196.R04 GA, February 25, 2019

is like this (example machine OS6900-X72-MACHINE)

`OS6900-X72-MACHINE login:`

So I have to edit the login code in aos7.rb to permit oxidized to get through.
Closes issue #1959

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Master
 
+* BUGFIX: better login modalities for telnet in aos7 (@optimuscream)
+
 ## 0.27.0
 
 * FEATURE: add automatic restart on failure for systemd (@deajan)
--- a/lib/oxidized/model/aos7.rb
+++ b/lib/oxidized/model/aos7.rb
@@ -47,7 +47,7 @@ class AOS7 < Oxidized::Model
   end
 
   cfg :telnet do
-    username /^login : /
+    username /^([\w -])*login: /
     password /^Password : /
   end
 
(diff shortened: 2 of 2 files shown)


wk:
This looks good, thanks. It's curious that the original had a space between the password and the colon character. Are there perhaps some versions of the firmware which prompt for a password in this manner? Should the regexp perhaps be updated with optional white-space matching in between?

wk:
I've introduced a regexp that would match both cases out of abundance of caution, feel free to remove the optional white-space match if you determine it is never necessary and the original is a bug.

optimuscream:
Helo,
Are there perhaps some versions of the firmware which prompt for a password in this manner?

Yes.
show system 
System:
  Description:  Alcatel-Lucent Enterprise OS6900-X72 8.4.1.170.R01 GA, April 25, 2017.,
  Object ID:    1.3.6.1.4.1.6486.801.1.1.2.1.10.1.6,
For this specific firmware no space between colon and password. For optional white space I don't know which firmware that present with white space between Password and colon but yes you can update it if you want.
Thanks for the respons.Regards

    On Monday, 20 January 2020, 1:31:10 am GMT+7, Wild Kat <notifications@github.com> wrote:  
 
 
This looks good, thanks. It's curious that the original had a space between the password and the colon character. Are there perhaps some versions of the firmware which prompt for a password in this manner? Should the regexp perhaps be updated with optional white-space matching in between?

PR #1971, opened 2019-11-15
Added 'enable' command for Arista EOS

### LINKED ISSUE #1969: Switch to 'enable' mode for Arista EOS
In my case, the RADIUS AVP's don't put the switch into 'enable' mode by default (as they do on Cisco). 

I've added an 'enable' command to the eos model.

ngen-brett:
## Description
Added 'enable' command for Arista EOS before running privileged commands (show running-config) in case enable mode is not default for collection user.

Fixes #1969

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Master
 
+## 0.27.1
+
+* BUGFIX: enable command for Arista EOS model in case RADIUS AVP's don't default to enable
+
 ## 0.27.0
 
 * FEATURE: add automatic restart on failure for systemd (@deajan)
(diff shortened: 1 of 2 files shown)


wk:
Thanks for sending this in, however - enable support appears to already be implemented in the model, and requires setting the enable knob (for the node) to 'True' so that 'enable' is sent (without a password), optional password support also appears to be present.

This change seems to make 'enable' unconditional and non-configurable, even when not needed, in parallel to the existing functionality.

Is this PR perhaps the result of user error in operating the existing Arista model according to its expectations? If that's not the case, please elaborate on why this change is needed.

PR #2014, opened 2020-01-14
EOS: Hides MACsec Key

krisamundson:
## Pre-Request Checklist

## Description
Hides macsec secret key: the CKN and CAK.

* Original Syntax Example

```
    mac security
       fips restrictions
       !
       profile default
          cipher aes256-gcm-xpn
          key 1ffab311 7 00C655A8E47F0E4756B32767F7C9A34854716C7F33F2E948A39CA9C00D0D0A0330
          mka session rekey-period 3600
          l2-protocol lldp bypass
```

* Replacement Syntax Example

```
    mac security
       fips restrictions
       !
       profile default
          cipher aes256-gcm-xpn
          key <secret hidden>
          mka session rekey-period 3600
          l2-protocol lldp bypass
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Master
 
-* MISC: scrubs macsec key from Arista EOS
+* MISC: scrubs macsec key from Arista EOS (@krisamundson)
 
 ## 0.27.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## Master
 
-## 0.27.1
-
 * MISC: scrubs macsec key from Arista EOS
 
 ## 0.27.0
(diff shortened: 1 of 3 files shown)


ytti:
The change looks fine to me, I'm not too interested in secret hiding, I think it's antiapattern but people expect it because rancid did it, so we have it. So looks OK to me. But we do not want to tie this with release of new version so please just change the change log for master and keep version as is.

We will build docker container nontheless for every commit and those docker container version strings will include the commit hash.

wk:
I've reverted the `version.rb` changes and updated `CHANGELOG.md` to make this ready to pull in. 

@krisamundson is there any reason why the key is matched as `\S` and not as `\h`, seeing how it appears to be HEX? Are there other improvements that can be made to constrain the regexp to make sure it never hits a false positive?

wk:
This looks good to me now, let me know if you are happy with this change set and I'll merge it in!

krisamundson:
+1 from me. Thanks for the quick turnaround and improving my regex-fu.

PR #1855, opened 2019-07-24
Add support for FS switches

davidc:
This model supports Fiberstore S3900-24F4S and S5800-8TF12S

FS.com seem to have a completely different on every switch they sell. The existing "gcombnps" model does not work even slightly with these two switches.

These two switches are even different from each other, one has "show system" and one does not. However this model is good enough to work with both.

I'm not sure how best to name it so I just called it "fs".

PR #2023, opened 2020-01-29
add Centec Networks CNOS support

freddy36:
## Pre-Request Checklist

## Description

Add Fiberstore S5800/S5850 series support

A better solution than #1855

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master
 
+* FEATURE: add Fiberstore S5800/S5850 series support via fsos_s5800 model (@freddy36)
 * FEATURE: include transceiver information in EdgeCOS model (@freddy36)
 * FEATURE: add Telco Systems T-Marc 3306 support via telco model (@SkylerBlumer)
 * FEATURE: add enable support to ciscosmb (@deesel)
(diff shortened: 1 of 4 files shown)


wk:
Since the "FS" branded devices seem to be composed of devices that run very different OSes, it might be helpful to add a comment to the model describing the specific versions of software and hardware this was tested against. See for example `aricentiss.rb` or `audiocodes.rb` models on what that might look like. 

This would allow users to determine if the model is or isn't relevant to them - especially if in the future the vendor complicates this product line by, for example, introducing additional minor variations which once again differ on OS, or instead introduce a single consistent OS across all devices in that family.

wk:
How does this model relate to #1855? Does this supersede it, or are we looking to bring in both?

wk:
It seems support here is for S5xxx, while #1855 aims to support both S3xxx and S5xxx

freddy36:
#1855 tries to support both switches (running completely different software stacks).
For the S3XXX switches the EdgeCOS model can be used (Documentation is alraedy updated accordingly), for the S5XXX this model adds support.
See the discussion in #1855

freddy36:
renamed everything to CNOS for most generic support

wk:
This looks great! Thanks for putting in the time to track the OEM and OS name. Pulling this in.

PR #2061, opened 2020-04-06
fortios: Strip cluster uptime even without :remove_secrets

jplitza:
## Pre-Request Checklist

## Description

Strip cluster uptime from FortiOS dumps even without :remove_secrets
    
This is not a secert but dynamic data that *always* changes between two dumps, thus it should never be included, regardless of the :remove_secrets setting


wk:
Well spotted! Pulling this.

PR #2065, opened 2020-04-09
fix paging output (disable press Q paging message)

vkushnir:
## Description
`/system history print` some times more than screen. And it waits for a key press.
```
U nat rule changed                         user    write
U address list entry added                 user    write
U address list entry added                 user    write
-- [Q quit|down]
```
**without-paging** after **print** solve all problems.


wk:
This looks good! I've added an attribution line to the `CHANGELOG.md` file for this to inform users of the change.

Pulling it in!

PR #2068, opened 2020-04-14
Add basic support for NSX

elmobp:
Add support for VMWare NSX, done it two objects configuration of the whole edge it self and firewall level configuration grouped by edge to make it easier to navigate between diff's.

Minor changes required to http input which are reflected in this PR


wk:
This looks pretty good, but could perhaps benefit from a bit of polish to allow others to orient themselves around the enhancements to the `http` input:

* Documentation on using the new `username`, `password` and `headers` parameters would be welcome (probably by extending `docs/Ruby-API.md, which currently mentions precious little about the HTTP input as a whole).
* Not validating certificates within the HTTP input could be a configuration knob, it's not clear if having the non-validating behavior is a reasonable default.
* A note within the NSX model (or in the model documentation) on the tested version, as well as which endpoint address should be used as node for configuration (Is it the NSX Edge Gateway itself? The NSX Manager? vSphere? vDirector?) may also be prudent.

elmobp:
Hows those changes look?

wk:
This looks excellent! Thanks for extending it!

I've created a small PR for the nsx branch of your repository to include these models in the supported models list, once that (or a similar update with your own wording) is in, I think we are ready to pull it into master.

elmobp:
Done sir :)

wk:
Great! 

I think you may have accidentally overwritten the `CHANGELOG.md` file with the last commit in this change set. The changelog was extensively reformatted between the time you're opened this PR and now, and the version in this PR includes a frankenfile which is a mashup of both old and new.

I've submitted a PR against the nsx branch of your fork to revert this.

elmobp:
Woops! Merged :)

PR #2075, opened 2020-04-27
Updated xos.rb regex-code to better match XOS-prompts

DarkCatapulter:
## Pre-Request Checklist

## Description

This regular expression is invalid on some XOS-prompts, example X620-series, where the OS adds a '\*' symbol in front of the hostname. For example, hostname **mydevice-eu-sw02** would give a prompt looking like this: ```* mydevice-eu-sw02 >```

I manually tested the current regex on xos.rb using [regex101.com](https://regex101.com) and also noticed that it generally does a poor job in matching valid hostnames. In this change, I basically copied the regex that existed on ```fortios.rb```, manually tested it and then slightly tweaked it to work for XOS-prompts as well.

### DIFF
--- a/lib/oxidized/model/xos.rb
+++ b/lib/oxidized/model/xos.rb
@@ -1,7 +1,7 @@
 class XOS < Oxidized::Model
   # Extreme Networks XOS
 
-  prompt /^[*]?\s?([-\w.~]+(\s[(\w\-.)]+)?~?\s?[#>$]\s?)$/
+  prompt /^\*?\s?[-\w.~]+(:\d+)? [#>] $/
   comment  '# '
 
   cmd :all do |cfg|
@@ -43,7 +43,7 @@ class XOS < Oxidized::Model
   cfg :telnet, :ssh do
     post_login do
       data = cmd 'disable clipaging session'
-      match = data.match /^disable clipaging session\n\*?[\w .-]+(:\d+)? # $/m
+      match = data.match /^disable clipaging session\n\*?[\w .-]+(:\d+)? [#>] $/m
       next if match
 
       cmd 'disable clipaging'
--- a/lib/oxidized/model/xos.rb
+++ b/lib/oxidized/model/xos.rb
@@ -1,7 +1,7 @@
 class XOS < Oxidized::Model
   # Extreme Networks XOS
 
-  prompt /^*?[\w .-]+# $/
+  prompt /^[*]?\s?([-\w.~]+(\s[(\w\-.)]+)?~?\s?[#>$]\s?)$/
   comment  '# '
 
   cmd :all do |cfg|
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - relax prompt requirements in ciscosmb (@Atroskelis)
 - fortios model strips uptime even without remove_secrets (@jplitza)
 - HP ProCurve now accepts ">" as apart of the prompt (@magnuslarsen)
-- updated regex-pattern to better match XOS-prompts (@darkcatapulter)
+- fixed issue where the regex-pattern for XOS-prompts used invalid syntax (@darkcatapulter)
 
 ## [0.27.0] - 2019-10-27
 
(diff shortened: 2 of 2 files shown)


DarkCatapulter:
Haven't done anything like "rake test", I have no experience of Ruby. Also haven't done rubocop code analysis since I don't know how to use that either. Though I don't think they are neccessary since I have not changed any code, only a slight tweak on the Regex for prompt. It checks the Codacy/PR Quality Review and Travic CI - PR checks.

wk:
Thanks for this! Besides the addition of the leading asterisk, this seems like quite an extensive change (does XOS, for example, include prompts that end with all of `#`, `$`, and `>`)? 

It's possible this model does not see sufficiently widespread use for other users to have encountered this before and all these are indeed legitimate possibilities - but some of the `fortios.rb` prompt regexp may be too permissive to be applicable here, especially since the lineage of these two operating system types appears completely unrelated.

Could you confirm these changes indeed encompass all applicable cases, or constrain the prompt as much as possible to legitimately configurable / available prompts for this OS type?

DarkCatapulter:
Thanks for the feedback! I am pretty sure the code I added might be a bit to permissive, especially when coming to the prompt ending. I am pretty sure that *$* is not a prompt ending used for XOS. Oxidized is a very new tool for me so I just came up with a quick fix that worked for me, but I also wanted to make sure that I don't add something that messes up for someone else. Thankfully this issue has helped me to get a better understanding of regex, and I can definitely do some research and add slight changes to make the prompt less permissive :)

DarkCatapulter:
I believe that whoever made the regex for ```fortios.rb``` made one that pretty much works with any prompt since I can't ever recall fortinet using ```$``` in its prompt. After playing around with it some more, I am able to use prompts from various vendors, making it pretty extensive. But the code is also very clean, whoever made it seems to have put good thoughts and effort into making it as extensive as possible while trying to reduce its "permissiveness". Kind of hard to explain, easier if you try it out yourself with something like [regex101.com](https://regex101.com).

Of course, we want to limit this code to work specifically with XOS-devices. I will take a closer look into that and update my commit 😄

DarkCatapulter:
I have changed the regex pattern once more to better fit XOS-prompts. This time I took some extra time to analyze the prompt on my devices, as well as searching for pictures and videos if I could find any XOS-devices with differing prompts. Also analyzed fortios-prompts on our company FortiGate, the previous regex I suggested had a lot of unnecessary formating.

After looking a bit more into the previous regex pattern (the one that didn't work) I actually noticed that the main reason it didn't work is because of a syntax error in the regex code (´´´/^*?[\w .-]+# $/´´´). This implies that anyone using that pattern will most likely face with an issue when trying to use Oxidized with an XOS-device. The error occurs because there is no escaping on the ´´´*´´´ in the beginning of the code. For this reason, I have also updated CHANGELOG.md to specify that this is a fix rather than an update on the regex.

I have spent about 4h testing the new regex pattern and fine tuning it so that no issues occur for anyone in the future when using this updated pattern.

lib/oxidized/model/xos.rb
```diff
@@ -1,7 +1,7 @@
 class XOS < Oxidized::Model
   # Extreme Networks XOS
 
-  prompt /^*?[\w .-]+# $/
+  prompt /^\*?\s?[-\w.~]+(:\d+)? [#>] $/
```
DarkCatapulter, line 4: Regular XOS-prompts usually look like this `my-device02.1 # ` where the `.1` is a value that gets appended and incremented in the prompt. The section `(:\d)?` was added because there are older XOS-devices that uses a similar prompt, but the dot is replaced with a colon. For example `my-device02:1 # `

zombah:
@DarkCatapulter if xos device have unsaved config prompt will have star prefix like "* device.X #"

DarkCatapulter:
@zombah the current xos.rb already had the star prefix embedded into the regex pattern. I was just too bad with regex to actually understand it. The real issue however was that the star prefix was not escaped like this `\*`, resulting in syntax error. Anyone using the current xos.rb will most likely face issues when using it in its current state.

zombah:
@DarkCatapulter Thanks for explanation, maybe i just didn't noticed issues then.

wk:
This looks good, pulling it in! Thanks for the additional work in constraining the regexp. Constraining prompts as close as possible to what the device can plausibly emit (thus, the allowed characters for the hostname, plus all decorations for the various possible states communicated via the prompt in each model) ensures the prompt isn't triggered prematurely by decorative MOTDs and login banners.

Issue #2080, opened 2020-05-05
Cumulus model polling kernel routes


plett:
The changes made in PR #537 added some additional features to the Cumulus model, including running `netstat -rn` to collect the kernel routing table. The other changes made in that PR seem solid, but I don't think the current route table is configuration which needs backing up.

I have frequent route changes in my network and this leads to diffs being generated for each of my Cumulus switches on each poll when no configuration has changed.

Would you welcome a PR to remove those lines, or is it functionality that other people find useful?

ytti:
I don't think transient data, like RIB or FIB should be in distribution model. We have mechanism where people can introduce local enhancement and even store that dynamic data to different repo or directory. So I would accept PR to remove any transient data in any model.

PR #2086, opened 2020-05-06
Cumulus model - Don't store transient data

plett:
## Description
This is a trivial change to remove the logging of kernel routes by the Cumulus model, as they are transient data and not configuration which needs backing up.

Closes issue #2080

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - scrubs macsec key from Arista EOS (@krisamundson)
 - rubocop dependency now ~> 0.80.0
 - rugged dependency now ~> 0.28.0
+- cumulus model no longer records transient data (@plett)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


wk:
This seems reasonable, and perhaps the same argument can be made about the output of `iptables -L -n` elsewhere in this model. 

Are there perhaps additional files which should be backed up instead in order to cover static routes and firewall rules, or are those perhaps already present in one of the configuration files being backed up?

plett:
You're right, `iptables` output is also transient data. I've pushed a new patch which fixes that too.

Configured static routes are done through the routing daemon, whose config is already backed up.

Configured ACLs are in a `policy.conf` file according to https://docs.cumulusnetworks.com/cumulus-linux-41/System-Configuration/Netfilter-ACLs/ , I've changed the model to copy that file instead of the `iptables` output

wk:
Looks great! Could you add a line to `CHANGELOG.md` describing the change, and optionally taking credit for it? This will typically be the first place existing users look at when they notice output has changed after an upgrade.

wk:
Thanks! Pulling this in.

PR #2087, opened 2020-05-06
Also store the JunOS machine parsable XML

devicenull:
JunOS configurations can also be represented as XML.  This patch also starts exporting the XML representation of the configuration, which is much more useful if you're doing any sort of parsing of configs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - add Telco Systems T-Marc 3306 support via telco model (@SkylerBlumer)
 - add enable support to ciscosmb (@deesel)
 - add Waystream iBOS model
+- also export JunOS configuration XML
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


ytti:
This would be better kept in your local extensions, you can create local junos.rb which extends the distribution one, so you'll keep picking up any upstream changes too.

It's hard to say
a) everyone wants XML config
b) everyone wants JSON config
c) everyone wants SET config

Personally to me 'set' is most interesting, because it's most greppable.

devicenull:
That's why I implemented as another configuration type.  This doesn't replace the existing `set` style (which I agree, is the most readable format), but just adds an alternate output directory containing the XML.

ytti:
But why XML then? Why not JSON? Why not SET? And as it's 3 lines of code, it seems best people decide locally what they want?

wk:
This seems like it would make a great addition to the [model notes for JunOS](https://github.com/ytti/oxidized/blob/master/docs/Model-Notes/JunOS.md), but would be inconsistent as an integral part of the model. 

Several operating system types offer multiple ways of representing configuration (for example, IOS-XR `formal`), but accommodating all potential output formats would be redundant, and making an opinionated choice about a specific selection thereof based on personal preferences is unlikely to align with the needs of most users.

A PR documenting the XML and SET outputs as alternative outputs in the model notes as a superset of this PR would be more than welcome!

PR #2024, opened 2020-01-29
EdgeCOS: improve model

freddy36:
## Pre-Request Checklist

- [X] Passes rubocop code analysis (try `rubocop --auto-correct`)
- [X] User-visible changes appended to [CHANGELOG.md](/CHANGELOG.md)

## Description
include the commands in the output and update patterns for minor software version dependent differences in the EdgeCOS model

PR #2097, opened 2020-05-15
Fix EdgeCOS model

moisseev:
## Pre-Request Checklist

### DIFF
--- a/lib/oxidized/model/edgecos.rb
+++ b/lib/oxidized/model/edgecos.rb
@@ -8,6 +8,8 @@ class EdgeCOS < Oxidized::Model
   end
 
   cmd :all do |cfg|
+    # Do not show errors for commands that are not supported on some devices
+    cfg.gsub! /^(% Invalid input detected at '\^' marker\.|^\s+\^)$/, ''
     cfg.each_line.to_a[0..-2].join
   end
 
@@ -28,8 +30,6 @@ class EdgeCOS < Oxidized::Model
   end
 
   cmd 'show watchdog' do |cfg|
-    # Do not show error as command is not supported on some devices
-    cfg.gsub! /^(% Invalid input detected at '\^' marker\.|^\s+\^)$/, ''
     comment cfg
   end
 
--- a/lib/oxidized/model/edgecos.rb
+++ b/lib/oxidized/model/edgecos.rb
@@ -48,6 +48,7 @@ class EdgeCOS < Oxidized::Model
 
   cfg :telnet, :ssh do
     post_login 'terminal length 0'
+    post_login 'terminal width 300'
     pre_logout 'exit'
   end
 end
--- a/lib/oxidized/model/edgecos.rb
+++ b/lib/oxidized/model/edgecos.rb
@@ -10,10 +10,13 @@ class EdgeCOS < Oxidized::Model
   cmd :all do |cfg|
     # Do not show errors for commands that are not supported on some devices
     cfg.gsub! /^(% Invalid input detected at '\^' marker\.|^\s+\^)$/, ''
-    cfg.each_line.to_a[0..-2].join
+    cfg.cut_both
   end
 
-  cmd 'show running-config'
+  cmd 'show running-config' do |cfg|
+    # Remove "building running-config, please wait..." message
+    cfg.cut_head
+  end
 
   cmd 'show system' do |cfg|
     cfg.gsub! /^.*\sUp Time\s*:.*\n/i, ''
--- a/lib/oxidized/model/edgecos.rb
+++ b/lib/oxidized/model/edgecos.rb
@@ -11,7 +11,11 @@ class EdgeCOS < Oxidized::Model
     cfg.each_line.to_a[0..-2].join
   end
 
-  cmd 'show running-config'
+  cmd 'show running-config' do |cfg|
+    cfg.gsub! /^(show running-config)$/, '! \\1'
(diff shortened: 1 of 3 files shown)


MajesticFalcon:
Thank you for the changes, I was wondering how to fix the vlan tagging issue in the configuration. The changes work for the model ECS4120-28Fv2-I as well.

moisseev:
Would give more details on that "vlan tagging issue"?

MajesticFalcon:
> Would give more details on that "vlan tagging issue"?

Without "terminal width 300", the configuration will error out when pasting the configuration via console. This is because the default terminal width causes line breaks. This is specifically annoying when using many vlans, 20+, on a port. The fix has been to simply paste into notepad remove line breaks and then paste into console. This commit fixes that issue by increasing the terminal width such that line breaks are not created on ports with lots of vlans.

lib/oxidized/model/edgecos.rb
```diff
@@ -11,7 +11,11 @@ class EdgeCOS < Oxidized::Model
     cfg.each_line.to_a[0..-2].join
   end
 
-  cmd 'show running-config'
+  cmd 'show running-config' do |cfg|
+    cfg.gsub! /^(show running-config)$/, '! \\1'
```
wk, line 15: Is the intention here to remove the command itself from the input? Perhaps cfg.cut_head would be a more obvious approach?

lib/oxidized/model/edgecos.rb
```diff
@@ -24,6 +28,8 @@ class EdgeCOS < Oxidized::Model
   end
 
   cmd 'show watchdog' do |cfg|
+    # Do not show error as command is not supported on some devices
+    cfg.gsub! /^(% Invalid input detected at '\^' marker\.|^\s+\^)$/, ''
```
wk, line 32: Should this perhaps be handled under `cmd :all` to suppress all errors regardless of where they occur? You can find an example in `fastiron.rb`, another model which targets devices where command availability varies.

lib/oxidized/model/edgecos.rb
```diff
@@ -44,6 +50,7 @@ class EdgeCOS < Oxidized::Model
 
   cfg :telnet, :ssh do
     post_login 'terminal length 0'
+    post_login 'terminal width 300'
```
wk, line 53: Is 300 the maximum?

lib/oxidized/model/edgecos.rb
moisseev: No, I am just avoiding accidental copy/pasting of `show running-config` command to a device.
Commands were recently added to the output in #2024. I have no idea why do we want them in the output, but it is ok for me unless they are not commented out.

lib/oxidized/model/edgecos.rb
moisseev: I don't know. I think it is better to selectively suppress errors only for those commands where we expect them.

lib/oxidized/model/edgecos.rb
moisseev: Yes. 300 is the maximum. At least on the devices (software versions) that I have at hand it is impossible to set terminal width to 0, accepted range is 80 - 300.

lib/oxidized/model/edgecos.rb
wk: This is a good point, @freddy36 - could you comment about the utility of including the commands in the output instead of cutting them out? I understand the concern of @moisseev - perhaps we can find a middle ground such as including a comment character before the first line or lines of the output?

lib/oxidized/model/edgecos.rb
wk: Do we know that the watchdog command specifically is omitted in some versions, while the other commands are universal across EdgeCOS versions? I don't have a strong preference here, but dealing with this in `:all` has the added benefit of not finding ourselves in the future in a situation where one contributor running one version of the model adds a command, for another user who runs a different version to discover the command does not exist there. 

There are upsides and downsides to either approach, I'll leave this at your discretion.

moisseev: I've made the change as you suggested.

lib/oxidized/model/edgecos.rb
wk: If we don't receive feedback from @freddy36 here, my preference would be to re-introduce the situation prior to #2024 where the commands were removed from the input, as this has been the implicit default for most models.

moisseev: I agree. Let's wait few more days for their response.

freddy36: We like to have the commands in the output to see which command generated the corresponding output. Commenting out the commands is definitely the best option.
But completely removing them is also fine with me.

wk: In that case, I suggest we remove the specific running-config comment/clean-up here, and revert the `:all` section to remove command names from output (as this is not a common behavior for Oxidized models in general).

We should perhaps consider this a feature request for `cmd` to optionally prefix the name of the command executed to the output using the defined comment character of each model as a global and/or device level setting, since I do see the use case @freddy36 mentions as potentially relevant to a broader audience.

moisseev: Done.

wk:
This looks good, pulling it in. Thanks for seeing this one through, @moisseev, and thank you @freddy36 for the feedback.

PR #2103, opened 2020-05-19
Handle pfSense 2.4.5+ passwords that are wrapped in CDATA

adutton:
## Description
pfSense 2.4.5 now wraps the data inside `<password>` tags with a `<![CDATA[` `]]>` wrapper.  The current regular expression does not match the new format.  This causes pfSense password data to be exposed by Oxidized.  

My regular expression update handles both the existing data (from versions 2.4.4 and prior) and data that is wrapped in CDATA (from version 2.4.5 and beyond).

Example of old configuration data from 2.4.4:
`<password>$1$bcrypthash/</password>`

Example of new configuration data from 2.4.5:
`<password><![CDATA[$1$bcrypthash/]]></password>`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Changed
 
 - rubocop dependency now ~> 0.81.0, the last one with ruby 2.3 support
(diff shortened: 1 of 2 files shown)


wk:
The patch itself looks good. Please include a line in `CHANGELOG.md` as part of the PR describing in short what's new and it's ready to be pulled in.

wk:
Thanks!

PR #2106, opened 2020-05-20
Enterasys 800 series switches

javichumellamo:
## Pre-Request Checklist

## Description
Add support for Enterasys series 800 switches


wk:
This looks good! It would require a couple finishing touches before the PR can be pulled in:
* model file renamed to `lib/oxidized/models/entrasys800.rb`, currently `entrasys800`
* model link added to `docs/Supported-OS-Types.md` so users can discover support
* model inclusion mentioned in `CHANGELOG.md`

Users can also be helped with a note in this PR (as a comment), in a comment in the model file itself, or in model notes describing which version of firmware and model of equipment this model was tested against (output of `show version` or similar).

wk:
I've made minor fix-ups to the PR itself that put everything in its right place and updated the documentation; if you could confirm this looks (and works) correctly, it should be ready to pull in.

PR #2111, opened 2020-05-29
Retrieve unencrypted keys from Instant APs

javichumellamo:
When the script is run against a wireless switch, the command 'encrypt disable' is issued before retrieving the configuration. In IAPs, this command is not implemented but you can run 'show running-config ' with the suffix 'no-encrypt' and the result is similar. Unfortunately if you add this suffix in a wireless switch you will get an error message. So before sending the show run.. command we check wether is a wireless switch or an IAP by executing a 'show switchinfo' and then deciding which is the right command to retrieve the config.
I've also added a blank to the prompt because the prompt detection fails if there is a blank in the hostname

## Pre-Request Checklist


mortzu:
Please process the checklist

Issue #2113, opened 2020-06-01
Extreme OS (xos) don't match prompt when config was changed and don't saved


glaubway:
Hello,
have a problem when config at switch was changed and didn't save. Prompt have "* " before hostname.
```
* sw0.l9.kv.1 # disable clipaging session
                                  ^
%% Invalid input detected at '^' marker.
* sw0.l9.kv.2 # 
```
and that prompt don't match with post_login.
```
    post_login do
      data = cmd 'disable clipaging session'
      match = data.match /^disable clipaging session[\n\r]\*?\s?[-\w]+\s?[-\w.~]+(:\d+)? [#>] $/m
      next if match

      cmd 'disable clipaging'
    end
```
If config saved, oxidized work perfectly.
```
sw0.l9.kv.1 # disable clipaging session
                                ^
%% Invalid input detected at '^' marker.
sw0.l9.kv.2 # disable clipaging
```

DarkCatapulter:
I mentioned this issue in #2107 and have also proposed a fix for it. You can apply it manually if you don't feel like waiting for a pull request.

DarkCatapulter:
I see that you followed my suggestion in #2107. Try using this instead: `/^disable clipaging session\n\r?\*?\s?[-\w]+\s?[-\w.~]+(:\d+)? [#>] $/m`

glaubway:
> I see that you followed my suggestion in #2107. Try using this instead: `/^disable clipaging session\n\r?\*?\s?[-\w]+\s?[-\w.~]+(:\d+)? [#>] $/m`.
> 
> And if that doesn't work, try replacing the `\r?` with `\s?`. Please do tell if it works or not :)

Hi, trying both variant, but no. :(

I saw in logs that oxi stoped at this moment:
```
D, [2020-06-04T09:52:53.046128 #38401] DEBUG -- : lib/oxidized/input/cli.rb: Running post_login command: nil, block: #<Proc:0x0000562564821290@/var/lib/gems/2.5.0/gems/oxidized-0.28.0/lib/oxidized/model/xos.rb:44> at sw0.l9.kv
D, [2020-06-04T09:52:53.046344 #38401] DEBUG -- : lib/oxidized/input/ssh.rb disable clipaging session @ sw0.l9.kv with expect: /^\*?\s?[-\w.~]+(:\d+)? [#>] $/
[...]

Issue #2107, opened 2020-05-25
extreme networks stacking not matching prompt


sinist3rr:
Standalone extreme switch backups works as expected but when I try add stacked switch then I got an error:

`raised Oxidized::PromptUndetect with msg "ExtremeXOS#015#012Copyright (C) 1996-2020 Extreme Networks. All rights reserved.#015#012This product is protected by one or more US patents listed at https://www.extremenetworks.com/company/legal/patents/ along with their foreign counterparts.#015#012==============================================================================#015#012#015#012Press the <tab> or '?' key at any time for completions.#015#012Remember to save your configuration changes.#015#012#015#012#015#012Slot-1 stack_office_x3.1 # not matching configured prompt (?-mix:^\*?\s?[-\w.~]+(:\d+)? [#>] $)"`

The actual prompt looks like **Slot-1 stack_office_x3.1 #**

So, I change a regexp from default
`prompt /^\*?\s?[-\w.~]+(:\d+)? [#>] $/`

to:
`prompt /^\*?\s?[-\w]+ ?[-\w.~]+(:\d+)? [#>] $/`

And it start working.
It would be great if such changes were made to the project code.

DarkCatapulter:
Interesting. I just recently updated that prompt in a pull request to fix an issue I noticed when using standalone devices, though I did not have stacked devices in mind when creating it. When I did change it though, I was unsure if I should change the prompt to:
`/^\*?\s?[-\w.~]+(:\d+)? [#>] $/` (this is what I decided to go with in the end)
OR
`/^\*?\s?[-\w.~ ]+(:\d+)? [#>] $/`

I think that if I had gone with the latter one, you might've not had this issue. Though what confuses me is that the current prompt should've at least caught `stack_office_x3.1 # `, ignoring the `Slot-1` part entirely. I think this might have to do with all of these `#015#012#015#012#015#012` patterns that to me seems to represent new-lines.

If it's okay by you, could you try using this prompt `/^\*?\s?[-\w.~ ]+(:\d+)? [#>] $/` and see if it contains any issues?

Also, if we can verify that stacked switches always have a prompt that begins with `Slot-XX`, then this might be the most optimal prompt to use: ^\*?\s?(Slot-[0-9]+ )+?[-\w.~]+(:\d+)? [#>] $

wk:
[...]

PR #2117, opened 2020-06-04
Update xos.rb prompt to work with stacked switches

DarkCatapulter:
The current xos.rb regexp for prompt does not take stacked switches into account and therefore only works for standalone devices. This is a slight update to make the prompt regexp include stacked switches.
Reference: #2107

This update also fixes an issue that wasn't entirely solved in the previous pull request for xos.rb where Oxidized can't match the prompt if the XOS-device has not saved its configuration.
Reference: #2113

## Pre-Request Checklist


wk:
Pulling this in!
Closes #2107 
Closes #2113

MrHamel:
When will there be a release that includes this? It's been about seven months since this was fixed, and have over 500 switch stacks that can't be backed up without it. Thanks!

PR #2126, opened 2020-06-10
Create PfSenseHttps model

glance-:
## Pre-Request Checklist

## Description
This adds a simple model to use the web ui on PfSense machines to download the configuration. I use this in situations where I use oxidized to backup config from remote PfSense machines where I only have access to the web UI.

This is a RFC. I'm probably abusing a lot of code here, and I should write both documentation and some explaining about the model but I'm throwing this out there for some early feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - model for Zyxel 1308 OLTs (@baldoarturo)
 - model for Linksys SRW switches (@glance-)
 - model for Cambium ePMP radios (@martydingo)
+- model for accessing pfSense over https (@glance-)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


wk:
Oxidized offers an `http` input type which should support https. Is there any benefit to relying on `mechanized` directly instead of that input here?

glance-:
This is way more "scraping" than what the http input can do. This needs Mechanize to handle session cookies, csrf-tokens and such.

ytti:
Here is mechinize model: https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/cambium.rb

glance-:
I didn't see any sane way of using the existing `@main_page` / `:login` `Mechanize` functionally here , where I needed to set `OpenSSL::SSL::VERIFY_NONE` and a custom https-port.
Also, I was a bit afraid of breaking the world if I tried to introduce those flags in `http`.

That was why I did it the way I did.

It's pretty easy to refactor this code if someone could point me in the right direction on how to get the http model working as needed.

mortzu:
Please process the checklist

mortzu:
Please fix the merge conflicts

Issue #2137, opened 2020-06-22
trouble writing new model


systeembeheerder:
I'm typing to backup routers made by LANCOM. I wrote an lancom.rb file:
````ruby
class LANCOM < Oxidized::Model
  # LANCOM Systems GmbH

  prompt />\s?$/

  cmd 'readscript'

  cfg :telnet do
    username  /login:\s/
    password  /^Password:\s/
  end

  cfg :telnet, :ssh do
    pre_logout 'exit'
  end
end
````
However oxidized seems unable to finish fetching the config. Debug output:
````bash
oxidized@nms:~$ cat .config/oxidized/logs/192.0.2.138-ssh


#
| LANCOM 1781EF+
| Ver. 10.32.0176RU9 / 21.04.2020
| SN.  40030099XXXXXX
| Copyright (c) LANCOM Systems

HOSTNAME, Connection No.: 005 (WAN)


root@HOSTNAME:/
> readscriptoxidized@nms:~$
````
https://github.com/ytti/oxidized/blob/master/docs/Troubleshooting.md gives some help but because the debug files lists the ``readscript`` command I assume the promtp regex is ok?

````bash
oxidized@nms:~$ ssh root@192.0.2.138
Local password authentication for root
Password:


#
| LANCOM 1781EF+
| Ver. 10.32.0176RU9 / 21.04.2020
| SN.  40030099XXXXXX
| Copyright (c) LANCOM Systems

HOSTNAME, Connection No.: 002 (WAN)


root@HOSTNAME:/
> this works without problems
 Syntax Error: this works without problems

root@HOSTNAME:/
> readscript
# Script (10.32.0176 / 21.04.2020) (0x0020021c,IDs:2,3,4,9,15,2b;0x00000403)

lang English
flash No

set /Setup/Name "HOSTNAME"

[SNIP]

set /Setup/CWMP/SSL/Hash-Algorithms MD5,SHA1,SHA-256,SHA-384
flash Yes

# done
exit


root@HOSTNAME:/
>
````

Any thought or steps to troubleshoot?

ytti:
Looks like maybe new line missing, we should be adding one. Maybe it wants specifically \r\n

systeembeheerder:
sounds plausible; 
````bash
oxidized@nms:~$ ssh root@192.0.2.138 readscript > readscript.txt
Local password authentication for root
Password:
oxidized@nms:~$ file readscript.txt
[...]

PR #2139, opened 2020-06-22
Add LANCOM support

systeembeheerder:
## Description
Adds support for LANCOM routers

Closes issue #2137

edit: somehow this pull request contains to much commits...


wk:
Hi!

Thanks for that, new models are always welcome. This PR will need an entry in CHANGELOG.md (documenting the change), and an entry in Supported-OS-Types.md (documenting the device support) before it can be pulled.

Finally, what is the reason for the model to be sending `\r\n` explicitly at the end of the command? Does simple invocation of `cmd` not do the right thing? Does it not do the right thing over both SSH and Telnet? This need is surprising as `telnet` sends `\r\n` at the end of a command by default, while `ssh` sends `\n`, so normally only an additional `\r` would be required to accomplish the same result.

wk:
Thanks for documenting this! Pulling it in.

PR #2143, opened 2020-07-05
Glaubway dlink

glaubway:
## Pre-Request Checklist

## Description

Added support for d-link dgs-1100 series switches, tested with dgs-1100-10/me


wk:
Does this device variant include the hostname as part of the login prompt for telnet, such as `switch01 Login:`? It would be helpful for purposes of reviewing this PR if the PR included an example of the prompt(s) that the change is intended to accommodate, as well as the version of the OS on which this occurs.

Change itself looks good and ready to pull if the described above is indeed the intention here.

glaubway:
In my case:
````
DGS-1100-10/ME login: test
Password: 
````
It's default login prompt.

wk:
Thanks! That's indeed what I deducted from the change. Pulling this in.

PR #2163, opened 2020-07-27
Added aoscx.rb model

jmurphy5:
## Pre-Request Checklist

## Description

This adds a model for the Aruba CX switch line. It is modeled after the procurve module, with updated commands to match the Aruba AOS-CX software.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - model for eltex mes-series switches (@glaubway)
 - model for zte c300 and c320 olt (@glaubway)
 - model for LANCOM (@systeembeheerder)
+- model for Aruba CX switches (@jmurphy5)
 
 ### Changed
 
--- a/docs/Supported-OS-Types.md
+++ b/docs/Supported-OS-Types.md
@@ -28,6 +28,7 @@
 * Arris
   * [C4CMTS](/lib/oxidized/model/c4cmts.rb)
 * Aruba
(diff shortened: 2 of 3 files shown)


wk:
Thanks for this! I notice this model is extremely similar to the `procurve` model. Is HP perhaps busy providing a unified NOS CLI experience for both `procurve` devices and these `Aruba-CX` devices? Could you perhaps provide the output of `show version` from such a device to understand what the difference may be?

If these devices and `procurve` devices indeed share a NOS/CLI, it would be preferred to maintain only one model and link the device to the correct model name, instead of introducing two parallel models which are largely similar.

jmurphy5:
They are similar, but there are enough commands that are different that I needed to  . I know Aruba has two software sets, the AOS-CX and AOS-Switch. I believe AOS-Switch is the new version of the procurve software, but I will have to double check.

`-----------------------------------------------------------------------------
ArubaOS-CX
(c) Copyright 2017-2020 Hewlett Packard Enterprise Development LP
-----------------------------------------------------------------------------
Version      : FL.10.05.0001
Build Date   : 2020-07-09 17:48:12 PDT
Build ID     : ArubaOS-CX:FL.10.05.0001:53cb98af4936:202007092351
Build SHA    : 53cb98af4936ce4b2e61fb4bb9dde7e7925c5e29
Active Image : primary

Service OS Version : FL.01.06.0004
BIOS Version       : FL.01.0002`

wk:
If this is a unique variant of ArubaOS, then lets pull this in as a model! Thanks for contributing and the additional detail!

Issue #2151, opened 2020-07-14
Backup Datacom/DmOS - DM4270/4370


kiq-moraes:
Pessoal, alguém conseguiu fazer os backups da série nova da datacom? Os DM3000 e DM4100 fazem sem problemas, mas a série nova (4270/4370) está dando erro...

ffrediani:
I have noticed that although they are very similar the old models (DM3000 and DM4100) after successfully login it gets you directly to the command prompt, while for the newer models (4270/4370) after successfully login there are two "Welcome" messages which Oxidized may not be able interpret correctly and some exceptions need to be made on the model file.

Zuquim:
I managed to solve the issue mentioned by @ffrediani by editing the `prompt` RegEx in the DataCom model.
Mine looks like this now: `prompt /([\n]?[\w-]+[(]*[\w-]*[)]*[#>][\s]?)$/` (the key is the `[\n]?` in the beginning).

Also, since I've been having issues regarding the `expect` method to paginate DmSwitches (I was not able to make it work as expected, ironically), I added a couple of commands to disable DataCom pagination while reading it's outputs.

Will make a PR...

kiq-moraes:
Can you could share and post your datacom.rb file? 
Is it necessary to make any modifications to the global configuration file (config)?

Zuquim:
@kiq-moraes [my modified datacom.rb](https://github.com/Zuquim/oxidized/blob/fix-datacom-2151/lib/oxidized/model/datacom.rb)
Make sure the RegEx I'm using works for your scenario, and keep in mind that the key to correctly match the post login prompt is the `[\n]?` in the beginning of the RegEx.

I did not have to change anything in my global configuration file.

If you have any thoughts about how to better fix this specific issue, please share them. I've never dealt with Ruby before this commit, so I'm a little lost...

giovanitestoni:
Hi all!

I fix this just add on the cmd 'show running-config | no more '

Zuquim:
Hey, @giovanitestoni 

I've just checked and there's no `| no more` option on neither (old OS and DmOS), but there is a `| nomore` on DmOS firmware (nothing like it available for the old firmware). Nice tip, but the main issue here is related to post login prompt detection on DmOS, not pagination.

[...]

PR #2170, opened 2020-08-13
datacom.rb: Fix DataCom prompt and pagination issues (#2151)

Zuquim:
## Pre-Request Checklist

### prompt issue
Since the release of the new DmOS firmware, right after a successful login, the device displayed a welcome message, such as:
>Password:
>Welcome to the DmOS CLI
><username> connected from <ip_address> using ssh on hostname
>hostname# 

For this reason, the default RegEx was not able to match the prompt.
It was fixed by adding a custom prompt for the DataCom model, with `[\n]?` in the beginning.

### pagination issue
It seams that the `expect(regex)` method that was supposed to workaround pagination was not working as it should. Several RegEx combinations were tried without success. Oxidized routine for these devices timed out every time a command that triggered pagination was executed.

In order to address this issue, considering the old and new firmware versions, the device terminal pagination setting is now being temporarily disabled during the backup routine.

----

Closes issue #2151

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - fixed an issue where Oxidized could not pull config from XOS-devices operating in stacked mode (@DarkCatapulter)
 - fixed an issue where Oxidized could not pull config from XOS-devices that have not saved their configuration (@DarkCatapulter)
+- improved scrubbing of show chassis in ironware model (@michaelpsomiadis)
 
 ## [0.28.0 - 2020-05-18]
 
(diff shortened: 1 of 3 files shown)


wk:
Hi there!

It has been the general policy of Oxidized to be able to successfully operate in "read only" mode, i.e., as an unprivileged user with no rights to make any modifications (temporary or permanent) to the configuration.

While I see the utility of the approach taken here, I suspect this breaks this promise as a change is introduced (even if promptly reverted). Are there no alternatives available for this model to overcome pagination, such as emulating key presses and stripping the prompts away, or utilizing some runtime / session command to modify this behavior?

Zuquim:
Hey, @wk !

Thanks for the feedback and explanation.

On the following weeks I'll be running some tests on a couple of Datacom device models. When I find a solution that meets the Oxidized policy, I'll commit the updated fix to this branch.

diegocanton:
Thanks...
It's Work in DM4370

mortzu:
Can you fix the conflict

PR #2190, opened 2020-09-26
migrate mail option to being part of the script to stop endless blank failure emails

nickhilliard:
## Pre-Request Checklist
[tests don't apply to extra/ directory #yolo]

## Description
As it stands, `oxidized-report-git-commits` will send an failure email if a device is offline.  This is an unmerciful nuisance which ends up with lots of spam.  Oxidized is not a monitoring system, so this is of limited use at best.

This update moves sending the email to the recipient into the script, and introduces the `-f` option which suppresses these failure emails.

### DIFF
--- a/extra/oxidized-report-git-commits
+++ b/extra/oxidized-report-git-commits
@@ -21,41 +21,98 @@
 #            Removed requirement for Git working directory
 #            Check job status
 #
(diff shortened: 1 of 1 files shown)


wk:
Thanks for cleaning this up! Pulling it in.

PR #2193, opened 2020-10-01
Add git-email in dockerfile to allow git send-email from docker image

jmurphy5:
## Pre-Request Checklist

## Description

To enable git send-email while using the docker image, added git-email and libmailtools-perl to the Dockerfile. When the container is built it allows for sending git diffs via git send-email.


wk:
Thanks! Pulling this in since it's a small dependency in an (unfortunately) large container; please consider documenting this use case for those who prefer to use this method over the mail sending scripts in `extra`.

PR #2194, opened 2020-10-02
add support for Dell N-Series switches

wasapi:
## Pre-Request Checklist

## Description
Add file for N-Series switches.
Based on the DellX, with more section replaced by 'terminal length 0' on post_login ("More:" not match on DellN), and add possibility to enable without password.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - model for zte c300 and c320 olt (@glaubway)
 - model for LANCOM (@systeembeheerder)
 - model for Aruba CX switches (@jmurphy5)
+- model for Dell EMC Networking N-Series switches
 
 ### Changed
 
--- a/docs/Supported-OS-Types.md
+++ b/docs/Supported-OS-Types.md
@@ -91,6 +91,7 @@
   * [PowerConnect](/lib/oxidized/model/powerconnect.rb)
   * [AOSW](/lib/oxidized/model/aosw.rb)
   * [DellX](/lib/oxidized/model/dellx.rb)
+  * [DellN](/lib/oxidized/model/delln.rb)
   * [Dell EMC Networking OS10](/lib/oxidized/model/os10.rb)
 * D-Link
   * [D-Link](/lib/oxidized/model/dlink.rb)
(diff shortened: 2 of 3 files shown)


wk:
Hi! Thanks for this PR!

This looks good, but seems to borrow much of the construction of the Dell-X model. Is there any fundamental incompatibility that would prevent extending the Dell-X model to also support the Dell-N?

wasapi:
The main incompatibility is the match of "More:" in the dellx model; for N-Series it must match "--More--". Also, I find it more elegant to add a 'post_login' to disable page view, but I don't know if it works with X-Series.

ntokos:
Hello,
I have a Dell N-2048 switch (running image version 6.6.0.10) and I'm using the powerconnect model, it works pretty good.
The only problem is it doesn't retrieve the output of the 'show system' command which is fine for me and also the code of the dellx model I think whould have he same problem.

wasapi:
I didn't take a look at the powerconnect model because the powerconnect 2848 switchs I replaced had almost nothing in the cli, except a `show tech-support` command. Indeed it seem it can work with N-Series.
But N-Series commands are the same as the X-Series, except the _more_ text, it could be nice if someone could test the `post_login 'terminal length 0'` instead of the `expect` block, I think it should work.
Maybe it should be better if I directly made a pull request for the dellx model.

cemtiger:
Sadly I'm still getting the follow return from a PowerConnect switch using this, another modified file I found, and the original file. It is interesting that it runs the first command no issue, then the second command runs but it only returns the last 4ish lines of a much larger output in the cli.

! 
! Machine Description............... Dell Networking Switch
! System Model ID................... N1548P
! Machine Type...................... Dell Networking N1548P
! Serial Number..................... ****************
! Manufacturer...................... 0xbc00
! Burned In MAC Address............. *****************
! System Object ID.................. *********************
! SOC Version....................... BCM56150_A0
! HW Version........................ 3
! CPLD Version...................... 16
! 
! unit active      backup      current-active next-active
! ---- ----------- ----------- -------------- --------------
! 1    6.3.3.14    6.6.3.3     6.3.3.14       6.3.3.14
! 2    6.3.3.14    6.6.3.3     6.3.3.14       6.3.3.14
! 
! System Description: Dell Networking Switch
! USB Port Power Status:
! ----------------------
! Device Not Present
                 ^
% Invalid input detected at '^' marker.

mortzu:
Please fix the merge conflict

PR #2206, opened 2020-10-27
TP-Link: working telnet, include enable-admin and other fixes

### LINKED ISSUE #1906: Issue with TP-Link T1700G-28TQ
Hello,

I'm currently experiencing some errors with my TP-Link switch.
Oxidized manages to connect to it, but then the second command (show running-config) fails with invalid parameters, which is surprising since it works well when I do the exact same thing by myself.

Here's the debug log:

```
sw-distri-1>enable

sw-distri-1#mypass
------------^
Error: Bad command

sw-distri-1#

sw-distri-1#show system-info
 Master Unit ID       - 1
 System Description   - JetStream 24-Port Gigabit Stackable Smart Switch with 4 10GE SFP+ Slots
 Device Name          - sw-distri-1
 Device Location      - XXXXX
 Contact Information  - XXXXX
 MAC                  - XXXX
 System Time          - 2019-09-18 11:34:53

Device Info
Unit 1
 Unit State                 - Ready
 Hardware Version           - T1700G-28TQ 3.0
 Firmware Version           - 3.0.0 Build 20190108 Rel.62365(s)
 Bootloader Version         - TP-LINK  BOOTUTIL(v1.0.0)
 Serial Number              -
 Running Time               - 76 day - 10 hour - 20 min - 38 sec
 Power Supply Module        - Operational
 Redundant Power Supply     - Not Support


sw-distri-1#show running-config
-----------------^
Error: Invalid parameter.

sw-distri-1#

sw-distri-1#

sw-distri-1#exit

sw-distri-1>logout
```

(It's also trying to input password in a new line, which is not required since in my case enable does not require any password. But this doesn't seem to be the issue there.)

Could you please take a look ? I am willing to run tests you'd like me to.

### LINKED ISSUE #2205: [BUG]: Telnet doesn't works in TP-Link switches
Connecting to a TP-Link switch using Telnet hangs in a `Login invalid` loop as seen in the debug input below (don't mind the wrong date):
```
***************** User Access Login ********************

User:

 Login invalid.
#2006-01-19 03:11:17,[User]/4/Login the CLI failed by vty0 (192.168.0.42).

***************** User Access Login ********************

User:

 Login invalid.
#2006-01-19 03:11:17,[User]/4/Login the CLI failed by vty0 (192.168.0.42).

***************** User Access Login ********************

User:

 Login invalid.
```
The login regular expression expects `User Name:` but the string sent is only `User:`

piterpunk:
## Pre-Request Checklist

## Description
Many fixes mainly to allow Oxidized to use a normal user to do the backup and the connection through Telnet:
#### Telnet mode
- Change the username regular expression to accept both `User Name:` and `User:` as valid prompts.
- Prevent an additional '\r' to be sent in login when using `telnet` as input. The old behaviour always send the user name as an empty string.
#### Allow non privileged user to be used
- Allow `enable` to be used with and without enable password. Previously it always send the contents of `enable` variable, even without the `Password:` prompt. To allow `enable` without password, set the `enable` variable to `true`. If you want to use a password, set `enable` variable to the password itself.
- Added 'enable-admin' to be sent. This allows a normal user to execute administrative commands. 
#### Miscelanea
- Added the command to disable paging: `terminal length 0`.
- Changed the regular expression to determine if a carriage return needs or not to be sent. Previously it always send a `\r` in the `Password:` prompt before send the defined password.

Closes #2205 
Closes #1906

### DIFF
--- a/lib/oxidized/model/tplink.rb
+++ b/lib/oxidized/model/tplink.rb
@@ -7,13 +7,14 @@ class TPLink < Oxidized::Model
     @input.class.to_s.match(/SSH/)
   end
 
-  # With disable paging this is not needed
-  # # handle paging
-  # # workaround for sometimes missing whitespaces with "\s?"
-  # expect /Press\s?any\s?key\s?to\s?continue\s?\(Q\s?to\s?quit\)/ do |data, re|
-  #   send ' '
-  #   data.sub re, ''
-  # end
+  # handle paging
+  # workaround for sometimes missing whitespaces with "\s?"
+  # With the disable paging feature this is not needed, but let's keep it if
+  # some older firmware doesn't accept the 'terminal length 0'
+  expect /Press\s?any\s?key\s?to\s?continue\s?\(Q\s?to\s?quit\)/ do |data, re|
+    send ' '
+    data.sub re, ''
+  end
 
   # send carriage return because \n with the command is not enough
   # checks if line ends with prompt >,#,: or \r,\n otherwise send \r
@@ -23,8 +24,8 @@ class TPLink < Oxidized::Model
   end
 
   cmd :all do |cfg|
-    # # remove unwanted paging line
-    # cfg.gsub! /^Press any key to contin.*/, ''
+    # remove unwanted paging line
+    cfg.gsub! /^Press any key to contin.*/, ''
     # normalize linefeeds
     cfg.gsub! /(\r|\r\n|\n\r)/, "\n"
     # remove empty lines
--- a/lib/oxidized/model/tplink.rb
+++ b/lib/oxidized/model/tplink.rb
@@ -3,23 +3,28 @@ class TPLink < Oxidized::Model
   prompt /^\r?([\w.@()-]+[#>]\s?)$/
   comment '! '
 
-  # handle paging
-  # workaround for sometimes missing whitespaces with "\s?"
-  expect /Press\s?any\s?key\s?to\s?continue\s?\(Q\s?to\s?quit\)/ do |data, re|
-    send ' '
-    data.sub re, ''
+  def ssh
+    @input.class.to_s.match(/SSH/)
   end
 
+  # With disable paging this is not needed
+  # # handle paging
+  # # workaround for sometimes missing whitespaces with "\s?"
+  # expect /Press\s?any\s?key\s?to\s?continue\s?\(Q\s?to\s?quit\)/ do |data, re|
+  #   send ' '
+  #   data.sub re, ''
+  # end
+
   # send carriage return because \n with the command is not enough
-  # checks if line ends with prompt >,# or \r,\nm otherwise send \r
-  expect /[^>#\r\n]$/ do |data, re|
-    send "\r"
+  # checks if line ends with prompt >,#,: or \r,\n otherwise send \r
+  expect /[^>#\r\n:]$/ do |data, re|
+    send "\r" if ssh
     data.sub re, ''
   end
 
   cmd :all do |cfg|
-    # remove unwanted paging line
-    cfg.gsub! /^Press any key to contin.*/, ''
+    # # remove unwanted paging line
+    # cfg.gsub! /^Press any key to contin.*/, ''
     # normalize linefeeds
     cfg.gsub! /(\r|\r\n|\n\r)/, "\n"
     # remove empty lines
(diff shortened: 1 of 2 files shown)


lib/oxidized/model/tplink.rb
```diff
@@ -34,27 +39,39 @@ class TPLink < Oxidized::Model
...
+        cmd "enable", /^[pP]assword:/
+        send vars(:enable) + "\n\r"
       end
+      # disable paging
+      cmd "terminal length 0"
+      # enable-admin gives admin privileges for regular users
+      # Sending enable-admin with an admin user returns a message warning that
+      # the user already is admin without further consequences. So, always
+      # send the enable-admin.
+      # There is an option to set a password for enable-admin, but testing
```
FlorianHeigl, line 72: Hi, setting the enable password DOES work. It's related to AAA configuration, there's one method list for login password and one for enable password, it needs to be set to 'local' instead of nothing.
(Testing on a `T1700G-28TQRev2` - still 2.0.1, gonna check if there's an update I need to add)

lib/oxidized/model/tplink.rb
```diff
@@ -34,27 +39,39 @@ class TPLink < Oxidized::Model
...
-        cmd vars(:enable)
+    post_login do
+      if vars(:enable) == true
+        cmd "enable"
+      elsif vars(:enable)
+        cmd "enable", /^[pP]assword:/
+        send vars(:enable) + "\n\r"
       end
+      # disable paging
+      cmd "terminal length 0"
```
FlorianHeigl, line 67: T1700G-28TQRev2 v 2.0.1 (supported, but has no newer fw) doesn't have the command, and it has paging enabled. i'd recommend not to remove the old handling, yet).

piterpunk: Oh, I thought that it maybe should be useful on older firmwares and already reverted the changes in 5470318eac27d94635fd40fe3a0be60174eed204

lib/oxidized/model/tplink.rb
piterpunk: I'll try again to configure and test the password when calling enable-admin

lib/oxidized/model/tplink.rb
MarcHagen: Maybe handy to know, T1700G-28TQRev2 can use the V3 firmware. I have it running for over a year. 
Support told me to upgrade because of some issues i had which where fixed in the V3 firmware.
But also `terminal length 0` does not exist in the v3 firmware, sadly

lib/oxidized/model/tplink.rb
```diff
@@ -34,27 +40,39 @@ class TPLink < Oxidized::Model
...
-  cmd 'show system-info' do |cfg|
+  cmd "show system-info" do |cfg|
     comment cfg.each_line.to_a[3..-3].join
   end
 
-  cmd 'show running-config' do |cfg|
+  cmd "show running-config" do |cfg|
     lines = cfg.each_line.to_a[1..-1]
     # cut config after "end"
     lines[0..lines.index("end\n")].join
```
MarcHagen, line 50: For my T1700G-28TQRev2 with V3 there is no end it just quits. Dunno how relevant it is tho :)  

```
 ip igmp snooping vlan-config 3004 ltime 1
 ip igmp snooping vlan-config 3004 querier
 ip igmp snooping  vlan-config 3004 querier general-query source-ip 172.16.4.1
#

sw2#

sw2#

sw2#
```

lib/oxidized/model/tplink.rb
jplitza: I'm confused. So TP-Link has Exec, Privileged Exec and Admin Exec, and you switch from Exec to Privileged Exec with "enable" and from Privileged Exec to Admin Exec with "enable-admin", and both commands can have (different?) passwords?

lib/oxidized/model/tplink.rb
piterpunk: I guess it isn't a problem. Problem is to have the `end` and save the garbage after it in the backup.

lib/oxidized/model/tplink.rb
piterpunk: Well, now both modes are supported, the code to handle paging is still there and, if supported, the `terminal length 0` will works.

lib/oxidized/model/tplink.rb
piterpunk: When you log in as a normal user it logs in Exec mode, then you need `enable` to enter `Enter Privileged EXEC Mode` and then use `enable-admin` to `Achieve the admin privilege`.

`enable` and `enable-admin` can have equal, different or empty passwords. Factory configuration defaults to both empty, if I remember correctly.

piterpunk: And @FlorianHeigl was correct, I just had a AAA config incomplete. Now I can have both `enable` and `enable-admin` passwords set.

And I don't know where to store the `enable-admin` password, there is another device with three different passwords? Or the better is to not support this configuration?

piterpunk:
This last change only resolve merge conflicts. Still only handles `enable-admin` with no password, I still think it's good to merge this code and next think about the support of `enable-admin` with password.

lib/oxidized/model/tplink.rb
mortzu: It will result in "Error: Bad command" on older devices. Maybe we should make it configurable?

lib/oxidized/model/tplink.rb
mortzu: Older devices does not have "terminal length 0". Maybe we should add a new model? Or make it configurable to use newer commands?

piterpunk: I don't understand the problem with the code as-is. It will send the `terminal length 0`, if it works, paging will be disabled. If not works Oxidized will get a `Error: Bad command` and the backup process will continue. If the `Press any key to continue` prompt appears, Oxidized will send `<space>` and the output will scroll until the end.

This will work with devices that have the `terminal length 0` and with those that doesn't have it.

If this is a problem it's easier to remove the `terminal length 0` command. It isn't an impeditive to the backup process.

lib/oxidized/model/tplink.rb
piterpunk: But, even with this `Error: Bad command`, the backup will be done on those older devices that doesn't have the `enable-admin`. Just tested sending some made up `Bad command`s and the backup ends successfully.

I guess it deserves a comment in the code stating that behaviour, so people won't be puzzled by this message.

As a new information, while testing sending invalid commands, I found that using the T1500G-10PS's newest firmware, the `enable-admin` is not needed to `show running-config` and the other information gathering commands. But I really don't know if this behaviour change is a bug or a feature. Regardless this, the `enable-admin` is still needed by some other devices and firmware versions.

mortzu:
Can you fix the merge conflict (again...)? Sorry!

GeComDohlad:
guys hi. Still not working for me. And still have output:
--------^
Error: Bad command

We have TP-Link devices TL-SG3428, TL-SG3210, TL-SG3210 and TP-Link JetStream T2500G-10TS 1.0.
Some devices are without telnet but SSH with oxidized should work

robertcheramy:
Open a new issue and don't comment a closed pull request. You can reference to the PR using #2206 in the Issue.

PR #2209, opened 2020-11-02
Collect config from HRP enabled devices

mp4lm:
## Pre-Request Checklist

- ~~[ ] Tests added or adapted (try `rake test`)~~ N/A

## Description

The change makes it so that devices that are in HRP mode can correctly be collected and parsed by oxidized. If not the * is removed is will not process devices in HRP mode. Even with this change other Huawei devices can be correctly collected as before.


mortzu:
Please process the checklist

PR #2211, opened 2020-11-05
Fix node path in git output.

sargon:
## Pre-Request Checklist

## Description
When single-repo is activated and a node has no group defined the prefix
should not be applied. Otherwise the api is not able to retrieve old
versions.

This may be a condition raised by the use of the web interface, when
the not given group is not nil? but empty string.


mortzu:
Please fix the merge conflict

PR #2216, opened 2020-11-19
Bugfixes for config loading, signal handling and worker deadlock

gs-kamnas:
## Pre-Request Checklist

## Description
This change aims to address a series of bugs encountered in our usage of Oxidized, namely:
- Oxidized would not exit on SIGTERM nor reload its node list on SIGHUP if the rest API was enabled.
- Oxidized's worker would enter an aggressive loop if the /node/next API was called such that the worker pool became larger than the node list.
- OXIDIZED_HOME environment variable was not affecting the location that the main user configuration was loaded from.


mortzu:
Can you split this in three PRs please?

PR #2220, opened 2020-11-23
hook/exec: preserve some environment variables

vincentbernat:
## Pre-Request Checklist

## Description
When executing a hook, the exec plugin removes all environment
variables. However, some variables can be useful to have: many software
crash if HOME is unset (including oxidized when we want to use `oxs`)
and PATH is useful to have for software not trying to populate a PATH
before invoking `exec*()` (unlike a shell).

### DIFF
--- a/lib/oxidized/hook/exec.rb
+++ b/lib/oxidized/hook/exec.rb
@@ -59,7 +59,12 @@ class Exec < Oxidized::Hook
 
   def make_env(ctx)
     env = {
-      "OX_EVENT" => ctx.event.to_s
+      'OX_EVENT'                => ctx.event.to_s,
+      'PATH'                    => ENV['PATH'],
+      'HOME'                    => ENV['HOME'],
+      'OXIDIZED_HOME'           => ENV['OXIDIZED_HOME'],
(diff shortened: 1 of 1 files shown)


seros1521:
You just added some variables to the list, but if you need to pass others in, you'll have to edit the source again. Better to do it differently, add parameters:
1. Whether to copy the environment variables of the main process
2. regular expression (or a list of them) allowing copying of environment variables with a name that matches it (the default value of your choice is HOME, PATH, etc).
3. regular expression prohibiting copying of environment variables with a name that matches it (the default value is OXIDIZED_SSH_PASSPHRASE, maybe others containing potentially sensitive data).

or just do it here:
https://github.com/ytti/oxidized/blob/63fbbbe6eff2ae43935e1e70821d8bb1979a7eb7/lib/oxidized/hook/exec.rb#L45
parameter `unsetenv_others` configurable.

A workaround: you can simply write all the necessary variables into a separate file, and run your main script through a wrapper:
```
#!/bin/sh
source <path_to_ENV_file>
exec <real_hook_script>
```

mortzu:
Please process the checklist

Issue #639, opened 2016-12-13
Possible to use another IOS device as VRF proxy?


ecm75:
I have the following challange while using oxidized:

I have a lot of IOS router which are not reachable over public internet, however I can reach them through a central IOS souter and reach the private router using the following command on the router CLI:
telnet A.B.C.D vrf VRF_TAG

Would it be possible to implement a model which does this:
- connect to router A, 
- login to router A, 
(- do enable on router A, )
- do telnet ... vrf ... to router B, 
- login to router B,
(do enable on router B,)
- save config of router B, 
- disconnect from router B,
- disconnect from router A

I'd like to achive this by setting additional vars for router B, like eg.
vrf_tag, vrf_host, vrf_user, vrf_pass, vrf_enable

My ruby knowledge is quite limited, so I don't even know where to start...

ytti:
We support SSH proxy right now, I've never tested if IOS router can act as SSH proxy, I don't have high hopes, but worth a shot.

If not, then easiest would be to install server to act as your SSH proxy between the networks. 

My suspicion is that this is highly atypical scenario, and while it wouldn't be hard to add support for this, it would be decent amount of work and refactoring. The complexity would be justifiable if were common requirement, but right now I don't think it is.

ecm75:
I will try with ssh_proxy option, thanks for the fast answer!

ecm75:
Is is possiböle to use ssh_proxy with telnet to the router?
oxidized-host --SSH--> ssh_proxy-host --TELNET--> router

ytti:
Not right now, you could try to change this: https://github.com/ytti/oxidized/blob/master/lib/oxidized/input/ssh.rb#L28 to telnet and tell if it works. 

If it does, I can make it configurable.

ecm75:
No, it doesn't seem to work, but i I understand the code correct it executes something like this on the oxidized host:
`ssh <ssh_proxy> -W <node_host>:<node_port>`
However I can connect to the router using this command on the shell from the oxidized host:
`ssh <ssh_proxy> -W <node_host>:23`
I then get the router login prompt, but even if i change the code in ssh.rb to 
`Net::SSH::Proxy::Command.new("ssh #{proxy_host} -W %h:23)`
it doen't work from within oxidized...

ytti:
I guess you're out of luck then and have to use SSH :/

ecm75:
[...]

PR #2230, opened 2020-12-07
input/ssh.rb: Proxy over cisco to cisco

SjonsDePlons:
## Pre-Request Checklist

## Description

As already noted in issue #639, there isn't a functionality that let's you use a proxy over another IOS device to an IOS device. So I have added proxy functionalities which also let's you specify the protocol (ssh or telnet). In the issue #639 he talks about proxy over vrf. This functionality was also added.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -142,6 +142,32 @@ vars_map:
 ...
 ```
 
+## Proxy over Cisco Device to Cisco Device
+
+When using the ssh proxy you will run into some problems if you use a cisco device as a jumphost/proxy. 
+
+This can be provided by defining the following variables in oxidized/config; 'cisco_jumphost', 'cisco_proxy_user' and 'cisco_proxy_pass'.
+
+optionally you can add 'vrf_name' if your proxy is via VRF, but this variable doesn't need to be defined.
+
+Below is an example with the VRF variable added with it.
+
+```yaml
+...
+map:
+  name: 0
+  model: 1
+  ip: 2
+  username: 3
+  password: 4
+vars_map:
+  enable: 5
+  cisco_jumphost: 6
+  vrf_name: 7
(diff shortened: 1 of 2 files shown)


seros1521:
1. duplication of code. The vrf_ssh_shell_open and proxy_ssh_shell_open functions are the same, they differ only in the command they execute on the device. Ditto for the vrf_telnet_shell_open and proxy_telnet_shell_open functions. telnet differs little from ssh too.
2. it will not work if exec command execution mode is used. In some models this mode is fixed explicitly and initially (for example, junos). You can disable it, but it is not necessary in your documentation. In addition, some model classes are simply not adapted to work in a mode other than exec. Of course, you won't be helping users to deal with this and fix bugs for devices that you don't have.
3. using your code will result in the model class being unable to determine which type of input it uses - ssh or telnet over an ssh connection to cisco. Accordingly, he cannot change his behavior when needed.
4. telnet proxying will only work for cisco and for some other devices that ask for a username and password exactly as recorded in your code. If the device asks not for "Username", but for "Login" or "User", then nothing will work.
5. very specific solution. There are many networks and not on cisco hardware. What happens if all vendors are added to the universal ssh-input? The code will turn into a mess.

Better create a new input type, inherit the class from ssh and make specific changes. At the same time, you can make these changes private, because the project is already difficult without a permanent maintainer, and there will simply be no one to debug and fix your specific changes with unobvious side effects.

P.s.:
Do not assign functions to routers that they should not perform. Get a dedicated jump-host. Nowadays, creating a virtual machine with minimal resources costs almost nothing. Or start a couple RPi. It is really safer, more flexible, scalable (did you forget that cisco has a limit on the number of sessions and is quite small by default? Errors will be unobvious, and questions again to the maintainer, because he is responsible for fundamental things).

mortzu:
Please process the checklist

Issue #2228, opened 2020-12-03
[Help] configure ssh_proxy for jumpbox


davama:
Hello,

Trying to understand how `ssh_proxy` works but looking at old issues and this brief [doc](https://github.com/ytti/oxidized/blob/master/docs/Configuration.md#ssh-proxy-command) I cant seem to figure out.

So oxidized is one host and my myjumpbox (which accesses all the devices) is another host

My goal is to have oxidized ssh into the myjumpbox via sshkey then from there ssh into the network devices via password. (all of this by the same user `oxidized`)

I've tried moving the `ssh_proxy` around but no success.

```
username: oxidized
password: blabal
...
...
vars:
  remove_secret: true
  #auth_methods: [ "password" ]
  auth_methods: [ "publickey", "password" ]
  ssh_keys: "~/.config/oxidized/.ssh/id_ed25519"
...
...
source:
  default: sql
  sql:
    adapter: sqlite
    database: "/home/oxidized/oxidized-repo/region-oxidized.db"
    table: devices
    map:
      name: name
      ip: ip
      model: model
      group: group
    vars_map:
      ssh_proxy: proxy
```
My sql db looks like so:
```
<netdevicehostname>|<ipaddr>|ios|USA-HT_WIFI|myjumpbox
```
Here is a sample log of what I get:
```
D, [2020-12-03T14:44:37.053290 #30086] DEBUG -- : lib/oxidized/job.rb: Starting fetching process for netdevice at 2020-12-03 14:44:37 UTC
D, [2020-12-03T14:44:37.053485 #30086] DEBUG -- : lib/oxidized/input/ssh.rb: Connecting to netdevice
D, [2020-12-03T14:44:37.053633 #30086] DEBUG -- : AUTH METHODS::["publickey", "password"]
D, [2020-12-03T14:44:37.054652 #30086] DEBUG -- net.ssh.transport.session[3ff09e078428]: establishing connection to <ipaddr>:22 through proxy
Bad stdio forwarding specification '<ipaddr>:22'
W, [2020-12-03T14:44:37.063932 #30086]  WARN -- : <ipaddr> raised Net::SSH::Proxy::ConnectError (rescued RuntimeError) with msg "command failed: ssh -o StrictHostKeyChecking=no myjumpbox -W <ipaddr>:22"
D, [2020-12-03T14:44:37.064098 #30086] DEBUG -- : lib/oxidized/node.rb: Oxidized::SSH failed for netdevice
[...]

PR #2245, opened 2021-01-08
add bracket on node if using proxy and node has ipv6 address

davama:
Should solve: https://github.com/ytti/oxidized/issues/2228

Note:
Not a ruby dev so i am sure there is a better way to do it.
Open to suggestions.

I also tested if node has an ipv4 address and it also works. So basically, at least from what i can tell, it does not break anything. :laughing: 

Thanks,
Dave

### DIFF
--- a/lib/oxidized/input/ssh.rb
+++ b/lib/oxidized/input/ssh.rb
@@ -140,11 +140,7 @@ module Oxidized
         if (proxy_port = vars(:ssh_proxy_port))
           proxy_command += "-p #{proxy_port} "
         end
-        if "#{@node.ip}".include? ":"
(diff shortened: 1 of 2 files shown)


davama:
find it strange that Travis complains when the change is soo minor....

I think other projects on github were having travis issues recently....

davama:
Sorry for the bump...

Just wanted to bring to the attention, since this change is literally the addition of two characters :smile: 
There is another proxy related PR which are very [interesting](https://github.com/ytti/oxidized/pull/2268). Hopefully it's merged soon too

Happy Friday!!

ytti:
Apologies and thank you.

davama:
@ytti 
No apologies needed here!
Thank you very much!!

have a good weekend

PR #2270, opened 2021-02-24
Adding polling time as Oxidized configuration parameter.

elpatopedraza:
## Pre-Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - model for zte c300 and c320 olt (@glaubway)
 - model for LANCOM (@systeembeheerder)
 - model for Aruba CX switches (@jmurphy5)
+- add polling time configuration (@elpatopedraza)
 
 ### Changed
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -283,6 +283,15 @@ rest: 127.0.0.1:8888
 rest: 10.0.0.1:8000/oxidized
 ```
 
+## Backup Polling Time
+
+Oxidized can be configured to start the backup configuration every x seconds, where x is the amount defined in `polling_time` parameter. The default value is 1.
+
+```yaml
+# Backup process starts every second
+polling_time: 1
+```
+
 ## Triggered backups
 
 A node can be moved to head-of-queue via the REST API `GET/POST /node/next/[NODE]`. This can be useful to immediately schedule a fetch of the configuration after some other event such as a syslog message indicating a configuration update on the device.
--- a/lib/oxidized/config.rb
+++ b/lib/oxidized/config.rb
@@ -11,7 +11,6 @@ module Oxidized
     ModelDir  = File.join Directory, %w[lib oxidized model]
     SourceDir = File.join Directory, %w[lib oxidized source]
     HookDir   = File.join Directory, %w[lib oxidized hook]
-    Sleep     = 1
 
     def self.load(cmd_opts = {})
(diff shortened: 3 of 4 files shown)


ytti:
Can you explain the utility of this? When would you like to change this value?

elpatopedraza:
> Can you explain the utility of this? When would you like to change this value?

Hi @ytti, sure I'll answer your questions explaining our current situation:

Seems like our organization has some rate limiter over the network devices my team has to monitor. This means that we can't create more that certain amount of SSH sessions per minute from the same IP address. We are using Oxidized to monitor a couple of network devices, but seems like the backup process is so fast that Oxidized tries to run multiple times the backup process in a single minute, so the rate limiter is blocking some SSH sessions after certain limit and Oxidized is marking those attempts as failed (besides that the rate limiter is triggering some alarms due to exceeded SSH sessions).

Basically this new feature allows to any user define the Oxidized polling time, so the platform starts the backup process and waits for x seconds until start the next iteration.

This fix our issue and would be nice to have it in the next Oxidized version.

I hope this little story is clear for you :)

ytti:
We should only try to fetch every device once per `interval`. But if we don't succeed, we try same box `retries` times back-to-back, before giving up.

This parameter you are changing is about how often we check if we have any work to do. So it shouldn't cause any network connections, if everything has been done.

If we try to hit boxes more often than `interval` * `retries`, then something is off.

elpatopedraza:
> We should only try to fetch every device once per `interval`. But if we don't succeed, we try same box `retries` times back-to-back, before giving up.
> 
> This parameter you are changing is about how often we check if we have any work to do. So it shouldn't cause any network connections, if everything has been done.
> 
> If we try to hit boxes more often than `interval` * `retries`, then something is off.

@ytti I understand your point regarding retries and interval parameters.

Let's suppose we have only 2 network devices registered, the rate limiter only allows 2 SSH sessions per minute and the interval parameter is the default one (3600 seconds). Oxidized runs the first backup iteration successfully (less than a minute)...good so far... Oxidized waits 1 second and starts the second iteration... good so far... waits 1 second and starts the third iteration... at this point our rate limiter starts to trigger issues regarding Oxidized IP and unfortunately this policy cannot be changed by my team because it comes from security policies. So we thought the best approach was to configure this parameter to avoid this.

Please let me know your comments.

ytti:
It shouldn't start the 3rd iteration until 3600s have passed. It should fetch both of the devices back-to-back, then do nothing for 3600s. 
If it does immediately fetch the 1st node again, then something is broken in the code :(

elpatopedraza:
> It shouldn't start the 3rd iteration until 3600s have passed. It should fetch both of the devices back-to-back, then do nothing for 3600s.
> If it does immediately fetch the 1st node again, then something is broken in the code :(

Ok, I'll check this more carefully with the team and I'll back to you. Thanks.

ytti:
Generally what the algorithm tries to do, it tries to ensure that we spread the work over as long period of time as possible. This is on premise the we want predictable CPU use, so we don't have to provision for high peak use, when box is mostly silent. However the 'slowest' we go, is as fast as single thread can go. So if the algorithm determines we can fetch all your devices with single thread, when that single thread still fetches back-to-back all the devices as fast as it can, and if there is time remaining it does nothing.

So if you have very strict SSH limits and the single thread performs too fast, you could exceed those limits. And if you need this sleep time configurable (awkward fix to your problem) I can accept it on some other name like 'loop_sleep' or something. 

The whole reason there is _any sleep_ at all, is to avoid busy loop, when we don't have anything to do, we'd potentially use very much CPU detecting we don't have anything to do rapidly over and over again.

mortzu:
Any news? Can this be closed?

Issue #90, opened 2015-03-13
Running SyslogMonitor


MrRJ45:
What is the best way to run the Syslog Monitor? I have tried this:

[program:oxidized-syslog]
command=/var/lib/gems/1.9.1/gems/oxidized-0.4.1/extra/syslog.rb
autostart=true
autorestart=true
user=oxidized
stderr_logfile=/var/log/oxidized-syslog.err.log
stdout_logfile=/var/log/oxidized-syslog.out.log

It fails with this:

/usr/lib/ruby/1.9.1/rubygems/custom_require.rb:36:in `require': cannot load such file -- ./rest_client (LoadError)
        from /usr/lib/ruby/1.9.1/rubygems/custom_require.rb:36:in `require'
        from /var/lib/gems/1.9.1/gems/oxidized-0.4.1/extra/syslog.rb:17:in `<main>'

Hard coding the full path to rest_client.rb works fine, but all of this is version specific.

What would be nice would be to have /usr/local/bin/oxidized-syslog which can be ran and doesnt depend on the version number or hard coding paths :)

MrRJ45:
The ability to also pass arguments would be great too, for example, running using port 1514 doesn't require extra privileges.

ytti:
I've not tried the example more less since initial commit, so it's not exactly user-friendly right now, it more explains how you can benefit from some oxidized features.
Perhaps it should be rolled in and be more 'productized', so it would work generally.

So first question, does it work, at all, today? I understand that as you're doing init script for it, you've been able to succesfully run it, which makes me quite impressed already. You're getting actual user in git (git blame) and all? I'm amazed if we've not changed the API enough to break it.

Perhaps ea7503d242 fixes your issue?

MrRJ45:
ea7503d does fix that, thank you :) I've changed syslog.rb and it starts, and listens on the port.

I'll test it shortly and let you know what happens!

ytti:
How about 02b3b0f8f15b does it fix your UDP port issue?

MrRJ45:
02b3b0f works fine, if you pass a port number. It should default to 514 if no argument is given.

ytti:
Fixed in fb0e4e191a

MrRJ45:
I have set up a 2960 to log the config changes, and SyslogMonitor is getting this as I saw an error connecting to port 8888 (I have it set to bind to the public IP).

[...]

PR #2275, opened 2021-03-05
Fix /node/show returning the wrong node when several share the same IP

sbraz:
## Pre-Request Checklist

- [ ] Tests added or adapted (try `rake test`) → I get `34 runs, 119 assertions, 2 failures, 0 errors, 0 skips` on my branch and master so I assume the breakage isn't related to this change
- [ ] Changes are reflected in the documentation → I don't think this is documented anywhere

## Description
Hi,
Before this change, calling `/node/show/<node_name>` would return the first node whose name *or* IP address was `node_name`.
This could cause problems when two nodes had the same IP address but a different name. In that case, the URL would not always return the correct node.

After this change, only the node name is taken into account.

As far as I'm aware, the name is the unique identifier of a node so this shouldn't break anything but please let me know if I'm wrong.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - improved scrubbing of show chassis in ironware model (@michaelpsomiadis)
 - fixed snmp secret handling in netgear model (@CirnoT)
 - filter next periodic save schedule time in xos model output (@sargon)
+- fixed /node/show returning the wrong node when several nodes share the same IP address (@sbraz)
 
 ## [0.28.0 - 2020-05-18]
 
--- a/lib/oxidized/nodes.rb
+++ b/lib/oxidized/nodes.rb
@@ -131,7 +131,7 @@ module Oxidized
     end
 
     def find_index(node)
-      index { |e| [e.name, e.ip].include? node }
+      index { |e| e.name == node }
     end
 
     # @param node node which is removed from nodes list


ytti:
https://github.com/ytti/oxidized/commit/4455a5b91accda5da7c1b8128cb045c7eb7bb5da

With proposed change it wouldn't work if we give IP as an argument, when it is not same as name. I fear that in the field there are all possible combinations so doing anything here will probably break some deployments.

sbraz:
@ytti I don't understand the commit you're linking to. Do you mean it used to just filter based on the name and it was changed later?

In any case, we can't have a reliable method that filters based on either the IP or the name. Especially when it breaks the links within the Web UI.

If the solution is creating a new method and updating the links from the Web UI, then I don't mind. I just want a reliable way to get info about my nodes, the old existing method can be left as it is.

ytti:
Yes, we used to match on name exclusively. I'm not sure what the fix is, without breaking something that exists. We probably should refactor so that we are explicit on by which key we want to find, and when possible input as many keys as possible. For web interaction, we actually already have the node object, as we already listed the data, so we could be very explicit.

sbraz:
Yes, we need to separate queries by name or by IP. I tracked the change that added the IP to something related to calls to `/nodes/next` in 2015: b0b50e8bba592bf9c96c198087dc19b077044f51.
This specific comment is interesting :stuck_out_tongue: https://github.com/ytti/oxidized/issues/90#issuecomment-79068563

PR #2399, opened 2021-11-13
Update fabricos.rb

supermanchw:
remove power usage display from output , like "Powser Usage: "


mortzu:
Closes because another PR fixed this already

PR #2401, opened 2021-11-16
Output for Power Usage has changed [FabricOS]

suom1:
## Description
In FabricOS v8 => the output for Power Usage has changed to `Power Usage:` instead of `Power Usage (Watts):`.
Adding `Power Usage:` to excluded lines to get rid of continues config changes when power draw on device changes.

Change is verified working on FabricOS v8 and v9 devices (8.2.2b & 9.0.1b4).


suom1:
Duplicate of #2399

Issue #2427, opened 2021-12-08
Mikrotik Backups Empty


azielke:
PR #2384 seems to break Mikotik Backups. `/export show-sensitive` is not available on current routeros 6.X and only available starting with routeros 7.

The Backups are saved as "empty files" (no error reported).

```
> /export show-sensitive
expected end of command (line 1 column 9)
```

mortzu:
Its fixed with #2435

PR #2435, opened 2021-12-16
routeros: /export params depending on version

azielke:
## Pre-Request Checklist

## Description
Fix Mikrotik Routeros Backups being empty for routeros versions <7.

Closes issue #2427

### DIFF
--- a/lib/oxidized/model/routeros.rb
+++ b/lib/oxidized/model/routeros.rb
@@ -17,6 +17,11 @@ class RouterOS < Oxidized::Model
   end
 
   cmd '/system package update print without-paging' do |cfg|
+    cfg.split("\n").each{ |line|
+      if line =~ /installed-version: ([0-9])/
+        @ros_version = $1.to_i
+      end
+    }
     comment cfg
   end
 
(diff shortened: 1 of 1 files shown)


jplitza:
Tested successfully (without `:remove_secrets`) on ROS 6.48.6 and ROS 7.1.

@azielke Could you please fix the [RuboCop styling warnings](https://app.travis-ci.com/github/ytti/oxidized/jobs/552557046)?

lib/oxidized/model/routeros.rb
```diff
@@ -25,7 +28,12 @@ class RouterOS < Oxidized::Model
   end
 
   post do
-    run_cmd = vars(:remove_secret) ? '/export hide-sensitive' : '/export show-sensitive'
+    Oxidized.logger.debug "lib/oxidized/model/routeros.rb: running /export for routeros version #{@ros_version}"
+    run_cmd = if @ros_version >= 7
+                vars(:remove_secret) ? '/export hide-sensitive' : '/export show-sensitive'
+              else
+                vars(:remove_secret) ? '/export hide-sensitive' : '/export'
+              end
```
jplitza, line 36: May I suggest the following?
```ruby
run_cmd = if vars(:remove_secret)
            '/export hide-sensitive'
          elsif @ros_version >= 7
            '/export show-sensitive'
          else
            '/export'
          end
```

jplitza:
Out of curiosity, what happens then no line matches `installed-version`? I have no clue about Ruby, but I guess that `@ros_version = nil` and `@ros_version >= 7` throws a `NoMethodError`? Maybe the conditional should cover that case:
```ruby
if not @ros_version.nil? and @ros_version >= 7
```

azielke:
installed-version *should* always be present - but it would indeed throw `NoMethodError` (which would show an error when fetching the config). I think doing the check for nil would indeed be best as this would behave exactly the same as it did previously.

jplitza:
LGTM.

Since there's not mention of the newly added ROS 7 support in the changelog yet, it doesn't really seem mandatory to add an entry for *this change* - but the ROS 7 support should be added. Maybe a maintainer can do that outside of this PR

PR #2445, opened 2021-12-27
Support for Dell EMC Networking OS9

baldoarturo:
## Pre-Request Checklist

## Description

Added support for Dell EMC Networking OS9. This is "cisco like" but way different from OS10, so it needed a new model.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - model for Aruba CX switches (@jmurphy5)
 - model for NEC IX devices (@mikenowak)
 - Added docs for Dell/EMC Networking OS10 devices (@davromaniak)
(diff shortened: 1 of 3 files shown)


mortzu:
Model looks mostly like the OS10 one. Do we need both of them?

baldoarturo:
A separate model is not totally needed as OS9 can run the same set of commands for 
```
show inventory
show inventory media
show running-config
```

But, OS9 is a totally different software platform which does not share the same command line structure on other commands.

I guess that as long the user is aware of this, you can maybe map OS9 devices to use the OS10 model.

mortzu:
Does your switch supports the command "show running-configuration" instead of "show running-config"?

baldoarturo:
I just tested and `show running-configuration` works. So well maybe I am trying to reinvent the wheel on this.
Maybe it is enough by adding docs explaning that the OS10 model works for OS9 too. I can take care of that if you agree,

mortzu:
Please

PR #2446, opened 2021-12-27
Support for Ubiquiti UFiber v3 OLTs

baldoarturo:
## Pre-Request Checklist

## Description

This closes my own issue ytti#2436

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - model for Aruba CX switches (@jmurphy5)
 - model for NEC IX devices (@mikenowak)
 - Added docs for Dell/EMC Networking OS10 devices (@davromaniak)
+- model for Ubiquiti UFiber OLTs v3 (@baldoarturo)
 
 ### Changed
 
--- a/docs/Model-Notes/UFiber-v3.md
+++ b/docs/Model-Notes/UFiber-v3.md
@@ -1,0 +1,8 @@
+# Model Notes
+UFiber OLTs works with a .tar.gz file for backups, as it comprises not just the running-config, but some additional config files.
+The output is obtained in base64 enconding.
+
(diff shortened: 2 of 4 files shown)


mortzu:
As in https://arturobaldo.com.ar/digging-into-ubiquitis-ufiber-olt/ UFiber OLT is an EdgeOS. So maybe you can use the edgeos model?

mortzu:
If we need the HTTP-stuff maybe you can use some code as in https://github.com/ytti/oxidized/blob/aaa301196e6186cc6399db8714f62aeaf281dd26/lib/oxidized/model/pfsensehttps.rb

baldoarturo:
Thanks Mortzu, that blog entry is mine indeed.  This platform is based on EdgeOS, but the main part of the config is located on separate files. Also, the only way to provision the OLT is using a .tar.gz file which comprises a folder structure with all those files. Most of the config is not being done on the CLI, but only via the web API.

I checked the pfsense model when I started with this, but the main diference with it is that pfsense uses a XML file which comprises the entire config. UFiber uses this .tar.gz which looks like this

```
Arturo@ARTURO-DELL c:\Program Files\WinRAR
# tar -tf c:\Users\Arturo\Downloads\edgeos_olt-ubnt-sm-01_20211229.tar.gz
config/
config/auth/
config/onu_config.json
config/scripts/
config/scripts/post-config.d/
config/support/
config/udapi-bridge/
config/user-data/
config/snmp/
config/snmp/superuser_pass
config/config.boot
config/config.boot.2015-01-01-0001.pre-migration
config/config.boot.2015-04-14-0130.pre-migration
config/wizard/
config/wizard/feature/
```

mortzu:
And you want to store the tar as binary in oxidized?

baldoarturo:
Indeed. Right now I am using this model to store the .tar.gz file encoded in base64, so it is easier for git to process the diffs
image

mortzu:
I really don't like storing base64 encoded binary data in git...

Issue #2471, opened 2022-02-09
Monkey-patching a command with `clear: true` removes all commands


mjbnz:
From documentation, "[new code for parsing 'show version', replaces the existing definition in the model](https://github.com/ytti/oxidized/blob/master/docs/Creating-Models.md#monkey-patching-blocks-in-existing-models)" (second example in the Monkey-patching section).

But, from model.rb [Lines 48-53](https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/model.rb#L48-L53), and [Lines 99-100](https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/model.rb#L99-L100) :

```
      def cmd(cmd_arg = nil, **args, &block)
        if cmd_arg.class == Symbol
          process_args_block(@cmd[cmd_arg], args, block)
        else
          process_args_block(@cmd[:cmd], args, [cmd_arg, block])
        end

.....

      def process_args_block(target, args, block)
        if args[:clear]
          target.replace([block])
```

a command like `show version` in the example is not a Symbol, so target becomes Array stored in the `@cmd[:cmd]` hash entry, meaning that instead of replacing the single block within that array that matches the `cmd_arg`, the entire array is replaced.

I'm not fluent in Ruby, so apologies if I'm missing some syntactic sugar here that means I'm way off the mark.

mjbnz:
Closed by merge, \o/

PR #2472, opened 2022-02-09
Prevent clearing of all commands when using `clear: true` with an explicitly defined command

mjbnz:
## Pre-Request Checklist

First two, I have no idea... not a Ruby coder, sorry!

## Description

I wrote a quick fix to close issue #2471 .

The fix is to check the type of block passed to `process_args_block()` when `clear: true` has been set as an arg, and if it's an Array, filter the target to remove any entries with the command given, then push the new command on to the end.

A better approach might be to use `map!` first to replace a single occurence, then filter afterwards to remove other occurences in order to retain some sort of ordering - however, I'm not confident enough to try to do that.


mjbnz:
Force push was a rebase onto rubocop-fixes to clear Travis check

mjbnz:
Replaced with #2486

Issue #2096, opened 2020-05-15
Add MySQL ssl_mode configuration variable


mrplow87:
on a recent CentOS 8, using the SQL input with MySQL, it uses encryption by default but fails because no keys and certificates are configured. 
It would be awesome to have the `ssl_mode` (and maybe all the other "SSL" configuration keys, if you want to use encrypted connection to your MySQL host) configurable in the Oxidized configuration file.

Not knowing enough about Ruby to propose a proper change for that, my current workaround is to change `lib/oxidized/source/sql.rb` and add `ssl_mode: 'disabled'` to the `Sequel.connect()` parameters.

PR #2515, opened 2022-04-05
Extended mysql source (TLS support) 

glaubway:
## Pre-Request Checklist

## Description
Extended mysql source configuration to include TLS options

Close the issue #2096


ytti:
I would make this drier. You are repeating yourself. Instead you could make an hash then add SSL keys to the hash if SSL is defined and then call the constructor with the hash.

glaubway:
Sorry, but I don't know ruby (have experience with another languages), so I just made it as I understand this and can say that this solution is working in my environment.

glaubway:
I made it a bit clean.

mortzu:
Please fix the merge conflicts

PR #2528, opened 2022-05-01
Add option: use_max_threads

### LINKED ISSUE #2527: Enhancement request: forced-multithread operation to support "snapshot"-like collection
(See #457 for some background detail.  I may as well just quote myself, more or less:)

I, like others, have a need to collect as much of a "snapshot" as possible, so I would prefer that - at the `interval` time - Oxidized should spin up as many threads as possible in order to accomplish data collection as rapidly as possible.  Oxidized is not resource-limited in my environment, and I still have max_threads to control that anyway.

This would also be a useful feature for troubleshooting my production instance more easily - when something doesn't work right and I have to restart Oxidized (e.g. editing a custom model), it can take almost an hour before Oxidized finishes its initial single-threaded poll of all the devices and reaches steady-state, at which point I can begin troubleshooting usefully.

This might make more sense when I say that many of my devices take 5-10min to collect running-config, so when those get serialized, some of the assumptions about e.g., jobs taking 5sec by default, are waaaaaay out of sync with my environment.

Ideal (for me) would be a new config option along the lines of `use_max_available_threads: [yes|no]` that could be changed and reloaded at runtime.  Although some discussion has previously occurred, I don't see anything like that today.

If I'm missing something, or if this is a _trivial_ local modification, great! - please tell me.
Otherwise, it looks like the section of code needing changed is jobs.want starting at https://github.com/ytti/oxidized/blob/8179f48545d94a1c6b3a096153f70e9eabff1e20/lib/oxidized/jobs.rb#L35 but I'm barely able to read this code base, forget about modifying core functionality.

Alternately, if MAX_INTER_JOB_GAP (at https://github.com/ytti/oxidized/blob/8179f48545d94a1c6b3a096153f70e9eabff1e20/lib/oxidized/jobs.rb#L4) was a config file parameter, and able to be reloaded at runtime, that might suffice?  I don't think I would want that permanently set to e.g. 1 sec, but I'm unclear on what negative effects that might have.

seros1521:
Allows to backup using the maximum number of threads, without using
autotuning (but not more than the maximum)

## Pre-Request Checklist

## Description
Fixes #2527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - model for Cambium ePMP radios (@martydingo)
 - Dockerfile rebased to phusion/baseimage-docker focal-1.2.0
 - model for Lenovo Network OS (@seros1521)
+- new option: use_max_threads
 
 ### Changed
 
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -184,7 +184,11 @@ model: junos
 interval: 3600 #interval in seconds
 log: ~/.config/oxidized/log
 debug: false
-threads: 30
+threads: 30 # maximum number of threads
+# use_max_threads:
+# false - the number of threads is selected automatically based on the interval option, but not more than the maximum
+# true - always use the maximum number of threads
+use_max_threads: false
 timeout: 20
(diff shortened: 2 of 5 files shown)


mortzu:
Can you please fix the conflict

seros1521:
Fixed.

PR #2553, opened 2022-05-25
Add missing check for if to do `enable` for Adtran devices

### LINKED ISSUE #2386: Adtran backup issue
I'm struggling to back up Adtran TA5000 devices on a LibreNMS integrated install. The error I get in debug is 
```
raised NoMethodError with msg "undefined method `+' for nil:NilClass"
```

I have sorted out the names using the model_map etc. The debug says that it can log in ok and the log under logs/IPADDRESS shows 

```
Last Login Date      : Oct 20 2021 10:23:46
Last Login Type      : IP Session(CLI)
Login Failures       : 0 (Since Last Login)
                     : 0 (Total for Account)
$hostname>
```

I have seen issue #2044 and think that it might be related but I'm unsure of the fix? 

Thanks all

repnop:
This adds a missing check for if the `enable` command should be run on Adtran devices. Without the check, it causes an exception to be raised if the `enable` command isn't specified in the `vars` configuration setting. Fixes #2386.


mortzu:
Please add your change to the CHANGELOG

repnop:
added :+1:

PR #2578, opened 2022-07-05
fix adva prompt

momorientes:
## Description

Change prompt for the adva module to be any combination of word characters, since the defualt `ADVA-->` prompt can be overridden in the config via:
```
home
network-element ne-1
  prompt "cpe123"
```


mortzu:
Please fix the merge conflict

Issue #2531, opened 2022-05-04
Netgear won't save correctly


domrcrel:
Building Oxidized from `master` and trying to connect to a Netgear host (in my case, Netgear M4300-24X) fails. All the credentials are correct and it can connect to the host over SSH, and is even able to get the config, but still reports a failure.

I believe this is due to not including any `pre_logout` in `netgear.rb`. I looked in there and found this:

````
    # quit / logout will sometimes prompt the user:
    #
    #     The system has unsaved changes.
    #     Would you like to save them now? (y/n)
    #
    # So it is safer simply to disconnect and not issue a pre_logout command
````
This is correct in a way, as every time I (or Oxidized) connects to the host via SSH this prompt does appear, even if no changes were made. As Oxidized is not here to make changes though, only record them, I don't believe ignoring this is correct. At the moment, the system doesn't logout as no `pre_logout` is issued, so it hangs onto the connection and times out because of it. 

I added the below to my local copy of `netgear.rb`, after the `post_login` commands:

````
    pre_logout 'quit'
    pre_logout 'n'
````
It now saves the config correctly.

Is there a reason for having the current version behave how it does?

Issue #1100, opened 2017-11-15
Netgear model: Does not detect end of configuration file


pv2b:
It seems part of #920 is still unsolved.

@candlerb mentioned:

> So for some reason it doesn't accept the returned prompt, even though it appears to match the prompt regexp in netgear.rb

I am seeing the exact same issue even with the fixed code. After running "show running-config", the prompt is returned at the end, but it seems that it is not matched correctly. Instead I get a timeout like this:

```
Nov 15 16:09:11 mgt02 oxidized[17108]: W, [2017-11-15T16:09:11.080677 #17108]  WARN -- : 10.0.9.6 raised Timeout::Error with msg "execution expired"
Nov 15 16:09:11 mgt02 oxidized[17108]: W, [2017-11-15T16:09:11.934272 #17108]  WARN -- : swge001-sto-dc1.ita.local status no_connection, retry attempt 1
Nov 15 16:09:35 mgt02 oxidized[17108]: W, [2017-11-15T16:09:35.985670 #17108]  WARN -- : 10.0.9.6 raised Timeout::Error with msg "execution expired"
Nov 15 16:09:36 mgt02 oxidized[17108]: W, [2017-11-15T16:09:36.941394 #17108]  WARN -- : swge001-sto-dc1.ita.local status no_connection, retry attempt 2
Nov 15 16:10:01 mgt02 oxidized[17108]: W, [2017-11-15T16:10:01.144171 #17108]  WARN -- : 10.0.9.6 raised Timeout::Error with msg "execution expired"
Nov 15 16:10:01 mgt02 oxidized[17108]: W, [2017-11-15T16:10:01.948562 #17108]  WARN -- : swge001-sto-dc1.ita.local status no_connection, retry attempt 3
Nov 15 16:10:26 mgt02 oxidized[17108]: W, [2017-11-15T16:10:26.034609 #17108]  WARN -- : 10.0.9.6 raised Timeout::Error with msg "execution expired"
Nov 15 16:10:26 mgt02 oxidized[17108]: W, [2017-11-15T16:10:26.955191 #17108]  WARN -- : swge001-sto-dc1.ita.local status no_connection, retries exhausted, giving up
Nov 15 16:14:26 mgt02 oxidized[17108]: W, [2017-11-15T16:14:26.007384 #17108]  WARN -- : 10.0.9.6 raised Timeout::Error with msg "execution expired"
[...]

PR #2629, opened 2022-10-03
Make sure SSH session can end on Netgear correctly

domrcrel:
At the moment, it's not possible for Oxidized to save config on a Netgear device when connected over SSH, as it doesn't correctly finish the SSH session since we don't issue any `pre_logout` commands. This looks to have been previously done as Netgear devices do prompt the below each time an SSH session is ended:

````
The system has unsaved changes.
Would you like to save them now? (y/n)
````

As there are no actual changes being done over the SSH session that Oxidized uses though, it's safe to simply choose "n" here and move along.

Ref #1100 #2531


mortzu:

Issue #2296, opened 2021-04-27
Telnet expect model function broken


rgergon:
When using telnet (ssh is fine) the "expect" functionality in models is broken and does not work. This used to work in oxidized-0.21 but is broken at least already in 0.26-3.
I've seen a lot of weird workarounds of other peoples in PRs or unspecific issue reports where I am certain they are all caused by the same issue: The fact that expect simply doesn't work anymore when using telnet. In most cases, the model's expect for pagination handling then causes timeouts but it could be of course more advanced features.
I am not a ruby programmer so I stopped trying to fix this issue after diagnosing it, but there's some FIXME's in the code there (depending on version).
Workarounds until fixed:
* don't use expect in model
* use ssh

candlerb:
I also find that nothing is written to `logs/x.x.x.x-telnet` except the initial username/password login (not even sending of 'enable')

Assuming this logging is done [here](https://github.com/ytti/oxidized/blob/master/lib/oxidized/input/telnet.rb#L104-L107) in `oxidized_expect`:

```
        if Oxidized.config.input.debug?
          @log.print buf
          @log.flush
        end
```

then it implies that `oxidized_expect` is not being called from somewhere else that it should be.

jvkassi:
Hello, 

I had the same issue.
I am using 0.21.1 and everything working as expected :)

Issue #2591, opened 2022-07-25
Netgear devices do not work properly


mWachu:
Hello,

I've just installed 0.28 Version of oxidized and i have encountered issue with pulling configuration from Netgear devices.
<img width="445" alt="Screenshot 2022-07-25 at 23 33 21" src="https://user-images.githubusercontent.com/77017007/180877742-f8584e22-c170-46db-94af-45813b1854ae.png">

It seems like "enable" is not triggered for some reason.

mortzu:
Could you please paste your config file and your device input (yml, output from librenms, etc.)

oyvindmo:
Which switch model is it? I saw similar symptoms on GS308T, which is in their "Smart Managed Pro" series. I have been fiddling with the netgear.rb model file to get it working, and it's now at a stage where it works decently for me: Oxidized can SSH to the switch and extract the config.

But the model file got more messy than I'd like, and for all I know it might break for other Netgear devices, so I wonder if the OS in the Smart Managed Pro series is different enough from the OS that netgear.rb was written for that it might merit a separate model file.

If you're interested, here's a diff showing my changes to lib/oxidized/model/netgear.rb:

    --- a/lib/oxidized/model/netgear.rb
    +++ b/lib/oxidized/model/netgear.rb
    @@ -1,10 +1,16 @@
     class Netgear < Oxidized::Model
       comment '!'
    -  prompt /^(\([\w\s\-.]+\)\s[#>])$/
    +  prompt /((\(?[\w\s\-.]+\)?\s?[#>]\s?)$)/
    +
    +  expect /--More--/ do |data, re|
    +    send ' '
    +    data.sub re, ''
    +  end
     
       cmd :secret do |cfg|
         cfg.gsub!(/password (\S+)/, 'password <hidden>')
         cfg.gsub!(/encrypted (\S+)/, 'encrypted <hidden>')
    +    cfg.gsub!(/snmp community (\S+) (.*)$/, 'snmp community <hidden> \\2')
         cfg.gsub!(/snmp-server community (\S+)$/, 'snmp-server community <hidden>')
         cfg.gsub!(/snmp-server community (\S+) (\S+) (\S+)/, 'snmp-server community \\1 \\2 <hidden>')
         cfg
    @@ -24,20 +30,28 @@ class Netgear < Oxidized::Model
             cmd vars(:enable)
           end
         end
    -    post_login 'terminal length 0'
    +    if !vars(:smartpro) || vars(:smartpro) == false
[...]

PR #2668, opened 2022-11-24
work on cmd method to get expects in models again

arrjay:
## Pre-Request Checklist

several rake tests did fail, but they were all related to git operations? I'm not sure the best ways to test this.
models now work as already documented when using telnet.
I think the maintainer gets to decide where to do that.

## Description
this updates the cmd method to run through oxidized_expect and grab the resulting output buffer. this allows for models using expect (like the GS108T referenced in #2591) to work. I did not update that model here, as this is more for the core expect functionality.

Closes issue #2296


mortzu:
Can you please update CHANGELOG.md

PR #2724, opened 2023-03-03
Use the git class of the ruby git project to improve performance

mwobst:
With this change and a commit count of 37000+, the required loading time of /node/version?node_full=[…] drops from 30+ (even 40) seconds down to not even one … (Note: To avoid class name overwriting, line 2 defines an alternativ class name)

## Pre-Request Checklist

## Description

Hello,

as our oxidized git repository reached way over 30000 commits, performance on showing the config version for a single node was horrible – sometimes even over 40 seconds whereas 25+ was pretty standard (per page load, mind you). Same goes for the diff display.

I am practically new to Ruby (until yesterday I haven’t even seen a single line of Ruby, I believe), so forgive me beginner mistakes … anyway: The Rugged class(? module?) runs over every commit, which is not really effective. There is an issue where that specific use case is adressed (https://github.com/libgit2/rugged/issues/343), but sadly neither the author or anyone else was motivated enough for a solution/a fix.

Nevertheless: In this specific case, Rugged is not beneficial, which is why I am using the same Git class as output/gitcrypt.rb does.

I am not sure, if 'git' has to become a runtime dependency, currently it is only a development dependency – and I had to install via "gem install", otherwise an error occurs. Could someone confirm this? Or is it just a missing line of source code in output/git.rb?

(By the way … am I blind or are there no tests …?)

### DIFF
--- a/lib/oxidized/output/git.rb
+++ b/lib/oxidized/output/git.rb
@@ -80,7 +80,6 @@ module Oxidized
       commits = repo.log.grep("#{path}$")
 
       commits.each do |commit|
-        puts "LIB_SHA: " + commit.sha
         hash = {}
         hash[:date] = commit.date.to_s
         hash[:oid] = commit.sha
--- a/lib/oxidized/output/git.rb
+++ b/lib/oxidized/output/git.rb
@@ -73,11 +73,12 @@ module Oxidized
     # give a hash of all oid revision for the given node, and the date of the commit
     def version(node, group)
       repo, path = yield_repo_and_path(node, group)
-      repo = RubyGit.bare repo
+      repo = RubyGit.bare(repo)
+      repo_log = RubyGit::Log.new(repo, false) #  do not limit amount of commits
 
       i = -1
       tab = []
-      commits = repo.log.grep("#{path}$")
+      commits = repo_log.grep("#{path}$")
 
       commits.each do |commit|
         hash = {}
--- a/lib/oxidized/output/git.rb
+++ b/lib/oxidized/output/git.rb
@@ -1,3 +1,6 @@
+require 'git'
+RubyGit = Git
+
 module Oxidized
   class Git < Output
     class GitError < OxidizedError; end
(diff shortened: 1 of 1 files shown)


jplitza:
If I'm reading [the docs for the log method](https://rubydoc.info/gems/git/Git%2FBase%2FFactory:log) right, by default, it only searches 30 commits. Can you observe that in your tests?

And if I'm reading the issue you linked to correctly, the slowness of the RuggedGit implementation is inherent to the problem and not caused by the implementation. Although the difference to `git log` mainly seems to focus around the handling of merge commits, which we rarely have here...

mwobst:
Hello,

> If I'm reading [the docs for the log method](https://rubydoc.info/gems/git/Git%2FBase%2FFactory:log) right, by default, it only searches 30 commits. Can you observe that in your tests?

But I’m using this one: https://rubydoc.info/gems/git/Git/Log#grep-instance_method
(That Factory class is suited for retrieving the latest x commit, possibly?)

And _Git::Log#grep_ definitely fetches more than 30 commits: For a certain switch, three versions exists. The corresponding commit dates are from 2023 and 2022 – and we already have 5700+ Commits in 2023 alone.

> 
> And if I'm reading the issue you linked to correctly, the slowness of the RuggedGit implementation is inherent to the problem and not caused by the implementation. Although the difference to `git log` mainly seems to focus around the handling of merge commits, which we rarely have here...

Not sure I can follow the first sentence … what I have seen in the code is that every commit is picked to be "analyzed" … and this takes some time. Even with (for example) only 1ms per commit processing, this would result in ~37000ms = 37 seconds.
Concerning merge commits: We had none in our repository(*), only commit with messages like "   update {hostname}".

(* until I made one myself for testing purposes (deleting old devices), but the performance issues happenen before)

jplitza:
> Hello,
> 
> > If I'm reading [the docs for the log method](https://rubydoc.info/gems/git/Git%2FBase%2FFactory:log) right, by default, it only searches 30 commits. Can you observe that in your tests?
> 
> But I’m using this one: https://rubydoc.info/gems/git/Git/Log#grep-instance_method (That Factory class is suited for retrieving the latest x commit, possibly?)

After digging into that gem a bit more, it seems that this `count=30` is passed as parameter to `git log --max-count`, which limits the number of commits that are *returned*, not those searched. So if you view the history of a device with more than 30 config revisions, you probably will only see the last 30.

And now that I know that this gem actually calls the `git` command line utility, I'm a bit taken aback.

Maybe a PR along the lines of [this comment's `entry_changed?` method](https://github.com/libgit2/rugged/issues/343#issuecomment-38767239) would give enough of a speedup without calling an external program?

(I am not a maintainer, though. It's not my call whether calling `git` is a problem or not.)

mwobst:
> > > If I'm reading [the docs for the log method](https://rubydoc.info/gems/git/Git%2FBase%2FFactory:log) right, by default, it only searches 30 commits. Can you observe that in your tests?

Thanks for pointing out – it’s exactly the case. I made a new commit, although I am not exactly satisfied with creating a separate Git::Log instance. But: changing the count attribute via `repo.log.instance_variable_set(:@count, {… integer or false or whatever …})` raises no errors, but does not change the attribute either (and there is no overwrite method defined).

> Maybe a PR along the lines of [this comment's `entry_changed?` method](https://github.com/libgit2/rugged/issues/343#issuecomment-38767239) would give enough of a speed-up without calling an external program?

Well, it still parses each commit. Maybe each iterations would be somewhat faster but it doesn’t change the fact that filtering by hostname happens after crawling through the complete commit swamp … and well, as example: A speed-up from 30 to 15/10(?) seconds would be less bad than before … but still too bad. :-/

(And that amount will raise over time. We’re using oxidized only for around a year, if at all … in 5 years the problem will be even bigger)

jplitza:
> (And that amount will raise over time. We’re using oxidized only for around a year, if at all … in 5 years the problem will be even bigger)

I know the problem, but to be honest, Oxidized itself is not the only problem here. Our repo has 127,385 commits, and Gitlab says it needs 3,3 GB. Cloning or even pulling after some time takes ages, and the only viable option seems to be to roll over to a new repo on a regular basis.

Also, I only access history etc. in Gitlab or CLI because of the size and inefficient Ruby implementation. But OTOH, what those tools offer is quite a bit more than Oxidized ever could (or should) - especially `git blame`

PR #2738, opened 2023-04-01
fix rubocop warnings

aschaber1:
## Pre-Request Checklist

## Description
I tried to fix rubocop issues. There might be more to come.


aschaber1:
down to 7 offenses 🚀

mortzu:
Can I merge?

aschaber1:
@mortzu , should I rebase (and reduce number of commits) first? I tried to have atomic commits, which turned out to be quiet a few. There are a few tests I had to skip, because my ruby skills are not good enough to fix them. 

let me update the readme that it now requires ruby >= 3

besides that, I'm happily running this for the last couple of weeks.

aschaber1:
@mortzu you can merge now, thank you :) added a test to make sure all quality targets are hit :)

Haeki:
This introduced a Bug in the `git.rb` file. 
In [Line 82](https://github.com/ytti/oxidized/blob/637434b94055d4227aaae295958157a9af0cd788/lib/oxidized/output/git.rb#L82) `next if commit.diff(paths: [path]).empty?` the return value of diff has no empty function. The old line `next if commit.diff(paths: [path]).empty?` was correct. I found this [commit](https://github.com/ytti/oxidized/commit/c1b3aa871bfde4bb5031ad3853a3ae198dbdffd9) where the same Bug was fixed in 2019.

Issue #2715, opened 2023-02-15
Wrong number of arguments given (given 0, expected 1+)


LontzroV:
Good day,
My oxidized git versioning has stopped working. Ruby 2.7.0 and Oxidized 2.7.0 and Oxidized web 0.13.1. The install is from last year in June so the git.rb is already set to walker.push(repo.head.target.oid). The error I'm getting on the web interface is:
ArgumentError at /node/version
wrong number of arguments (given 0, expected 1+)
file: versions.haml
location: count
line: 24
Any pointers or hints will be very appreciated!

Issue #2282, opened 2021-03-16
Unable to view versions. 


githubuser5842:
When clicking the version button i am presented with the following error

"ArgumentError at /node/version
wrong number of arguments (given 0, expected 1+)"

I am running Oxidized v.0.28

githubuser5842:
I'm using debian and this only happens if I run it as a service. Below is the oxidized.service file that is in /lib/systemd/system


[Unit]
Description=Oxidized - Network Device Configuration Backup Tool
After=network-online.target multi-user.target
Wants=network-online.target

[Service]
ExecStart=/usr/local/bin/oxidized
User=root
KillSignal=SIGKILL
Environment="OXIDIZED_HOME=/etc/oxidized"
Restart=on-failure
RestartSec=300s

[Install]
WantedBy=multi-user.target

PR #2758, opened 2023-04-19
Fix git diff detection

Haeki:
## Pre-Request Checklist

~~- [ ] Tests added or adapted (try `rake test`)~~
~~- [ ] Changes are reflected in the documentation~~
~~- [ ] User-visible changes appended to [CHANGELOG.md](/CHANGELOG.md)~~

## Description

Fixes a bug caused by the RuboCop autocorrect.
In [/lib/oxidized/output/git.rb](https://github.com/ytti/oxidized/blob/637434b94055d4227aaae295958157a9af0cd788/lib/oxidized/output/git.rb) the code in [line 82](https://github.com/ytti/oxidized/blob/637434b94055d4227aaae295958157a9af0cd788/lib/oxidized/output/git.rb#L82) caused an error. 
Cause is `next if commit.diff(paths: [path]).empty?`, the return value of diff has no `empty?` function. 
The old line `next if commit.diff(paths: [path]).empty?` was correct. 
The same mistake was corrected in this [commit](https://github.com/ytti/oxidized/commit/c1b3aa871bfde4bb5031ad3853a3ae198dbdffd9) back in 2019.
This time i also disabled the RuboCop check for this line so that autocorrect doesn't change this again.

Possibly this also solves issues [2715](https://github.com/ytti/oxidized/issues/2715) and [2282](https://github.com/ytti/oxidized/issues/2282) because the error when clicking the "Version" button led me to this bug. And this bugfix solved it for me.

### DIFF
--- a/lib/oxidized/output/git.rb
+++ b/lib/oxidized/output/git.rb
@@ -80,10 +80,8 @@ module Oxidized
       tab = []
       walker.each do |commit|
         # Diabled rubocop because the suggested .empty? does not work here.
-        # rubocop:disable Style/ZeroLengthPredicate
-        next if commit.diff(paths: [path]).size.zero?
+        next if commit.diff(paths: [path]).size.zero? # rubocop:disable Style/ZeroLengthPredicate
 
-        # rubocop:enable Style/ZeroLengthPredicate
         hash = {}
         hash[:date] = commit.time.to_s
         hash[:oid] = commit.oid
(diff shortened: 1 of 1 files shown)


mikegleasonjr:
Fized for me too

aschaber1:
I rewrote this into an inline disablement, as it's used more commonly elsewhere in the repo instead of the block, when just a single line is marked.

Thank you for your help :)

PR #2780, opened 2023-05-05
Update firewareos.rb

### LINKED ISSUE #2779: Prompt for Watchguard FirewareOS not matching the Regex when the node is non-master
This is related to issue https://github.com/ytti/oxidized/pull/1063

Any non master node in a Fireware cluster has a prompt containing the \<non-master\> text.
E.g my non master node has this prompt in this moment:

[FAULT]WG\<non-master\>>

The configured Regex in firewareos.rb:
`^\[?\w*\]?\w*?(<\w*>)?(#|>)\s*$`
cannot capture this prompt because of the dash into the \<non-master\> section.

Changing the Regex to

`^\[?\w*\]?\w*?(<[\w-]*>)?(#|>)\s*$`

seems to solve the issue.

I put this regex into a modified firewareos.rb file in the config/model directory to override the default one, and it works for the master and non-master node of the cluster.

Regards
Lorenzo

netdiver:
Changed regex for prompt to accept the dash character and match with the <non-master> nodes of a fireware cluster. Fixes #2779

## Description

Changed regex for prompt to accept the dash character and match with the <non-master> nodes of a fireware cluster. Fixes #2779

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ## Fixed
 
 - fixed empty lines for ZyXEL GS1900 switches (@jluebbe)
+- fixed prompt for Watchguard FirewareOS not matching the regex when the node is non-master (@netdiver)
 
 ## [0.29.1 - 2023-04-24]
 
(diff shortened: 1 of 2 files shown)


mortzu:
please update CHANGELOG.md

Issue #2803, opened 2023-05-23
Git output broken since 0.29.1


setaou:
When upgrading to oxidized 0.29.1, the daemon crashes at start with the following error :

```
undefined method `name' for "set version 20.4R3-S3.4\nset [...] ":String [NoMethodError]
--------------------------------------------------
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/output/git.rb:47:in `block (2 levels) in store'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/output/git.rb:46:in `each' 
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/output/git.rb:46:in `block in store'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/output/git.rb:43:in `each'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/output/git.rb:43:in `store'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/worker.rb:65:in `process_success'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/worker.rb:47:in `process'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/worker.rb:16:in `block in work'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/worker.rb:16:in `each'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/worker.rb:16:in `work'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/core.rb:36:in `run'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/core.rb:29:in `initialize'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/core.rb:4:in `new'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/core.rb:4:in `new'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/lib/oxidized/cli.rb:13:in `run'
/var/lib/gems/3.0.0/gems/oxidized-0.29.1/bin/oxidized:8:in `<top (required)>'
/usr/local/bin/oxidized:25:in `load'
[...]

PR #2804, opened 2023-05-23
Add missing "using Refinements" (closes issue #2803)

setaou:
Add missing "using Refinements" in lib/oxidized/output/git.rb, which was making the git output unusable since version 0.29.1.

Closes issue #2803


setaou:
On a side note, 0.29.1 should really have been tagged 0.30, as the need to add "using Refinements" everywhere is abruptly breaking compatibility with user created models: this was clearly not just a fix release !

PR #2828, opened 2023-06-13
Fix Netscaler with hostname backups

electrocret:
Currently Netscalers with "set ns hostName _name_" set aren't able to be backed up cause Oxidized doesn't recognize a valid prompt.

Prompt when not set:
`>`

Prompt when set:
`username@hostname>`

Current regex isn't liking the change.

I'm sure there's a more graceful way of dealing with the change, but I'm not good with regex. This works for me.

## Pre-Request Checklist


mortzu:
can you please fix the conflict (sorry)

electrocret:
Done

PR #2859, opened 2023-07-07
using slack-ruby-client api

pubyun:
https://github.com/slack-ruby/slack-ruby-client/tree/master/examples


mortzu:
why the change?

PR #2889, opened 2023-07-28
allow finding nodes by IP

Habbie:
## Pre-Request Checklist

- [x] Passes rubocop code analysis (try `rubocop --auto-correct`) (it has complaints, but not about this file)

## Description

Since b0b50e8bba592bf9c96c198087dc19b077044f51, most operations allow either node name or node IP as index. This is reflected in what oxidized-web accepts in URLs as well. However, for `/node/version` (and perhaps other endpoints), by-IP did not work yet because the change was made in one place instead of two. This PR makes the other change.

My use case: in LibreNMS, my switches are identified by IP instead of a host name. This means that in the LibreNMS/Oxidized integration, I can view details about the switches, but not versions, because that API endpoint did not take IPs.


Habbie:
I left the first two commits separate in case you decide the first attempt looks cleaner to you. Can squash any time.

yeled (approved):

Habbie:
rebased to fix conflict in `CHANGELOG.md`

PR #2796, opened 2023-05-15
tp-link: fixed enable mode port login entrance

mirackle-spb:
## Pre-Request Checklist

- [x] Passes rubocop code analysis (try `rubocop --auto-correct`)
- [x] Tests added or adapted (try `rake test`)
- [x] Changes are reflected in the documentation
- [x] User-visible changes appended to [CHANGELOG.md](/CHANGELOG.md)

## Description
tp-link: fixed enable mode port login entrance

PR #2912, opened 2023-08-18
Fix enable with password on TPLink

lukas-hetzenecker:
## Pre-Request Checklist

## Description
Follow-Up to #2796 

This PR seems to have introduced a bug when enable is used WITH a password
In that case "enable" still needs to be sent before the actual password.

Current debug log when connecting:

```
switch>**sensitive enable password***
---------------^
Error: Bad command

switch>

switch>show system-info
---------------^
Error: Bad command

```

### DIFF
--- a/lib/oxidized/model/tplink.rb
+++ b/lib/oxidized/model/tplink.rb
@@ -58,6 +58,7 @@ class TPLink < Oxidized::Model
       if vars(:enable) == true
         cmd "enable"
       elsif vars(:enable)
+        send "enable\r"
         cmd vars(:enable)
       end
     end


ytti:
Likely almost never `send` is correct, because then we don't wait for anything, we don't know if far end is ready to receive what ever is next.

`cmd` is special case of send which waits for something, if not specified prompt. 

`cmd`does take second parameter, which can be used to wait for something else than prompt. For example enable prompt. That is, you likely want something like `cmd "enable", /^enablepromptseen/`

lukas-hetzenecker:
Strangely here `send "enable\r"` seems to be the only thing that actually works, at least for my TL-SG2008 1.0.

Any combination of `cmd "enable"`, `cmd "enable", /^\r?Password:$/`, `cmd "enable", /[pP]assword:/`, etc. also results in a:


```
switch>enable

Password:
Error password!


switch>sensitivepassword
---------------^
Error: Bad command
```

whereas with `send` the result and log file is as expected:

```
switch1>enable

Password:

switch1#

switch1#show system-info

 System Description   - 8-Port Gigabit Smart Switch
[...]
```

The regular expression should match the configured patterns, so I don't know what else to try (except reverting to send as it also was before #2796)

PR #2915, opened 2023-08-22
Update iosxr.rb with admin keyword for show inventory command

maugli13:
## Pre-Request Checklist

- [] Passes rubocop code analysis (try `rubocop --auto-correct`)
The core code didn't change. Only specific model
The core code didn't change. Only specific model
Particular Model didn't have the documentation
N/A

## Description
"show inventory" command without admin keyword is not showing some useful inventory information, such as FAN, PSU and Chassis serial and part numbers

### DIFF
--- a/lib/oxidized/model/iosxr.rb
+++ b/lib/oxidized/model/iosxr.rb
@@ -16,7 +16,7 @@ class IOSXR < Oxidized::Model
     cfg
   end
 
-  cmd 'show inventory' do |cfg|
+  cmd 'admin show inventory' do |cfg|
     comment cfg
   end
 


aschaber1:
Does it show nothing or just less information?
Does `admin` need elevated privileges?

maugli13:
@aschaber1 without admin command "show inventory" providing less information. 
For example, FAN, PSUs, and Chassis related data are absent.

Yes, admin mode requires root-system group access privileges; commands executed by the elevated users can be limited to RO by AAA or any other options.

robertcheramy:
As "show running-config" already needs elevated privileges, I think "admin show inventory" should be OK.

Issue #1802, opened 2019-05-07
Backup APC config using SCP?


danpoltawski:
We've got some APC devices, but don't have FTP access to them. However scp of config.ini seems to work.

Is it possible using the oxidized framework to use scp to simply download a config file, similar to the way the current FTP implement seems to work?

ytti:
Do they support SFTP? I'm happy to implement either, but OpenSSH people urge people to stop using scp and migrate to sftp. 

If you can't do it, I can don it in a week or so, I'm currently traveling. If you want me to do it, setup sanitised device, make it reachable from 91.198.120.1/ytti.fi and send me credentials via saku@ytti.fi

danpoltawski:
Alas, it doesn't seem to support SFTP. Out of interest, is there a url about the openssh thing? It's the first i've heard of it.

ytti:
It is regularly mentioned in openssh mailing list, it's also mentioned in release notes. I can find specific urls if that is important. 

No worries, I'll be happy to implement scp support.

cppmonkey:
https://www.openssh.com/releasenotes.html

> Security
> ========
> 
> This release contains mitigation for a weakness in the scp(1) tool
> and protocol (CVE-2019-6111): when copying files from a remote system
> to a local directory, scp(1) did not verify that the filenames that
> the server sent matched those requested by the client. This could
> allow a hostile server to create or clobber unexpected local files
> with attacker-controlled content.
> 
> This release adds client-side checking that the filenames sent from
> the server match the command-line request,
> 
> The scp protocol is outdated, inflexible and not readily fixed. We
> recommend the use of more modern protocols like sftp and rsync for
> file transfer instead.

Well, that is unfortunate. I've just started looking for alternative to FTP, as most of our APC UPS devices have FTP disabled.
Also looking to build a module for Ansible-Networking too

ytti:
I am still happy to support SCP, so if you provide access to device, I'll create model for it using SCP. I mean we support telnet too, which is clearly lot worse option.

danpoltawski:
Hoping I might be able to provide access to a device, but it's going to require some internal work, stay tuned and thanks!

nwhistler:
@ytti I can provide you a apc pdu if you are still willing to create a model for it.

ytti:
Please do, I'd come from 91.198.120.1.

I'll probably call the module sftp anyhow and have config block to change it to scp for legacy devices. We'll see.

brobare:
[...]

PR #2919, opened 2023-08-23
add SCP input script

aeiodelic:
## Pre-Request Checklist

## Description
Needed for some, mainly old, devices that only support backup over SCP.

Solves [issue 1802](https://github.com/ytti/oxidized/issues/1802) that was already closed due to not planned.


aschaber1:
This looks like an interesting extension. I'm going to reopen this :) 

Hopefully maybe one of my co-maintainers with more ruby experience have an opinion on this.

ytti:
I think it is fine basically regardless of its condition, because it doesn't affect anything existing, so there is no real other pressure than 'does it help someone', which is satisfied.

Issue #2897, opened 2023-08-02
"source returns no usable nodes" error after docker host reboot


n1c0x:
Hello everyone, 
I know, this error appears a lot of times here, but I couldn't find any solution. I read my config files multiple times to see if there is a typo somewhere, but didn't find anything. 

I installed oxidized and LibreNMS a few weeks ago on the same server, but each in its own docker container (multiple docker container for libreNMS). All the container are in the same docker network, which is a bridge. 

Everything worked fine until I restarted my host server. Since then, Oxidized is unable to start, with the following error : 

```
I, [2023-08-02T09:12:01.964453 #2635]  INFO -- : Oxidized starting, running as pid 2635
I, [2023-08-02T09:12:02.020787 #2635]  INFO -- : lib/oxidized/nodes.rb: Loading nodes
I, [2023-08-02T09:12:02.043168 #2635]  INFO -- : lib/oxidized/nodes.rb: Loaded 0 nodes
F, [2023-08-02T09:12:02.043231 #2635] FATAL -- : Oxidized crashed, crashfile written in /home/oxidized/.config/oxidized/crash
source returns no usable nodes
```

Here is my oxidized config file : 

```
---
rest: 0.0.0.0:8888
username: top_secret_username
password: top_secret_password
interval: 3600
resolve_dns: true
use_syslog: false
threads: 30
use_max_threads: false
timeout: 20
retries: 3
prompt: !ruby/regexp /^([\w.@-]+[#>]\s?)$/
next_adds_job: true
vars: {}
groups: {}
group_map: {}
models: {}
pid: "/home/oxidized/.config/oxidized/pid"
crash:
  directory: "/home/oxidized/.config/oxidized/crashes"
  hostnames: false
stats:
  history_size: 10
input:
  default: ssh
  debug: false
  ssh:
    secure: false
  utf8_encoded: true
source:
  default: http
  debug: false
  http:
    url: http://docker_host:8001/api/v0/oxidized
    map:
      name: hostname
      model: os
      group: group
    headers:
      X-Auth-Token: 'token'
  csv:
    file: "/home/oxidized/.config/oxidized/router.db"
    delimiter: !ruby/regexp /:/
    map:
      name: 0
      model: 1
    gpg: false
output:
  default: git
  git:
    user: oxidized
    email: o@example.com
    repo: "/home/oxidized/.config/oxidized/git-repos/default.git"
  file:
    directory: "/home/oxidized/.config/oxidized/configs"
```

[...]

PR #2923, opened 2023-08-29
Bug Fix: source http

davama:
Fixes: #2897


ZummiGummi:
Confirmed that these changes fixed the integration problem with LibreNMS as reported here, https://github.com/ytti/oxidized/issues/2897

electrocret:
+1 Fix works

Issue #2021, opened 2020-01-23
FortiGate backups fail - login banner "Press 'a' to accept"


chrisr0880:
G'day,

First time playing with oxidized (and pretty new to this open source world) - brilliant work!

I ran into a minor issue where FortiGate firewalls would not backup when the post login banner is configured `set post-login-banner enable`.  This seems to be due to the additional prompt `(Press 'a' to accept):` when the feature is enabled.

I modified the [fortios.rb](https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/fortios.rb) script with the following snipit and initial testing appears to have resolved the issue for me.

```
  prompt /^([-\w.~]+(\s[(\w\-.)]+)?~?\s?[#>$]\s?(\(Press\s\'a\'\sto\saccept\)\:)?)$/

  expect /\(Press\s\'a\'\sto\saccept\)\:/ do |data, re|
    send 'a'
    data.sub re, ''
  end
```

DarkCatapulter:
This looks like something nice to have if others experience the same issue. Do you know if this new code works even if you were to disable the post-login banner? Because in that case, this could definitely be added to code as a Pull Request, always nice to fill in the blanks for future users :)

wk:
This looks like a valuable improvement. It's expected that modifying the prompt in this manner should be unnecessary, as the `expect` is meant to trigger even before the first prompt is seen. If this isn't happening here, it'll be useful to figure out why.

Is the `Press 'a'` banner ever displayed on the same line as the prompt, or is it always on its own line? In the latter case, it might be best to constrain the expect to match the beginning of a line (`^`).

ryanmerolle:
@chrisr0880 this feels like a simple PR.  Do you want to submit this?  If not I can.

chrisr0880:
> This looks like something nice to have if others experience the same issue. Do you know if this new code works even if you were to disable the post-login banner? Because in that case, this could definitely be added to code as a Pull Request, always nice to fill in the blanks for future users :)

Unfortunately I don't have access to the environment any more to double check - I believe I tested this… but would double check first

chrisr0880:
> This looks like a valuable improvement. It's expected that modifying the prompt in this manner should be unnecessary, as the `expect` is meant to trigger even before the first prompt is seen. If this isn't happening here, it'll be useful to figure out why.
> 
[...]

PR #2934, opened 2023-09-10
Fixing the stale issue 2021. 

jforeman:
This corrects the behavior when a post login banner is enabled.

## Description
Instead of sending a space when --More-- is presented, an "a" key is sent when post login banner is enabled, which covers both cases.

Closes issue #2021.


aschaber1:
How does this work if `--More--` is displayed?

D1gnit:
I have one Fortigate with post login banner and one without, and this works on both, if that answers your question.

robertcheramy:
@jforeman - could you document the change into CHANGELOG.md (##Fixed) ?

robertcheramy:
Thanks for your PR! This Issue has already been fixed in PR #2996.

PR #2954, opened 2023-09-29
Update the prompt detection RegEx to support Netgear M4250-10G2XF-PoE+ and M4300-28G-PoE+

rexhaugen:
## Description

The pull request is to update the prompt detection RegEx to support Netgear M4250-10G2XF-PoE+ and M4300-28G-PoE+.


mortzu:
Please update CHANGELOG.md

PR #2969, opened 2023-10-19
Extend zynos.rb content

sharteeya:
## Pre-Request Checklist

## Description

Extend `zynos.rb` content since it does not have an update for a while.
New contents have tested on the following ZyNOS version:
- V4.10 (2014 - GS2210)
- V4.70 (2021 - GS2220)


aschaber1:
Hello @sharteeya , please add a line to changelog.

sharteeya:
@aschaber1 Changes added into changelog. Thanks.

robertcheramy:
@sharteeya - please correct the unsuccessful checks

robertcheramy:
@sharteeya I dismissed the alerts, as they are false positive.
I see that you commented `cmd 'config-0'` out. I have never seen a zynos in my life, so my understanding might be wrong:
- you changed the model so that the config could be read by ssh / telnet
- the `cmd 'config-0'` would be for ftp (see docs/Model-Notes/XGS4600-Zyxel.md) 
Do we introduce a regression here?

sharteeya:
@robertcheramy Thanks for helping.
- That's right. The config could be read via ssh and telnet with those changes.
- Unfortunately, I'm not quite understand about XGS4600 series and lack of information about it. Shall we keep those part of TFTP with comment and leave some info? Then users who need TFTP, they can just remove `#` to enable TFTP config backup.

robertcheramy:
So you are renewing the whole zynos model so it will work with ssh/telnet only. It's okay, and should be documented:
- in CHANGELOG.md (backup with ftp won't work anymore - open an issue if you still need it )
- in docs/Model-Notes/XGS4600-Zyxel.md (uncoment the ftp lines in zynos.rb to use ftp)
- in zynos.rb ("uncomment folowing lines to use ftp", and `cfg :ftp do // end` should also be commented out)

Could you modify the PR with this points?

robertcheramy:
Thanks!

PR #3060, opened 2024-02-09
Prevent potential busy loop in scheduler from jobs > nodes

gs-kamnas:
## Pre-Request Checklist

## Description
This change remediates a potential busy-loop in the scheduler which results from either next_adds_job allowing the job count to exceed the node count or a retry after configuration fetch failure causing the same invalid state.

### DIFF
--- a/lib/oxidized/jobs.rb
+++ b/lib/oxidized/jobs.rb
@@ -44,6 +44,16 @@ module Oxidized
       @want = @max if @want > @max
     end
 
+    def increment
+      # Increments the job count if safe to do so, which means:
+      # a) less threads running than the total amount of nodes
+      # b) we want less than the max specified number of threads
+
+      return unless (@want < @nodes.size) && (@want < @max)
+
+      @want += 1
+    end
+
     def work
       # if   a) we want less or same amount of threads as we now running
       # and  b) we want less threads running than the total amount of nodes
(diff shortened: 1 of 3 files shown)


lib/oxidized/jobs.rb
```diff
@@ -44,14 +44,26 @@ def new_count
       @want = @max if @want > @max
     end
 
+    def increment
+      # Increments the job count if safe to do so, which means:
+      # a) less threads running than the total amount of nodes
+      # b) we want less than the max specified number of threads
+
+      return unless (@want < @nodes.size) && (@want < @max)
```
ytti, line 52: What do you think of @want < [@nodes.size, @max].min?

ytti: Or [@want, @nodes.size, @max].min == @want

gs-kamnas: Can actually summarize the whole function into:

`want = [(@want + 1), @nodes.size, @max].min`

ytti: If you think it's best. I didn't suggest the option to reduce code size, I suggested it because it takes longer for me to parse and understad condition1 && condition2, than [a,b,c].min == a, but this is just personal opion.

I'm happy to accept the PR as is, but if you want to change, that's also fine.

gs-kamnas: @ytti I could go either way (separate conditional vs. inline with the assignment); your call here.

Both are functional equivalent and I don't find one easier/harder to interpret personally.

gs-kamnas:
@ytti caught a mistake in this; should have assigned to `@want`. Followup PR incoming.

PR #3075, opened 2024-02-19
Revert "Add support for Fibrestore FSOS"

systeembeheerder:
Reverts ytti/oxidized#3069

docs/Supported-OS-Types.md

There already is a Fiberstore Vendor in the list. Now you add another one between Cisco and Citrix.


tcrichton:
Rather than remove the support, let's sort the docs since this is a different model.

PR #3100, opened 2024-03-08
Initial implementation of the Fujistu 1Finity

jerji:
## Pre-Request Checklist

## Description
I added basic support for the Fujitsu 1finity switches. They ship without telnet enabled so I did not implement it.

### DIFF
--- a/lib/oxidized/model/onefinity.rb
+++ b/lib/oxidized/model/onefinity.rb
@@ -1,0 +1,19 @@
+class OneFinity < Oxidized::Model
+  using Refinements
+
+  # Fujitsu 1finity
+
+  prompt /(\r?[\w.@_()-]+[>]\s?)$/
+
+  cmd :all do |cfg|
+    cfg.each_line.to_a[1..-3].join
+  end
+
+  cmd 'show configuration | display set | nomore' do |cfg|
+    cfg
+  end
+
+  cfg :ssh do
+    pre_logout 'exit'
+  end
+end
--- a/lib/oxidized/model/onefinity.rb
+++ b/lib/oxidized/model/onefinity.rb
@@ -13,5 +13,6 @@ class OneFinity < Oxidized::Model
 
   cfg :ssh do
     pre_logout 'exit'
+    exec true
   end
 end
--- a/lib/oxidized/model/onefinity.rb
+++ b/lib/oxidized/model/onefinity.rb
@@ -9,9 +9,7 @@ class OneFinity < Oxidized::Model
     cfg.each_line.to_a[1..-3].join
   end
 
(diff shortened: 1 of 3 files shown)


lib/oxidized/model/onefinity.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  # Fujitsu 1finity
+
+  prompt /(\r?[\w.@_()-]+[>]\s?)$/
+
+  cmd :all do |cfg|
+    cfg.each_line.to_a[1..-3].join
+  end
+
+  cmd 'show configuration | display set | nomore' do |cfg|
```
robertcheramy, line 12: Could you try `cmd 'show configuration | display set | nomore'`(without do `|cfg| ... end`). If you do not need to manipulate the output, there is no need to add a block here.

lib/oxidized/model/onefinity.rb
```diff
@@ -0,0 +1,19 @@
...
+    cfg
+  end
+
+  cfg :ssh do
+    pre_logout 'exit'
```
robertcheramy, line 17: Have you tried if `exec true` works for the model?
As all commands are independent, it would make the model independent from the prompt.
Note: `exec true` needs the ssh-server to support exec channels.

robertcheramy (changes requested):
Thank you vor your PR.
Please add the model to docs/Supported-OS-Types.md

When you are at it, please test the two comments in the code. They are not mandatory, they would make your model code nicer if they works.

lib/oxidized/model/onefinity.rb
jerji: Thank you, I saw this in other models and assumed it was necessary, it's been fixed

lib/oxidized/model/onefinity.rb
robertcheramy: Lines 14-16 would look like:
```ruby
cfg :ssh do
  pre_logout 'exit'
  exec true
end
```

jerji: I added the line :)

I'm not sure I understand what it does because it's not in the docs but it didn't seem to break anything 👍

jerji:
Do you have a planned release date for the next version? Just trying to guess when my organization can update ^-^

robertcheramy:
> Do you have a planned release date for the next version? Just trying to guess when my organization can update ^-^

I've still an open issue in [Milestone 0.30](https://github.com/ytti/oxidized/milestone/4), I think I will be able to fix it in the next days, so I plan to release this month.

jerji:
@robertcheramy do you think this is done to your satisfaction?

robertcheramy (approved):

Issue #3033, opened 2024-01-15
I want to use SQL as my data source. SQL is configured in my configuration file, but an error occurs after startup.


Evanfmm:
```yaml
source:
  default: sql
  debug: true
  sql:
    adapter: mysql2
    host: 10.130.110.32
    user: xxxx
    password: xxxxxxx
    database: xx
    port: 33061
    table: app_net_dev
    map:
      name: ip
      model: model
      username: username
      password: password
      group: group
      port: port
    vars_map:
      enable: enable
```

error:
```bash
nohup: ignoring input
cannot interpret as DNS name: nil
/usr/local/lib/ruby/3.1.0/resolv.rb:1269:in `create': cannot interpret as DNS name: nil (ArgumentError)
	from /usr/local/lib/ruby/3.1.0/resolv.rb:1106:in `generate_candidates'
	from /usr/local/lib/ruby/3.1.0/resolv.rb:1135:in `resolv'
	from /usr/local/lib/ruby/3.1.0/resolv.rb:530:in `fetch_resource'
	from /usr/local/lib/ruby/3.1.0/resolv.rb:516:in `each_resource'
	from /usr/local/lib/ruby/3.1.0/resolv.rb:406:in `each_address'
	from /usr/local/lib/ruby/3.1.0/resolv.rb:118:in `block in each_address'
	from /usr/local/lib/ruby/3.1.0/resolv.rb:117:in `each'
	from /usr/local/lib/ruby/3.1.0/resolv.rb:117:in `each_address'
	from /usr/local/lib/ruby/3.1.0/resolv.rb:95:in `getaddress'
	from /usr/local/lib/ruby/gems/3.1.0/gems/oxidized-0.29.1/lib/oxidized/node.rb:20:in `initialize'
	from /usr/local/lib/ruby/gems/3.1.0/gems/oxidized-0.29.1/lib/oxidized/nodes.rb:22:in `new'
	from /usr/local/lib/ruby/gems/3.1.0/gems/oxidized-0.29.1/lib/oxidized/nodes.rb:22:in `block (2 levels) in load'
	from /usr/local/lib/ruby/gems/3.1.0/gems/oxidized-0.29.1/lib/oxidized/nodes.rb:17:in `each'
	from /usr/local/lib/ruby/gems/3.1.0/gems/oxidized-0.29.1/lib/oxidized/nodes.rb:17:in `block in load'
[...]

PR #3107, opened 2024-03-15
Allow connecting to a non standard port for sql source

abashmakov:
## Description
Allow connecting to a non standard port for sql source. Trivial change.

Some users seem to using non standard ports like in issue #3033 (different error, but once it's fixed, the user will probably open another issue)


robertcheramy (changes requested):
Thank you for the improvement.
Please document it in CHANGELOG.md
As we released a new version of Oxidized, you should merge master in your branch first.

PR #3198, opened 2024-06-18
Add JSON source class.

sargon:
With this class we can utilise JSON files directly without the redirection via a HTTP server. This massivly improved our configuration delivery process, because we are able to deliver cached configuration to oxidized without changes on the delivering service, which would have been much more complicated.

## Pre-Request Checklist

## Description
Adds an additional source class which is able to process JSON encoded files, like the HTTP source does.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - dlink: added support for 'enable admin' before getting configuration, if enable=true (@as8net)
 - dlinknextgen: strip uptime and ntp update time from config
 - Updated slackdiff.rb to use slack_ruby_client instead of slack-api (@Punicaa)
+- Add JSONFILE source (@sargon)
 
 ### Fixed
 - fixed prompt for vyos/vyatta to allow logins with non-priviliged accounts. Fixes #3111 (@h-lopez)
--- a/docs/Sources.md
+++ b/docs/Sources.md
@@ -55,6 +55,24 @@ and within: `~/.gnupg/gpg.conf`
 pinentry-mode loopback
 ```
 
+## Source: JSONFile
+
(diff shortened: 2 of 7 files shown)


ytti:
Do we want to have separate string_navigate methods for HTTP object  and JSON file object? I feel like in absense of having good reason to have duplicate methods, we should only have one method for maintainability and least amount of surprises.

ytti:
You could add it to 'source.rb' as all sources inherit from it, so then it'll be available in any source.

I think it's generic enough, because it has nothing to do technically with JSON, it's more navigating any ruby hash like object.

sargon:
> I think it's generic enough, because it has nothing to do technically with JSON, it's more navigating any ruby hash like object.

There is a section method (CSV.open_file) that is utilized, I guess the same argument can be applied.

ytti:
> > I think it's generic enough, because it has nothing to do technically with JSON, it's more navigating any ruby hash like object.
> 
> There is a section method (CSV.open_file) that is utilized, I guess the same argument can be applied.

i'm not sure I understand that. I mean moving `string_navigate` from http.rb, to source.rb, should make it usable in all sources. 
I'm also not against changing the name of the method, if something more describe pop-ups, like string_navigate_object or whatever.

sargon:
Yes, we are on the same page with the string_navigate method.

The jsonfile source class had a second method duplicated from the csv source class open_file.
Should I proceed with this function in the same way?

ytti:
> Yes, we are on the same page with the string_navigate method.
> 
> The jsonfile source class had a second method duplicated from the csv source class open_file. Should I proceed with this function in the same way?

Yeah, that's fine by me. When moving methods to source.rb, there may be need to reflect little bit, if the method name should remain same or if more clarity is needed on the name. As contextual documentation is removed, the names may become too generic. But I'm ok with or without renaming.

ytti:
Looks good to me.

PR #3224, opened 2024-07-12
Add RouterOS maintainer

systeembeheerder:
## Pre-Request Checklist

- [ ? ] Passes rubocop code analysis (try `rubocop --auto-correct`)
- [ N ] Tests added or adapted (try `rake test`)
- [ Y ] Changes are reflected in the documentation
- [ N ] User-visible changes appended to [CHANGELOG.md](/CHANGELOG.md)

## Description
Add myself as Mikrotik RouterOS model maintainer.

PR #3252, opened 2024-09-02
Revert "Add RouterOS maintainer"

systeembeheerder:
Reverts ytti/oxidized#3224

See #3250 if this needed.


robertcheramy:
I regret your decision removing yourself as a model maintainer. I was very happy that you chosed to become a model maintainer. I would be very happy if you would reconsider your decision, although I understand if you choose to keep it that way.

What I cannot accept is that you want to remove the model notes for RouterOS. Please change the PR, so that the model notes are not removed by it.

PR #3273, opened 2024-09-18
ciscosmb: fix backup when no enable password is set

### LINKED ISSUE #3044: Cisco SmallBusiness is not working
Tested on SB250+SB350 and various CBS Switches.

I hat to outcoment:

  
      if vars(:enable) == true
        cmd 'enable'
      elsif vars(:enable)
        cmd 'enable', /^\r?Password:$/
        cmd vars(:enable)
      end

Those switches haven't an enable comment...

benasse:
## Pre-Request Checklist

## Description

ciscosmb: fix backup when no enable password is set. Fixes #3044

Closes issue #3044

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - fixed digest authentication when using http input (@spike77453)
 - fixed aosw prompt; now working with ArubaOS 8 (@mabezi, @robertcheramy)
 - routeros: fix system info for CHR. Fixes #3180 (@systeembeheerder)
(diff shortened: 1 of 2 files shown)


ytti:
This will not work. 

Problem is, people have enable prompts, with no password, which expect empty password or new line.

Today we do this: https://github.com/ytti/oxidized/blob/master/lib/oxidized/source/source.rb#L18-L25

That is, if enable password string is 'nil' or 'false', then we turn it  into boolean, and the model will consider no enable password was given.

There is currently no solution to pass verbatim 'nil' or 'false' enable passrod, but those are expected to be much more rare than empty enable password, which is fairly common.

benasse:
ok thanks for the explanation.
I confirm that setting enable to false solves my problem !

Issue #2154, opened 2020-07-15
oxidized superclass mismatch for class http


typecookie:
When using http as input and connected to librenms (using http as source)

possibly i am just dumb

`input:
  default: ssh, telnet, http
  debug: false
  ssh:
    secure: false
  ftp:
    passive: true
  utf8_encoded: true`

`source:
        default: http
        debug: false
        http:
          url: https://librenms/api/v0/oxidized
          map:
            name: hostname
            model: os
            group: group
          headers:
            X-Auth-Token: '01582bf94c03104ecb7953dsadsadwed'`

typecookie:
any help on this Im willing to do some testing just need pointed in the right direction

ytti:
Where ever this is happening, we need to change reference there to fully qualified class path or we must at least include the Input::X Source::X parent class to disambiguate.

typecookie:
i think it is happening at least for the source in lib/oxidized/nodes.rb  and possibly rewriting the "@source" definition.
Im not sure how to redefine using the Qualified class path Ive tried putting in a couple of variants in a couple of places but i just keep breaking it.

Im a script monkey  and all my code is haky

ytti:
If you have traceback, that would help. Somewhere we say Http, we should say Source::Http or Input::Http to disambiguate.

typecookie:
its not traceback but i have the error log

`superclass mismatch for class HTTP [TypeError]
--------------------------------------------------
/var/lib/gems/2.5.0/gems/oxidized-0.28.0/lib/oxidized/input/http.rb:6:in `<module:Oxidized>'
/var/lib/gems/2.5.0/gems/oxidized-0.28.0/lib/oxidized/input/http.rb:1:in `<top (required)>'
/usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/var/lib/gems/2.5.0/gems/oxidized-0.28.0/lib/oxidized/manager.rb:9:in `load'
/var/lib/gems/2.5.0/gems/oxidized-0.28.0/lib/oxidized/manager.rb:60:in `loader'
/var/lib/gems/2.5.0/gems/oxidized-0.28.0/lib/oxidized/manager.rb:35:in `add_input'
/var/lib/gems/2.5.0/gems/oxidized-0.28.0/lib/oxidized/node.rb:155:in `block in resolve_input'
[...]

PR #3283, opened 2024-10-09
Updated source files to avoid namespace duplication

laf:
## Pre-Request Checklist

## Description
This possibly fixes issue https://github.com/ytti/oxidized/issues/2154. I've got a limited test environment I've run this in and without changing the config file this works ok for me.

manager.rb may need more doing to it but this worked for me.


laf:
@robertcheramy Would you be able to review this change?

robertcheramy:
Yes, I will look at your PR but must work on other issues and older PR first.

michieltimmers:
FYI. I tested the pull request and it fixed the problem I had. I can now have both Librenms has http source, and also have input with http.

@laf Thanks!!!

lib/oxidized/manager.rb
```diff
@@ -8,7 +8,7 @@ class << self
       def load(dir, file)
         require File.join dir, file + '.rb'
         klass = nil
-        [Oxidized, Object].each do |mod|
+        [Oxidized, Source, Object].each do |mod|
```
robertcheramy, line 11: It helped me understanding the code better when using the full module path (`Oxidized::Source`) even if `Source` is  correct. What do you think?

Could you add a comment describing what this is doing, something like:
```
# Search the object in the different namespaces
# - Most objects are in the namespace Oxidized
# - Source objects have their own namespace (Oxidized::Source)
# - Models have no namespace (Object)
```

robertcheramy (changes requested):
1. Could you add a unit test for your change in the manager class under /spec/manager_spec.rb. I suggest:
```
require_relative 'spec_helper'

describe Oxidized::Manager do
  before(:each) do
    Oxidized.asetus = Asetus.new
    Oxidized.asetus.cfg.debug = false
    Oxidized.setup_logger
  end

  describe '#add_source' do
    it 'loads a source when available' do
      Oxidized.config.source.csv.file = '/some/path'
      result = Oxidized.mgr.add_source('csv')
      _(result['csv']).must_equal Oxidized::Source::CSV
    end

    it 'returns nil when the source is not available' do
      result = Oxidized.mgr.add_source('XXX')
      _(result).must_be_nil
    end
  end
end
```
I would have included it into your PR, bud you didn't give the oxidized maintainer the right to modify your PR.

2. Please document your change in CHANGELOG.md

laf:
> 1. Could you add a unit test for your change in the manager class under /spec/manager_spec.rb. I suggest:
> 
> ```
> require_relative 'spec_helper'
> 
> describe Oxidized::Manager do
>   before(:each) do
>     Oxidized.asetus = Asetus.new
>     Oxidized.asetus.cfg.debug = false
>     Oxidized.setup_logger
>   end
> 
>   describe '#add_source' do
>     it 'loads a source when available' do
>       Oxidized.config.source.csv.file = '/some/path'
>       result = Oxidized.mgr.add_source('csv')
>       _(result['csv']).must_equal Oxidized::Source::CSV
>     end
> 
>     it 'returns nil when the source is not available' do
>       result = Oxidized.mgr.add_source('XXX')
>       _(result).must_be_nil
>     end
>   end
> end
> ```
> 
> I would have included it into your PR, bud you didn't give the oxidized maintainer the right to modify your PR.
> 
> 2. Please document your change in CHANGELOG.md

Sorry for the delay on this one @robertcheramy. I've updated the CHANGELOG and also added the test file in with a small modification. I appreciate the feedback.

lib/oxidized/manager.rb
laf: I'm happy to take direction on using the full module path, I've updated the PR to use it and added your comment in.

robertcheramy:
Thanks for the corrections. I've found two new problems for the PR:
- It relies on the order in which add_source and add_input are called.
- It does not resolve the conflict with the Output http

I'm currently fixing these problems and will then merge into master. You do not need to modify this PR.

robertcheramy:
Manualy rebased/squashed/merged into master: 73c8861204c7536276336ddda5eda62398217634 and b3f3f9542d5196e044871a2d24753c1c23b2eb1a
Closes: #2154

laf:
> Thanks for the corrections. I've found two new problems for the PR:
> 
> * It relies on the order in which add_source and add_input are called.
> * It does not resolve the conflict with the Output http
> 
> I'm currently fixing these problems and will then merge into master. You do not need to modify this PR.

Thanks @robertcheramy. I appreciate the feedback and help on this. Glad it got sorted in the end :)

PR #3284, opened 2024-10-10
no error, if the default config is created

twille007:
## Pre-Request Checklist

## Description
If you use saltstack for example, on init oxidized, you will get an error, because it writes to _stderr_, if there is no configfile and if there is none, it creates one. It just tries to create the file, and if it doesn't work, it will write to _stderr_. Otherwise it will inform you about the new config file via _stdout_, and that you should edit it.

### DIFF
--- a/lib/oxidized/config.rb
+++ b/lib/oxidized/config.rb
@@ -69,8 +69,10 @@ module Oxidized
 
       if asetus.create
         puts "Configuration file created at #{@configfile}. Please edit it."
+      elsif !File.exist?(@configfile)
+        raise NoConfig, "Configuration file #{@configfile} is missing and could not be created."
       else
-        raise NoConfig, "Configuration file #{@configfile} is missing."
+        puts "Configuration file #{@configfile} already exists."
       end
 
       # override if comand line flag given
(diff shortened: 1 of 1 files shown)


robertcheramy:
The whole Oxidized is based on using Oxidized.logger, defined in https://github.com/ytti/oxidized/blob/41f59ac25d4b7dec79216d79b90bd64899d68e39/lib/oxidized.rb#L39

You should use it and not puts.

Beside this, without knowing details about saltstack, I would think that someone deploying oxidized with saltstack knows that he/she also has to deploy a configuration with it.

twille007:
Ok, i changed it now to the logger.

And I completely agree that anyone deploying Oxidized with SaltStack should be aware that a configuration file needs to be deployed alongside it. However, I think it’s still a good practice not to raise an error if the configuration file is automatically created during the process. In cases where SaltStack (or any other deployment tool) handles the deployment, it’s useful to log a message indicating that the file was created, but raising an error could unnecessarily halt the process even when everything is working as expected.

robertcheramy:
I think this is a philosophy thing. When there is no config, oxidized will exit with exit code 1. And yes, this is an error.

PR #3303, opened 2024-10-26
Base for Yard docs

skiera6:
## Pre-Request Checklist

## Description

- Started yard documentation.
- Changed namespaces in a way that will make for a better documentation.
- Separated errors into classes.
- Adapted rake tests.


skiera6:
This commit should fix couple of the rubocop errors, my IDE was acting up and instead of keeping the changes i wanted it merged both old and new code. Just to be safe i will review the pull request when i have some free time also i didn't want to change metrics and i won't change cnos model just to be safe that it won't break anything.

robertcheramy:
While I **greatly** appreciate the effort that went into creating documentation for Oxidized (Thank you!), I must say I am quite overwhelmed by this PR.

My feeling is that you are doing too much at one time, resulting in a PR that is very difficult to handle:
- rubocop does not pass
- rake test does not pass
- there are modifications (for example in spec/node_spec.rb) that do not seem to be related to documentation
- I've found some code duplication somewhere - can't find it anymore - there are too much file changed [fixed by last commit]
- The use of @!visibility private does not seem to be used in the intended way (hiding nicely written documentation)
- when the points above will have been corrected, no one will be able to check what has been done

I suggest breaking the PR down into smaller steps (and PRs):
- A first step would be to make a yard documentation and add yard to gemspec and Rakefile (`rake doc`). No code change, only documentation (even if yard complains and the result is ugly).
- When this works, add a submodule for input
- When this works, add a submodule for hooks
- When this works, add a submodule for errors

Submodules for source and output have already been introduced by another PR - no work to do here.

I'm not sure I want submodules for models, as they are kind of a DSL and that makes writing models unnecessarily complicated. I'm not even sure I want a yard documentation for models, the yard output doesn't make a lot of sense (Instance Method Summary: #clean, #prompt).

The steps above could also be done as commits inside a branch, but I think it is easier to work with PR, as I can check and merge at each step, and changes done somewhere else in master will be included in the next step, reducing conflicts. The steps mentioned above are independent from each other an can be merged independently into master.

skiera6 (commented):
Fixed more errors

skiera6:
> My feeling is that you are doing too much at one time, resulting in a PR that is very difficult to handle:

I felt that might be an issue and you are right

> I've found some code duplication somewhere - can't find it anymore - there are too much file changed [fixed by last commit]

I think i found two more and fixed them

> there are modifications (for example in spec/node_spec.rb) that do not seem to be related to documentation

Yeah i tried to make the tests pass after changing submodules and those were the leftovers of debugging now fixed

> The use of @!visibility private does not seem to be used in the intended way (hiding nicely written documentation)

There's a lot of scattered comments that don't make sense and would make creating the yard docs harder instead of deleting them i opted to just make all of the comments private so they can be cleaned up later as the documentation gets better.

>when the points above will have been corrected, no one will be able to check what has been done

Also one of the main issues i was thinking about you are right


When i'll find some time i will split this PR into separate PRs making it easier to deal with

robertcheramy:
Hello Eryk,

I think that if you go on fixing the issues of the PR, it will just make more to split it afterwards in smaller PRs, which would be unnecessary work.
So I'm also fine with the approach you have taken so far - fix the issues of the PR, rebase or merge with main. When you are finished, mention me with @robertcheramy and I'll take a few cups of coffee and will review the Big PR.

robertcheramy:
Closing this stale PR as it cannot keep up with the code changes.

I (and am under the impression that ytti thinks the same) prefer punctual comments on difficult code sections over comments that describe easy to understandable code, so I don't think we need an API-Like documentation for oxidized.

PR #3317, opened 2024-11-12
Riverbed

Swaeltjie:
## Pre-Request Checklist

## Description
New model for Riverbed Steelheads.

### DIFF
--- a/lib/oxidized/model/riverbed.rb
+++ b/lib/oxidized/model/riverbed.rb
@@ -1,101 +1,82 @@
 class Riverbed < Oxidized::Model
   using Refinements
 
-  # Define the prompt
-  PROMPT = /^.*\s*[\w-]+\s*[#>]\s*$/i
-  prompt PROMPT
+  # Define the prompt directly
+  prompt /^.* *[\w-]+ *[#>] *$/
 
   # Define comment character
   comment '! '
 
-  # Use procs to remove unwanted lines
-  procs do
-    # Remove "Last login" message
-    proc { |output| output.gsub!(/^Last login:.*\n/, '') }
-  end
-
   # Remove sensitive information
   cmd :secret do |cfg|
-    cfg.gsub! /^(\s*tacacs-server (.+ )?key) .+/, '\\1 <secret hidden>'
-    cfg.gsub! /^(\s*username .+ (password|secret) \d) .+/, '\\1 <secret hidden>'
-    cfg.gsub! /^(\s*ntp server .+ key) .+/, '\\1 <secret hidden>'
-    cfg.gsub! /^(\s*ntp peer .+ key) .+/, '\\1 <secret hidden>'
-    cfg.gsub! /^(\s*snmp-server community).*/, '\\1 <configuration removed>'
-    cfg.gsub! /^(\s*ip security shared secret).*/, '\\1 <secret hidden>'
-    cfg.gsub! /^(\s*service shared-secret secret client).*/, '\\1 <secret hidden>'
-    cfg.gsub! /^(\s*service shared-secret secret server).*/, '\\1 <secret hidden>'
+    cfg.gsub! /^( *tacacs-server (.+ )?key) .+/, '\\1 <secret hidden>'
+    cfg.gsub! /^( *username .+ (password|secret) \d) .+/, '\\1 <secret hidden>'
+    cfg.gsub! /^( *ntp server .+ key) .+/, '\\1 <secret hidden>'
+    cfg.gsub! /^( *ntp peer .+ key) .+/, '\\1 <secret hidden>'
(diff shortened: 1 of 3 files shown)


lib/oxidized/model/riverbed.rb
```diff
@@ -0,0 +1,131 @@
+class Riverbed < Oxidized::Model
+  using Refinements
+
+  # Define the prompt
+  PROMPT = /^.*\s*[\w-]+\s*[#>]\s*$/i
```
robertcheramy, line 5: - Directly use `prompt /^.*\s*[\w-]+\s*[#>]\s*$/`
- You don't need case insensitive mode as you don't use characters in the regexp
- use a whitespace instead of \s if you are just matching a whitespace

lib/oxidized/model/riverbed.rb
```diff
@@ -0,0 +1,131 @@
...
+  prompt PROMPT
+
+  # Define comment character
+  comment '! '
+
+  # Use procs to remove unwanted lines
```
robertcheramy, line 11: - The "Last login" message should never appear, as Oxidized does not interpret the initial prompt.
- Do not use procs directly. Use [cmd :all](https://github.com/ytti/oxidized/blob/master/docs/Ruby-API.md#cmd)

lib/oxidized/model/riverbed.rb
```diff
@@ -0,0 +1,131 @@
...
+  procs do
+    # Remove "Last login" message
+    proc { |output| output.gsub!(/^Last login:.*\n/, '') }
+  end
+
+  # Remove sensitive information
+  cmd :secret do |cfg|
+    cfg.gsub! /^(\s*tacacs-server (.+ )?key) .+/, '\\1 <secret hidden>'
```
robertcheramy, line 19: prefer space to \s when possible, as \s also matches \n and \r

lib/oxidized/model/riverbed.rb
```diff
@@ -0,0 +1,131 @@
...
+    end
+    @version_info = comments.join("\n")
+    ''
+  end
+
+  # Get hardware information and store it
+  cmd 'show hardware all' do |cfg|
+    # Remove the command echo and the hostname from the output
+    cfg = cfg.cut_head
+    cfg.gsub!(PROMPT, '')
```
robertcheramy, line 57: Consider using [cut_both](https://github.com/ytti/oxidized/blob/master/docs/Ruby-API.md#cut_both)

lib/oxidized/model/riverbed.rb
```diff
@@ -0,0 +1,131 @@
...
+          comment_line + "\n" + command_part + "\n"
+        end
+      else
+        line
+      end
+    end.join
+
+    # Prepend date, version, hardware, and serial information to the configuration
+    header = []
+    header << comment("Date of version: #{Time.now.strftime('%Y-%m-%d %H:%M:%S %Z')}")
```
robertcheramy, line 114: Printing the time is not a good idea, as oxidized will store a config after every fetch. The idea is to strip everything changing out, and store the configs in git, so you only get differences and not thousand of identical files.

lib/oxidized/model/riverbed.rb
```diff
@@ -0,0 +1,131 @@
...
+    header << comment(@serial_info) if @serial_info
```
robertcheramy, line 115: Wouldn't it be simpler to output these informations above and not store them into a variable? Most other models do so.

robertcheramy (changes requested):
Thank for your PR and also submitting a YAML-Simulation File AND a unit test. I love this, it makes reviewing the PR much easier.

Please have a look at my review and consider updating the model.

lib/oxidized/model/riverbed.rb
Swaeltjie: Removed the entire procs block.

lib/oxidized/model/riverbed.rb
Swaeltjie: Replaced \s with explicit spaces to match only space characters.

lib/oxidized/model/riverbed.rb
Swaeltjie: Replaced with cfg = cfg.cut_both to remove both the first and last lines in one step.

lib/oxidized/model/riverbed.rb
Swaeltjie: Removed the line that adds the date to the header.

lib/oxidized/model/riverbed.rb
Swaeltjie: Modified the code to output the extracted information directly as comments within the cmd blocks.

Swaeltjie:
Thank you for taking the time to review the PR. It's my first time doing this.

I've updated the file with all the recommendations and replied to all comments. Tested and all is working as expected.

PR #3362, opened 2025-01-09
Allow using regexp for model_map and group_map keys

### LINKED ISSUE #3360: Oxidized unable to parse/search multiple tags in http api call from PRTG
I'm not a seasoned programmer, nor am I a linux engineer. I'm a network engineer with a goal, and I'm motivated.

We are running PRTG with a myriad of devices.
I'm running Oxidized and using the PRTG API to retrieve a list of devices containing certain tags like fortios, procurve, aoscx, etc.
(Since I believe this is the only way Oxidized is able to figure out the type of device and what script to use for logging in and making a backup?)

I've configured it according to the documentation, and that works. As long as I only have 1 tag per device.

```
source:
  default: http
  http:
    url: [Redacted]
    hosts_location: devices
    scheme: https
    secure: false
    map:
      name: device
      ip: host
      model: tags
      group: group
model_map:
  juniper: junos
  cisco: ios
  fortinet: fortios
  aoscx: aoscx
  procuve: procurve
  aossw: procurve
```

When I have more than 1 tag per device in PRTG the tag field in the API call is shown as a single string separated by spaces.

For example:

```
    {
      "device": "LAB-XXX-XXX-SW01",
      "device_raw": "LAB-XXX-XXX-SW01",
      "host": "xxx.xxx.xxx.xxx",
      "host_raw": "xxx.xxx.xxx.xxx",
      "tags": "procurve switch",
      "tags_raw": "procurve switch",
      "group": "LAB-Probe",
      "group_raw": "LAB-Probe"
    },
```

Oxidized then doesn't show the device in the webUI and is not contained in the backup.
I've seen someone (although commented out) use a ruby regex in the config file after model: tags.
(https://github.com/ytti/oxidized/issues/1999)

```
      model: tags !ruby/regexp /\bios|\bdnos|\bnxos|\bftos|\bfortios|\betherhaul|\bsonicos|\bairfiber/ig
```

I tried this, (added procurve to the regex ofcourse) and it doesn't work. all devices are added as Junos (default in the config file)

We use multiple tags for other automation tasks and right now I'd have to break them in order to automatically back-up devices with the correct model in oxidized.

Is there a way to use a regex? and so, what would be the right one to use?

ytti:
Use case is that the source returns string which has unrelated and variant data but the actual model is included somewhere in this data

For example source could have space separated list of tags and one of these tags could be model name. Eg 'procurve switch'.

Closes #3360

## Summary by CodeRabbit

- **New Features**
  - Configuration now supports regular expressions for `model_map` and `group_map` keys.
  - Added support for procurve device model mapping.

- **Documentation**
  - Updated Configuration.md with new mapping and group configuration examples.
  - Added CHANGELOG.md entry for the new configuration feature.

- **Improvements**
  - Simplified configuration mapping logic in source handling.
  - Added new `map_value` method to support flexible key matching.

- **Tests**
  - Added comprehensive test suite for model mapping functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Added
 - junos: add unit test (@systeembeheerder)
 - apc_aos: support for scp (@robertcheramy)
+- config: allow model_map and group_map keys to be regexp. Fixes #3360 (@ytti)
 
 ### Changed
 - sonicos: accept policy message. Fixes #3339 (@Steve-M-C, @robertcheramy)
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -226,6 +226,7 @@ source:
 model_map:
   cisco: ios
   juniper: junos
+  !ruby/regexp /procurve/: procurve
 ```
 
 ## Advanced Group Configuration
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -174,7 +174,16 @@ input:
 
 ## Advanced Configuration
 
-Below is an advanced example configuration. You will be able to (optionally) override options per device. The router.db format used is `hostname:model:username:password:enable_password`. Hostname and model will be the only required options, all others override the global configuration sections.
+Below is an advanced example configuration.
+
+You will be able to (optionally) override options per device.
+The router.db format used is `hostname:model:username:password:enable_password`.
+Hostname and model will be the only required options, all others override the
+global configuration sections.
+
+Custom model names can be mapped to an oxidized model name with a string or
+a regular expression.
+
 
 ```yaml
 ---
@@ -269,7 +278,8 @@ groups:
           ssh_keys: "~/.ssh/id_rsa_bar_vyatta"
 ```
 
-For mapping multiple group values to a common name
+For mapping multiple group values to a common name, you can use strings and
+regular expressions:
 
 ```yaml
 group_map:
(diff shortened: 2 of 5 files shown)


ytti:
I honestly cannot remember why I am normalising all keys to strings, but tests at least pass even if we don't normalise them.

ytti:
Another option, which doesn't require changing the key_to_s would be to introduce new config stanza like.

```yaml
source:
  map:
    model:
       "oxidized_model": "source_model"
```

By swapping the order, we would guarantee the key is string, and value could be potentially any other class. 

How we could transition to this new config stanza is that we could check existence of old, if old is found, we could emit warning that it's being deprecated and removed later in X, potentially link to documentation. Then we swap key/value and process the old and new stanza the same way. 

I don't feel strongly.

ytti:
Closing and reopening, as I want to see what CodeRabbit does.

ytti:
@coderabbitai full review

robertcheramy:
I'd like to extend the documentation:
- Add some more texte in docs/Configuration.md (and not just an example)
- Briefly comment the code. It is difficult for an AI to find the point between too short and over verbose. We will see if I am better at it ;-)
I'll need a few day for this - I'm currently working on other topics.

ytti:
I'm ok with that. Personally I don't write or read ever comments which is English version of the code, to me it is non-sensical to maintain two versions of code, one in formal language another in natural language. I don't read them, because I don't trust that they are in sync. I also try to write methods which are short, do one thing without side effects and fit in mental space, in which case it is often faster to just read the code. I think comments make more sense when the 'why' is important, not the 'what', like when you know what the code does, but you might not understand why it is needed. Or if code is particularly tricky and I couldn't figure out how to write it elegantly and simply. But each to their own, and I'm not going to reject any style of commenting.

PR #3384, opened 2025-02-02
Cache git logs to accelerate displaying  the versions of a node in oxidized-web

### LINKED ISSUE #3121: Git performance problems with large device count
Dear Maintainer,

My company uses Oxidized to store device configurations for various kinds of network devices. One of our instances currently has about 700 devices and the bare git repo has grown to 300~GB. Currently, it times out when requesting versions for devices probably because of the size of the repo. 

Are there any optimizations you think might help in this situation or is there any profiling I could help with? Oxidized is an important tool for us and want to help where we can. :)

robertcheramy:
## Pre-Request Checklist

## Description
In order to address the performance issues observed in oxidzed-web when listing the versions of a node, this PR caches git log into memory. At the first time, we still have a slow performance, subsequent calls only need to scan new commits and are *very* much faster.

Closes #3121

### DIFF
--- a/lib/oxidized/output/git.rb
+++ b/lib/oxidized/output/git.rb
@@ -64,8 +64,8 @@ module Oxidized
       # Returns the configuration of group/node_name
       #
       # #fetch is called by Nodes#fetch
-      # Nodes#fetch creates a new Output object each time, so we cannot
-      # store the repo index in memory. But as we keep the repo index up
+      # Nodes#fetch creates a new Output object each time, so it not easy
+      # to cache the repo index in memory. But as we keep the repo index up
       # to date on disk in #update_repo, we can read it from disk instead of
       # rebuilding it each time.
       def fetch(node, group)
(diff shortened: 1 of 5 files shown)


lib/oxidized/output/git.rb
```diff
@@ -143,6 +127,76 @@ def get_diff(node, group, oid1, oid2)
...
+              cache[filename] = [hash]
+            end
+          end
+        end
+
+        cache.each_pair do |filename, hashlist|
+          if @gitcache[repo_path][:nodes][filename]
+            # using the splat operator (*) should be OK as hashlist should
+            # not be very big when working on deltas
+            @gitcache[repo_path][:nodes][filename].prepend(*hashlist)
```
robertcheramy, line 185: @ytti - do you have experience or an opinion on using the splat operator for prepending an array to an array?
There are alternatives to the splat operator, but I found this to be the most readable and efficient.
I read that when the array is smaller than 1000 entries, the splat operator is not a problem.

PR #3433, opened 2025-02-27
srosmd: add ignore regex for 64-bit system up time

Emiliaaah:
## Pre-Request Checklist

## Description

Somewhere between version `B-23.7.R2` and `B-23.10.R2` on the Nokia 7750 SR, a field for 64-bit system uptime was added. Before this didn't get filtered, so this PR fixes that.

The system information now includes both system uptime fields like this:
```
...
System Up Time         : 1 days, 01:01:01.11 (hr:min:sec)
System Up Time (64-bit): 1 days, 01:01:01.11 (hr:min:sec)
...
```

### DIFF
--- a/lib/oxidized/model/srosmd.rb
+++ b/lib/oxidized/model/srosmd.rb
@@ -26,7 +26,7 @@ class SROSMD < Oxidized::Model
     #
     # Strip uptime.
     #
-    cfg.sub! /^System Up Time.*\n/, ''
+    cfg.gsub! /^System Up Time.*\n/, ''
     comment cfg
   end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Fix frozen string literals (@robertcheramy)
 - powerconnect: Cleanup login/logout logic. Fixes #3437 (@clifcox)
 - aos7: remove extra lines occuring when `show hardware-info` runs slow (@rouven0)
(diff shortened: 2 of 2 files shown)


lib/oxidized/model/srosmd.rb
```diff
@@ -27,6 +27,7 @@ class SROSMD < Oxidized::Model
     # Strip uptime.
     #
     cfg.sub! /^System Up Time.*\n/, ''
+    cfg.sub! /^System Up Time \(64-bit\).*\n/, ''
```
robertcheramy, line 30: This is basically the same regexp as above, matching less text. You probably want to use `gsub!` instead of `sub!`.

Emiliaaah: I wasn't aware of there being a `gsub!`, that fits the bill perfectly though. In my latest commit I changed it to use that instead.

Emiliaaah:
I agree with @wdoekes that `[^\n]*` is a nicer solution, so I updated this PR to use `[^\n]*` instead of `.*`.

lib/oxidized/model/srosmd.rb
```diff
@@ -26,7 +26,7 @@ class SROSMD < Oxidized::Model
     #
     # Strip uptime.
     #
-    cfg.sub! /^System Up Time.*\n/, ''
+    cfg.gsub! /^System Up Time[^\n]*\n/, ''
```
robertcheramy, line 29: This is unnecessary as `.` [Matches any character except a newline](https://ruby-doc.org/3.3.2/Regexp.html#class-Regexp-label-Shorthand+Character+Classes)

robertcheramy (approved):

PR #3441, opened 2025-03-03
New model for Eaton Gigabit Network Card M3

thanegill:
## Pre-Request Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - unifiap: new model for Unifi APs, switches, and AirOS APs (@clifcox)
 - github: Issue templates for bugs, feature requests and support requests (@robertcheramy)
 - model for Ingate Operating System (@thanegill)
+- model for Easton Gigabit Network Card (@thanegill)
 
 ### Changed
 - fortios: support for FortiADC (@electrocret)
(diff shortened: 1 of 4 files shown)


lib/oxidized/model/eaton_network.rb
```diff
@@ -0,0 +1,49 @@
+class EatonUPS < Oxidized::Model
+  using Refinements
+  # Eaton Gigabit Network Card M3
+
+  # -p option is a passphrase used to encrypted parts of the config data, the
+  # encrypted data is nondeterministic and changes with each run. Set a static
+  # passphrase, and remove all references to encrypted data.
```
robertcheramy, line 7: The task of removing passwords should be done in a secret bloc, so that the use can choose to keep passwords or not:
- https://github.com/ytti/oxidized/blob/master/docs/Ruby-API.md#cmd
- https://github.com/ytti/oxidized/blob/master/docs/Configuration.md#removing-secrets
- Example: https://github.com/ytti/oxidized/blob/af6d6af5b16425f9d17d63712a770b2278bd5b5f/lib/oxidized/model/ios.rb#L28

thanegill: Done. Moved secret removal to `:secret`

thanegill:
Added model note about how to enable secret removal for all `eatonnetwork` models, as a deterministic backup—at least for me—is the desired default.

PR #3476, opened 2025-03-29
Add iosxr support to SyslogMonitor

deesel:
## Pre-Request Checklist

## Description

- Simplify `SyslogMonitor` code by using regular expressions for syslog message parsing.
- Added `iosxr` support to `SyslogMonitor`.
- Added `content-type` header for `PUT` request in [`rest_client.rb`](/extra/rest_client.rb), since newer versions of Sinatra apparently require usage of correct `content-type` header for payload parsing.


ytti:
I am not against this, I'm just worried that the next model added or already in someone's local copy, they depend on having more complex way to work with the data, than regexp, and they need model specific function to work with the data. 

I did also notice that this is much more strict, in that it assumes configuration comes from console and IPv4 address. The existing implementation basically splits on whitespace, which doesn't care about that. So this may break some existing setups. 

It also does not handle Junos returning 'none' as commit message, when commit message is not given. I'm not sure if this is important and certainly adding negative lookahead to regex would be quite messy.

deesel:
> I am not against this, I'm just worried that the next model added or already in someone's local copy, they depend on having more complex way to work with the data, than regexp, and they need model specific function to work with the data.
> 
> I did also notice that this is much more strict, in that it assumes configuration comes from console and IPv4 address. The existing implementation basically splits on whitespace, which doesn't care about that. So this may break some existing setups.
> 
> It also does not handle Junos returning 'none' as commit message, when commit message is not given. I'm not sure if this is important and certainly adding negative lookahead to regex would be quite messy.

I understand. In that case I just added support for `iosxr` with original parsing method.

ytti:
I didn't have much confidence in my argument, and would have accepted the whole PR as-is. Apologies if it read as dismissive of the work, I think it's legitimate alternative and I just don't have high confidence which trade-offs are best here.

deesel:
No problem, I plan to add support for new device models (mostly wireless APs since I work for an ISP) in relatively near future, so we can revisit this topic. I will think about your arguments and hopefully come up with better solution.

Issue #1805, opened 2019-05-09
Remove saved configuration for devices not listed in sources anymore


cr41gc:
I would like the ability to archive a device using the git output. I don't need to delete the device's history, but it would be nice to prevent the file from showing up when other users clone the backups repo.

I'd like the latest commit of the git repo to represent the state of the network, not just be an all-knowing backups location.

I've tried stopping oxidized, deleting device files from the git repo (not from oxidized's copy), then doing a `git fetch -q origin master:master`.

I tried to get around this by using the file output and doing `git push` commands myself, but then I found out the file output doesn't support `type_as_directory`, which I'm using to collect extra commands and put them into another folder.

Does anyone have a good way to archive or delete a backup file with the git output?

devicenull:
We run this bash script weekly to cleanup the old devices... it'll probably need some updating for your environment:

```
#!/bin/bash


export OXIDIZED_PATH='/home/oxidized/.config/'

export WORKING_DIR='/tmp/oxidized_cleanup'

if [ `whoami` != 'oxidized' ]; then
        echo "Must run as oxidized user to avoid trashing repo permissions"
        exit 1
fi

rm -rf $WORKING_DIR
mkdir -p $WORKING_DIR
cd $WORKING_DIR
echo "Checking out repo"
git init-db
git remote add origin $OXIDIZED_PATH/oxidized/oxidized.git
git pull origin master

sudo service oxidized stop
echo "Evalulating files"
for filename in `git ls-tree --full-tree -r HEAD | awk '{print $4}' | grep -v \.xml`; do
        DEVICENAME=${filename//\.xml//}
        echo $DEVICENAME
        grep $DEVICENAME $OXIDIZED_PATH/router.db 2>&1 > /dev/null
        if [ $? -ne 0 ]; then
                echo "remove $filename"
                git rm -f $filename ${filename}.xml
        fi
done

echo "Comitting results"
git commit -a -m "Auto-purge"
git push origin master
sudo service oxidized start
```

991jo:
Hello

I have the same issue. I would like to remove old devices from the HEAD of the git repo. It is okay that they remain in the git history, but we often use tools like "grep" on the config files to check stuff like "how many devices have config setting X". And when you have the old devices that no longer exist in the network in the repo you have false positives.

We have a script that cleans up the git repo, but the first commit that oxidized does adds all files that have been deleted.

the short version of what we are doing is:

```
systemctl stop oxidized
[...]

PR #3486, opened 2025-04-09
Begin the commit process with index reset

garryshtern:
## Pre-Request Checklist

## Description

outputs/git: Begin the commit process with the index reset to the tree of the HEAD commit.

This prevents the unintentional re-inclusion of files in the repo that are not referenced by the HEAD commit or the device being updated

This should close issue #1805

### DIFF
--- a/lib/oxidized/output/git.rb
+++ b/lib/oxidized/output/git.rb
@@ -252,6 +252,7 @@ module Oxidized
         oid = repo.write data, :blob
         # Read the index from disk
         index = repo.index
+        index.read_tree repo.head.target.tree unless repo.empty?
         index.add path: file, oid: oid, mode: 0o100644
 
         repo.config['user.name']  = @user


robertcheramy:
I disagree on this.

Re-reading the tree costs computing ressources, see the comment above the function. Rereading at each commit seems a waste of ressources to me.

The only reason why the index becomes out of sync is the user tempering with the git repository of oxidized. This is documented at https://github.com/ytti/oxidized/blob/master/docs/Troubleshooting.md#oxidized-ignores-the-changes-i-made-to-its-git-repository 

If we want to get rid of the index on disk, so that it becomes more intuitive for the user (stop oxidized, fiddle with git, start oxidized), we could cache the index with the help of a cache method. Or we keep the index on disk and only track the information that the index wasn't re-read since last start of oxidized.
Last time I worked on this part of oxidized, I decided that the documentation above is enough and the effort was not worth it. But I may be wrong.

#1805 is about deleting devices configuration from when they are not listed in source anymore., so this PR does not solve the issue.

garryshtern:
That’s not the only way for index to be out of sync. You are assuming that there is only a single instance of oxidized that is setup, which isn’t always the case. If you have two parallel deployments with the same repo backend, then a change in one will break the other one, without an index reread.

robertcheramy:
I would not recommend using the same repo for two oxidized processes. You will run in unexplainable sync issues, even if you re-read the index each time: if both oxidized make a commit at the same time, only one would persist, because they both start with the same index.

PR #3494, opened 2025-04-19
bugfix: skip enable password in awplus

shigechika:
## Pre-Request Checklist

- [x] Passes rubocop code analysis (try `rubocop --auto-correct`)
- [x] Tests added or adapted (try `rake test`)
- [x] Changes are reflected in the documentation
- [x] User-visible changes appended to [CHANGELOG.md](/CHANGELOG.md)

## Description

bugfix: skip enable password in awplus

PR #3503, opened 2025-04-30
awplus: fix bug with enable password when supplied

sgsimpson:
## Pre-Request Checklist

## Description

- #3494 introduced a bug where there is no return character when an enable password is supplied. The ssh session just hangs and times out.
- Test on `x610-5.4.5-3.7` with enable password supplied and on `x210-5.4.6-1.2` with enable = true.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - aosw: fix secret parsing (@rouven0)
 - mlnxos: handle ANSI-ESC codes and pager requests. The prompt has been
   reengineered, open an issue if you experience timeouts. Fixes #3469 (@robertcheramy)
+- awplus: fix enable password when supplied (@sgsimpson)
 
 
 ## [0.33.0 - 2025-03-26]
--- a/lib/oxidized/model/awplus.rb
+++ b/lib/oxidized/model/awplus.rb
@@ -61,25 +61,29 @@ class AWPlus < Oxidized::Model
     cfg
   end
 
-  # Config required for telnet to detect username prompt
-  cfg :telnet do
-    username /login:\s/
+  enable_cmd = proc do |cr|
+    if vars(:enable) == true
+      cmd "enable"
+    elsif vars(:enable)
+      cmd "enable", /^[pP]assword:/
+      cmd vars(:enable) + (cr ? "\r" : "")
+    end
   end
 
-  # Main login config
-  cfg :telnet, :ssh do
-    post_login do
-      if vars(:enable) == true
-        cmd "enable"
-      elsif vars(:enable)
-        cmd "enable", /^[pP]assword:/
-        cmd vars(:enable) + "\r\n"
-      end
-      #      cmd 'terminal length 0' #set so the entire config is output without intervention.
-    end
-    pre_logout do
-      #      cmd 'terminal no length' #Sets term length back to default on exit.
-      send "exit\r\n"
-    end
+  # Config required for telnet to detect username & password prompt.
+  # Telnet doesnt need to send an additional CR for the enable password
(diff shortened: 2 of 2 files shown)


ytti:
cmd does include new line, of course none of the above uses of cmd would work if not. 

Perhaps more likely explanation is that prompt after enable prompt doesnt match, maybe due to ansi codes? And as a workaround you force 2nd prompt.

ytti:
@shigechika can you look at this please

mattie47:
👋👋👋

Just gave my old issue a read through. Any changes around this would need testing with and without enable password for both telnet and ssh. 

I don't work for Allied Telesis any more so also wouldn't have any test equipment for checking things again. 

Perhaps just note my key observations near the end of the linked issue for areas of risk

ytti:
I wonder if we should just always send \r\n in ssh cmd as well. Isn't that what SSH clients anyhow do? We could also gate the behaviour behind variable, if to send \n or \r\n, so we could possibly set that value in model.

mattie47:
> I wonder if we should just always send \r\n in ssh cmd as well. Isn't that what SSH clients anyhow do? We could also gate the behaviour behind variable, if to send \n or \r\n, so we could possibly set that value in model.

I'm going back 8 years in my memory here, but I have a feeling it added an extra new line in the saved config file. So if you switch between ssh and telnet, it would have a diff for an extra new line. Likely a corner case that doesn't matter, but I think I tried to avoid it.

mattie47:
> Sorry for late response. I test on enabel=true which is without password. I will test enable with password on next week.

What model device and software version are you using?

robertcheramy:
> I wonder if we should just always send \r\n in ssh cmd as well. Isn't that what SSH clients anyhow do? We could also gate the behaviour behind variable, if to send \n or \r\n, so we could possibly set that value in model.

I would change such a behavior on a model basis only, as the risk to break a lot of models is too high.

mattie47:
Nice one @sgsimpson . I've added a couple PR comments. 

I think we still need to hear from @shigechika as to what was tested, and on what devices to understand why any change was needed in the first place before any merge.

ytti:
I’m not so sure about this. If this model always wants \r\n on enable pw, making it conditional does not seem reasonable.

I was thinking of implementing newline for cfg block, which takes a string argument and is \n if not specified.

So this model would be

cfg :ssh { newline “\r\n” }

This would cause ‘cmd’ to always end at \r\n instead of bare \n.

ytti:
https://github.com/ytti/oxidized/compare/master...feat-newline
```diff
diff --git a/lib/oxidized/input/cli.rb b/lib/oxidized/input/cli.rb
index af9a0294..d16fcf75 100644
--- a/lib/oxidized/input/cli.rb
+++ b/lib/oxidized/input/cli.rb
@@ -52,6 +52,10 @@ module Oxidized
         @password || (@password = regex)
       end

+      def newline(newline_str = "\n")
+        @newline || (@newline = newline_str)
+      end
+
       def login
         match_re = [@node.prompt]
         match_re << @username if @username
diff --git a/lib/oxidized/input/ssh.rb b/lib/oxidized/input/ssh.rb
index b94d8a75..6c3d6892 100644
--- a/lib/oxidized/input/ssh.rb
+++ b/lib/oxidized/input/ssh.rb
@@ -112,7 +112,7 @@ module Oxidized

     def cmd_shell(cmd, expect_re)
       @output = String.new('')
-      @ses.send_data cmd + "\n"
+      @ses.send_data cmd + newline
       @ses.process
       expect expect_re if expect_re
       @output
```

Can you add this, and then in the model:
```ruby
cfg :ssh do
  newline "\r\n"
end
```

And report back?

sgsimpson:
Thanks @ytti, yes that worked perfectly. I have tested all four combinations again and each worked. I have updated this pull request to use the new newline option (once avaialble in main).

I have left the telnet password prompt fix in place as well.

PR #3508, opened 2025-05-07
Housekeeping

### LINKED ISSUE #3368: Problems running Oxidized on Rocky Linux 9.5
The documentation mentions CentOS 6 and 7 which are ancient by now.
I've been running Oxidized on Ubuntu 24.4 with succes, however to bring Oxidized into Production I have to conform to Rocky Linux 9.

I've tried installing oxidized, the documentation wasn't helpfull. I could share what is going wrong during the proces but with a lot of pain I managed to get Ruby and gem install oxidized and oxidized-web.
However, whenever I try to run oxidized, I get the following error:

```
[oxidizer@localhost ~]$ oxidized
Ignoring debug-1.9.1 because its extensions are not built. Try: gem pristine debug --version 1.9.1
Ignoring racc-1.7.3 because its extensions are not built. Try: gem pristine racc --version 1.7.3
Ignoring rbs-3.4.0 because its extensions are not built. Try: gem pristine rbs --version 3.4.0
/usr/share/rubygems/rubygems.rb:278:in `find_spec_for_exe': can't find gem oxidized (>= 0.a) with executable oxidized (Gem::GemNotFoundException)
        from /usr/share/rubygems/rubygems.rb:297:in `activate_bin_path'
        from /usr/local/bin/oxidized:25:in `<main>'
```

Now this is still an A environment, not yet P.
But I'd like to get the installation process nailed down before I move to P.

also:
```
[oxidizer@localhost ~]$ ruby -v
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-linux]
```

robertcheramy:
## Pre-Request Checklist

## Description
This housekeeping branch cleans up the code and documentation a bit:
- Housekeeping in the code. Although done carefully and baked with unit tests,
  these changes may have introduced new bugs in the code. On the long term,
  they will improve the quality of the code.
  - Maximal line length: 120 char
  - Prefer single-quoted strings when no interpolation or special symbols.
  - Ruboocop fixes: Style/ParallelAssignment, Style/RescueModifier,
    Style/Not, Rake/DuplicateTask, Style/OpenStructUse
- AricentISS: fix codeQL alert #15 (@robertcheramy)
- Update installation instructions on Rocky Linux 9. Fixes #3368 (@robertcheramy)

Closes: #3368


ytti:
Not objecting to change, just pointing out some of this is highly subjective. 

Like I think ’x unless y’ is more readable in simple cases than blocks.
I think ’x = y rescue z’ is fine and readable, despite broad rescue.
I think ’not foo’ can be lot easier to pickup than ’!foo’ particularly in more complex logic cases.
How strings should be is debatable, what is in committed is fine, but also fine is to use ”” by default and use ’’ to signal that interpolation is intentionally turned off for cause. 

Sadly there isnt any agreed style for ruby.  And despite I have preferences, ultimately it is not very important, so this is fine.

robertcheramy:
> Not objecting to change, just pointing out some of this is highly subjective.
> 
> Like I think ’x unless y’ is more readable in simple cases than blocks. I think ’x = y rescue z’ is fine and readable, despite broad rescue. I think ’not foo’ can be lot easier to pickup than ’!foo’ particularly in more complex logic cases. How strings should be is debatable, what is in committed is fine, but also fine is to use ”” by default and use ’’ to signal that interpolation is intentionally turned off for cause.
> 
> Sadly there isnt any agreed style for ruby. And despite I have preferences, ultimately it is not very important, so this is fine.

I tried to stick to rubystyle.org and rubocop defaults, but I have no preferences, so I will revert the subjective commits.

ytti:
Please dont, I dont feel sufficiently strongly. And I suspect most people would rather agree with something online rather than my subjective preferences. Just wanted to add color and how I dont think many of these changes actually increase or decrease quality in any objectively measurable way. Naybe subjectively it now matches to more people than it did.

robertcheramy:
I have documented in `.rubocop.yml` that we explicitly allow the _rescue modifier_ and _not_.  As I also do not feel strongly about this, I prefer to not to enforce a coding style on these points and let developers choose what is right for them.

If someone steps in with strong arguments, we can still change this in the future.

As of strings should be, we have three options:
- Enforce [Single Quotes](https://rubystyle.guide/#consistent-string-literals-single-quote) - the Style in this PR
- Enforce [Double Quotes](https://rubystyle.guide/#consistent-string-literals-double-quote) - the Style you proposed
- Do not enforce a style (revert the commit, let oxidized as it was). If so, we will have no [Consistent string literals](https://rubystyle.guide/#consistent-string-literals)

My argument so far was the performance of using single quotes, but this is not a god argument, as the gain in performance is not very big. I chose single quotes because it was in the majority of the code.
I really like your proposition to use `’` to signal that interpolation is intentionally turned off.
Both are not consistently used in the code (before this PR).

I tend to revert the commit 388cec5995b448c563106d1e4bf037897bb5708e) and document in `.rubocop.yml` that the preferred (but not enforced) Style is double quote.

What du you think?

ytti:
If you like the idea that we only use single quotes to show that interpolation is off for cause, I think we can enforce double quotes.

ytti:
I keep thinking of ’x unless y’ and ’x if y’ compare to blocks. And I wonder or what natural languages you speak affects the perception of readability. In English we could say ’unless it is warm, im staying home’ or ’IM staying home unless it is warm’. And to me the first one sounds better. But in Finnish it would be the opposite, particularly with unless condition would more naturally follow statement.

More generally I think for short examples in-line has lower cognitive cost and for longer examples blocks. Which doesnt lend itself for very formal standard :/

robertcheramy:
> If you like the idea that we only use single quotes to show that interpolation is off for cause, I think we can enforce double quotes.

Enforcing double quotes changes 2680 String Literals in 243 Files (including `require 'oxidized/version'` > `  require "oxidized/version"`.

I'm pretty sure this is not what we want, so I prefer to stick to the old rule - `Style/StringLiterals:  Enabled: false`

robertcheramy:
> I keep thinking of ’x unless y’ and ’x if y’ compare to blocks. And I wonder or what natural languages you speak affects the perception of readability. In English we could say ’unless it is warm, im staying home’ or ’IM staying home unless it is warm’. And to me the first one sounds better. But in Finnish it would be the opposite, particularly with unless condition would more naturally follow statement.
> 
> More generally I think for short examples in-line has lower cognitive cost and for longer examples blocks. Which doesn't lend itself for very formal standard :/

Are you addressing Commit bb9f7b3e21d500f72e1aae1acfe6f7178a0abd90 ?

I think [Rubocop Style/IfUnlessModifier](https://docs.rubocop.org/rubocop/cops_style.html#styleifunlessmodifier) does exactly what you describe:
- If the resulting line is < 120 char, it will go for inline
- If the resulting line is too long, it will go for blocks

I wasn't happy with Style/IfUnlessModifier before this PR as we did not enforce a maximum line length. Now we have max 120 char, so it's OK for me.

I'm OK with both inline and block when using if and the line doesn't become too long.
Bur I have difficulties with unless in both inline and block. Probably because I need two to three words in French and German to express "unless" ;-)

Back to this PR - I did not change the cop `Style/IfUnlessModifier` (rubocop default = enabled), just the max line length, so I assume this is OK.

ytti:
```Style/StringLiterals:  Enabled: false``` is fine, thanks


Also agree with ```Style/IfUnlessModifier``` 

Having said so, while I have some preference, if someone has strong preference I'm happy to change my mind.

Issue #3512, opened 2025-05-09
PR #3495 modifies saved configuration on non-cluster devices


robertcheramy:
@systeembeheerder writes in https://github.com/ytti/oxidized/pull/3495#issuecomment-2865477260:

> Other case outputs error. but no problem.
I do think that is a problem. After this update i get a diff on ALL my SRX devices:

`+# error: Chassis cluster is not enabled.`  

An successful config backup shouldn't add errors imho

_Originally posted by @systeembeheerder in https://github.com/ytti/oxidized/issues/3495#issuecomment-2865477260_

systeembeheerder:
A solution would be something like:
```
  when /^srx/
    out << cmd('show chassis cluster status') do |cfg|
      cfg.include?('error: Chassis cluster is not enabled.') ? '' : comment(cfg)
    end
```

PR #3495, opened 2025-04-19
Added support for "show chassis cluster status" output on SRX series

shigechika:
## Pre-Request Checklist

- [x] Passes rubocop code analysis (try `rubocop --auto-correct`)
- [x] Tests added or adapted (try `rake test`)
- [x] Changes are reflected in the documentation
- [x] User-visible changes appended to [CHANGELOG.md](/CHANGELOG.md)

## Description

Added support for "show chassis cluster status" output on SRX series

PR #3515, opened 2025-05-12
Fix error message introduced by PR #3495

robertcheramy:
## Pre-Request Checklist

## Description
Fix error message introduced by PR #3495
Applying the fix proposed by @systeembeheerder
Closes: #3512

Note - `cfg = cfg.cut_both if screenscrape` does not seam to work correctly. resulting in the prompt being spread through the output. But this is not the topic of this fix.

### DIFF
--- a/lib/oxidized/model/junos.rb
+++ b/lib/oxidized/model/junos.rb
@@ -33,7 +33,9 @@ class JunOS < Oxidized::Model
     when /^(ex22|ex3[34]|ex4|ex8|qfx)/
       out << cmd('show virtual-chassis') { |cfg| comment cfg }
     when /^srx/
-      out << cmd('show chassis cluster status') { |cfg| comment cfg }
+      out << cmd('show chassis cluster status') do |cfg|
+        cfg.include?('error: Chassis cluster is not enabled.') ? '' : comment(cfg)
+      end
     end
     out
   end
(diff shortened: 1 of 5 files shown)


systeembeheerder:
@robertcheramy seems to work for SRX. But I think we should also fix this for the switches; unfortunately i  don't have any junos switches that are not member of a virtual chassis.  Maybe revert #3495 until we have a solid test on those as well?

ytti:
Personally I dont think static unchanging error line on config which you can load to device, because error is comment isnt broken. I’m not against fixing it, but I’m not willing to say as a rule it is broken.

Perhaps easy test is return ”” if cfg.len < 2 —- or some such.

robertcheramy:
This is a pattern used some models, at least for unsupported commands (https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/aosw.rb#L57)

But I agree that this case is different as the command is supported, only the output of the command itself is an error.

Would `if cfg.lines.count <= 3 and cfg.include?("\nerror:")` be an option?

@shigechika - do you have some srx switches not member of a virtual chassis? Knowing how their output is on `show chassis cluster status` would help to fix the problem.

robertcheramy:
> Note - `cfg = cfg.cut_both if screenscrape` does not seam to work correctly. resulting in the prompt being spread through the output. But this is not the topic of this fix.

It does. The YAML Simulation file is not in exec mode.

PR #3522, opened 2025-05-17
Model for Netgate TNSR

Vantomas:
## Pre-Request Checklist

## Description

Hello, I have implemented the netgate tnsr model. I use it in my dayjob, so I'm willing to be a maintainer of this model.


lib/oxidized/model/tnsr.rb
```diff
@@ -0,0 +1,49 @@
+class TNSR < Oxidized::Model
+  using Refinements
+
+  # Netgate TNSR #
+
+  prompt /^.+[#]\s$/
```
robertcheramy, line 6: This prompt is quite unspecific, as it will match any line ending with `# `. This may lead to timeouts if there is one line ending with `# ` which is not a prompt.
https://docs.netgate.com/tnsr/en/latest/basics/working-cli.html#command-prompt specifies the prompt to be `<hostname> tnsr<(mode)># `, perhaps `/^\s tnsr# /`could be a more specific regexp?

Note: you can test different prompts with https://github.com/ytti/oxidized/blob/master/docs/ModelUnitTests.md#device-prompt

robertcheramy (commented):
Thank you for proposing a new model and steeping in to be a model maintainer!

Please see my comment on the prompt definition.

lib/oxidized/model/tnsr.rb
Vantomas: Thanks for your suggestion, I improved it.

Vantomas: Okay, now it is fixed for my usecase.

PR #3547, opened 2025-06-21
Remove uptime from vyatta devices.

ShaneMcC:
## Pre-Request Checklist

## Description

Some Ubiquiti Edgerouter devices include the system uptime in the `show version` output, this change removes that.


systeembeheerder:
@robertcheramy  @ShaneMcC 

> Some Ubiquiti Edgerouter devices include the system uptime in the show version output, this change removes that.

Then why is this done in [lib/oxidized/model/vyatta.rb](/ytti/oxidized/tree/master/lib/oxidized/model/vyatta.rb) instead of [lib/oxidized/model/edgeos.rb](/ytti/oxidized/tree/master/lib/oxidized/model/edgeos.rb) ?!

I only have VyOS as Vyatta model and it does not show uptime in `show version`. Hence it is not in (my) commit f48b8b5 .

ShaneMcC:
Oh you're right, there is a separate `edgeos` that hadn't been noticed here - looks like the devices I noticed this on were being polled as vyatta because edgeos is also a fork and that worked.

Issue #3106, opened 2024-03-12
Fastiron Suggested Tweaks


klookee:
Hi All,

Apologies if this is not the right place for this sort of feedback, but here goes.

I've been working with Oxidized collecting configs from a network of Ruckus switches running Brocade's Fastiron software. There are a couple of changes I needed to make to the fastiron.rb file, to get it to work, which I thought I should share.

### Not Matching Configured Prompt

The first issue was a 'not maching configured prompt' error when connecting to a stack of Ruckus switches:

> WARN -- : xxx.xxx.xxx.xxx raised Oxidized::PromptUndetect with msg "SSH@switch-001/002/003/123# not matching configured prompt (?-mix:^([\w.@()-]+[#>]\s?)$)"

The switch stack's prompt is a combined reference of the switch hostnames that make up the stack. The separating forward-slash (/) did not allow the prompt to meet the regular expression defined on line 4 of the fastiron.rb file.

I made the following change to the regular expression on line 4 to correct the issue:

`prompt /^([\w.@\/()-]+[#>]\s?)$/`

### Switch Stack Uptime Value

Once Oxidized was able to collect the configs, it was updating the Git repo after every run. I discovered that the output from the 'show stack' command, included a stack uptime value. For example:

> T=58d16h38m53.9: alone: standalone, D: dynamic cfg, S: static

Obviously the T value continues to tick away, so is seen by Oxidized/Git as a change at each sweep.

To correct this I altered line 44 of the fastiron.rb, to exclude the T value from the 'show stack' output:

`cmd 'show stack | exclude T=' do |cfg|`

I am not a developer or regular expression expert, so am very happy for those who are to suggest better fixes for these problems. I hope my feedback helps in some small way.

Thanks for Oxidized, keep up the great work.

robertcheramy:
Please submit a PR for this.

PR #3579, opened 2025-07-18
adjust prompt to account for stacks, remove time from stack output

ManoftheSea:
## Description
Closes (already closed and stale) issue #3106


robertcheramy:
Please describe your changes in CHANGELOG.md
It would also be nice to provide some example prompts in `/spec/model/data/fastiron#generic#prompt.yaml`. See https://github.com/ytti/oxidized/blob/master/docs/ModelUnitTests.md#device-prompt

Issue #3581, opened 2025-07-21
config file is relative to homedir, should be absolute path


ManoftheSea:
- [X] I have read and understood [docs/Issues.md](https://github.com/ytti/oxidized/blob/master/docs/Issues.md).

**Describe the bug**
I expected to be able to run oxidized like `oxidized --config-file /path/to/my/config`, but oxidized looks for `$OXIDIZED_HOME/path/to/my/config`

**Expected behavior**
I want to be able to provide an absolute path to the configuration file.

**Running environment (please complete the following information):**

- OS: NixOS 25.05
- oxidized version: 0.33.0

I will shortly submit a PR for this issue.  This is likely a breaking change for anyone currently using the "--config-file" option, but I do not see this being used in the extra/ directory's service files.

robertcheramy:
As I wrote in PR #3582 , I don't think it is possible to implement the behavior you wish without changing Asetus. What will work is to use --home-dir to set the path and --config-file to set the name.

robertcheramy:
The documentation of the command line may be improved, but I don't have time to spend on it.

PR #3582, opened 2025-07-21
allow use of absolute directory to config file

ManoftheSea:
## Description
Allow use of absolute path to configuration file rather than requiring it to be under the homedir.
Closes issue #3581

### DIFF
--- a/lib/oxidized/config.rb
+++ b/lib/oxidized/config.rb
@@ -16,9 +16,9 @@ module Oxidized
 
     def self.load(cmd_opts = {})
       usrdir = File.expand_path(cmd_opts[:home_dir] || Oxidized::Config::ROOT)
-      cfgfile = cmd_opts[:config_file] || 'config'
+      cfgfile = cmd_opts[:config_file] || File.join(usrdir, 'config')
       # configuration file with full path as a class instance variable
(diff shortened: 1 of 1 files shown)


lib/oxidized/config.rb
```diff
@@ -16,9 +16,9 @@ class Config
 
     def self.load(cmd_opts = {})
       usrdir = File.expand_path(cmd_opts[:home_dir] || Oxidized::Config::ROOT)
-      cfgfile = cmd_opts[:config_file] || 'config'
+      cfgfile = cmd_opts[:config_file] || File.join(usrdir, 'config')
       # configuration file with full path as a class instance variable
-      @configfile = File.join(usrdir, cfgfile)
+      @configfile = cfgfile
       asetus = Asetus.new(name: 'oxidized', load: false, usrdir: usrdir, cfgfile: cfgfile)
```
robertcheramy, line 22: This will not work as Asetus awaits a config filename, not an absolute path. Asetus then checks for this filename in `/etc/oxidized` and usrdir.

robertcheramy (commented):
I don't think it is possible to implement the behavior you wish without changing Asetus. What will work is to use --home-dir to set the path and --config-file to set the name.

PR #3584, opened 2025-07-22
Refactor config/vars

ytti:
Currently we accept nil value as terminating search, which means we don't fall back to less specifics when nil is found. This refactor fixes that, adds tests and hopefully improves readability.


ytti:
Ah crap, these need to be changed for new logging.

PR #3623, opened 2025-08-28
New microchip syncserver model

dprillxyz:
## Pre-Request Checklist

## Description
This PR adds support for Microchip SyncServer S650/S650i devices running firmware 5.2.x.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ## [Unreleased]
 
 ### Added
+- Add support for Microchip SyncServer S650/S650i model (@dprillxyz)
 - Allow setting timeout on per node basis. Closes #3612 (@ytti)
 
 ### Changed


ytti:
This does seem to do excessive amount of work in the model, which likely should be in the input/http. I didn't spend time thinking about why it is doing and if input/http needs refactoring to reduce work in the model site.

Because so few models use input/http, we likely do not have yet have the right input/http, so it might be that only way to get this model working is to modify input/http.

Particularly because this does the GET request itself, it does not pick up the recent fix to GET on IPv6 address, the more work we perform on models, the less fixes they get.

ytti:
One obvious refactor is `http_get`.

We could the model start using `http_get_res` for `cmd`. And `http_get_res` is just:

```ruby
def http_get_res(path) 
  http_get(path).res
```

So that callbacks can call `http_get` and perform HTTP GET while getting the response object. 

I'm sure there are many other opportunities to improve the input/http to ensure that the models need to do less work.

ytti:
If you would prefer not to review how to refactor input/http to simplify the model, I can do that, if you can provide Internet access to one node to test against.

dprillxyz:
I'll have to circle back to this when I have a bit more free time. Unfortunately I'm not able to provide access to these devices.

PR #3637, opened 2025-09-11
Set Docker shell to bash

electrocret:
## Pre-Request Checklist

## Description
Change the Docker image to use bash as its shell.  

This PR is made in conjunction with a PR to the LibreNMS project. https://github.com/librenms/librenms/pull/18203#issuecomment-3281809514

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@ RUN groupadd -g "${GID}" -r oxidized && useradd -u "${UID}" -r -m -d /home/oxidi
 
 # Set Oxidized user's shell to bash
 RUN chsh -s /bin/bash oxidized
-
+# Remove the existing /bin/sh symlink and create a new one pointing to bash
+RUN rm /bin/sh && ln -s /bin/bash /bin/sh
 
 ##### MSMTP - Sending emails
 # link config for msmtp for easier use.


ytti:
I'm happy to merge this. Just for my education, why is update-alternatives not a solution here?

electrocret:
> I'm happy to merge this. Just for my education, why is update-alternatives not a solution here?

Because that didn't come up as a suggestion in my search results when I was researching how to change the default shell used by a container. 😅 

I'd be happy to use whichever alternatives you'd prefer. 😄

ytti:
I may very well be confused, but if I am not, in Debian and derivates update-alternatives is exactly for this problem.

electrocret:
From my research into update-alternatives, it's not advised to use update-packages to change the shell. https://askubuntu.com/questions/1090939/update-alternatives-for-sh-reports-bash-is-used-but-i-still-get-dash

However according to Docker documentation, the Shell command might be all that's needed. I just haven't had a chance to test if it has the desired effect, meanwhile the "rm /bin/sh && ln -s /bin/bash /bin/sh" has done the trick on my already running container. https://docs.docker.com/reference/dockerfile/#shell

I've updated the PR to just the shell command.

electrocret:
Reread the docs and realized my understanding was wrong. This appears to be the correct way of changing a user's shell.

electrocret:
Looks like chsh isn't being respected. Added back the symlink overwrite.

electrocret:
Looks like the symlink overwrite is needed due to Ruby. It executes everything via /bin/sh, and I'm not finding a way to change that. per: https://ruby-doc.org/core-2.6.5/Process.html#method-c-exec

robertcheramy:
I am not happy with this PR. Changing the system shell to bash is dangerous - we should keep it at the Debian default (dash as a POSIX shell).
Changing the shell of the official Oxidized docker image because of needs in libreNMS is inconsiderate to other users who don't use libreNMS. A better solution would be to write a hook for libreNMS where an API call can be used (without using curl through a shell). A less ideal solution would be to build your own container for this.

Note: I'm currently working on the container image (migration from phusion/baseimage to plain debian/trixie-slim),  and I plan to revert the /bin/sh part of this PR. Setting the shell of the oxidized user seems fine to me, so I'll keep that.

ytti:
Personally I feel like this is war that was lost before it started. Generally speaking scripts are more compatible with bash than dash. 

I could rant extensively how Unix being open makes it immutable and stagnant, as it's nearly impossible to introduce any significant improvements to the ecosystem. Compare to MSFT Powershell which is very modern programming language, Unix has plethora of more or less common shells which have actual proper programming languages, unlike csh/sh et.al. which as absolutely terrible, but there is no reasonable way to actually standardise into any of those superior languages, we're forever doomed to bash. 

Idea that we can pick up a fight here, and force the ecosystem into dash seem futile.  Bash is de-facto and expected. 

Having said that, I'm not going to assert my position here. If someone strongly feels like dash is the right way to go, and everything else should fix around it, so be it.

robertcheramy:
I surely do not want to start a flame war ;-)
Let us try it with bash linked as /bin/sh, and only switch back to dash If problems arise after next release.

PR #3653, opened 2025-10-20
fix: hide power voltage values on aruba aoscx devices

martadams89:
## Description
Adjusts the Aruba AOS-CX config to change the Voltage Input values from a frequently changing numbers to <hidden> blocks.

This stops frequent commits for minor voltage and temperature changes


robertcheramy:
Hi @martadams89 ,

thank you for this PR. As your first commit was a bit raw, I decided to take the idea and implement it "my way", which resulted in PR #3656.

Can you check if the PR addresses your needs of if you need something more? The second commit on your PR makes the diff a bit hard to follow for me, so I'm not sure if PR #3656 fully addresses you needs.

Thanks again for your contribution!

Cheers,
Robert

Issue #366, opened 2016-03-24
Request to add support for Netonix switches


gr0mit1:
I would like to manage Netonix switches using Oxidized.

The config is in a .json file, accessable via ssh according to the following commands:

cmdline
cat config.json
exit

The "exit" gets you back to the CLI.

Any chance of taking a look at this?  I'm happy to make available a switch for testing on a public IP if anyone wants to have a crack.  i've tried doing it myself but have not managed to get it working!

ytti:
I'll take crack at it. You can send access details to saku@ytti.fi hopefully will have time no later than weekend.

PR #3655, opened 2025-10-25
#366 Fix(http): Prevent duplication when handling pagination

kquilliam:
## Pre-Request Checklist

## Description
Fixed loop logic that held onto data on subsequent paginated runs, resulting in duplicated nodes. This was likely to occur with Netbox and large inventory >1000 nodes, beyond the default `MAX_PAGE_SIZE` of 1000 when utilizing `limit=1000` with the Netbox source. For example, if you had 1200 nodes and were pulling batches of 1000, you would end up with 1400 nodes because 1000-1200 were duped. Behavior is as described in a bug report over in the `oxidized-web` project for https://github.com/ytti/oxidized/issues/3676 but wasnt related to `oxidized-web` directly so I am correcting it here.

Should close issue in `oxidized-web` https://github.com/ytti/oxidized/issues/3676


robertcheramy:
The CI failing on ruby-head is OK - it has been fixed on master.
Please document your change in CHANGELOG.md.

robertcheramy:
Note - I transferred issue from oxidized-web to oxidized, it now has the ID #3676.

kquilliam:
Changelog has been updated

PR #3660, opened 2025-10-28
ParksOLT model added

paulocoimbrati:
## Description
Added module for Parks vendor OLT models Fiberlink 40XXX


systeembeheerder:
you might want to add your model to the list:

https://github.com/ytti/oxidized/blob/master/docs/Supported-OS-Types.md

lib/oxidized/model/parksolt.rb
```diff
@@ -0,0 +1,22 @@
+class PARKSOLT < Oxidized::Model
+  using Refinements
+
+  # PARKS OLT #
+
+  prompt /^(\r?[\w.@:\/-]+[#>]\s?)$/
+  comment  '! '
+
+  cmd :all do |cfg|
+    cfg.each_line.to_a[2..-2].join
```
robertcheramy: You are in sum cutting 3 Lines at the beginning and at the end of the running-config.
Is this intentional?
If so, prefer cfg.each_line.to_a[3..-3].join in cmd 'show running-config', as you have no other commands here.

lib/oxidized/model/parksolt.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  cmd 'show running-config' do |cfg|
+    cfg = cfg.each_line.to_a[1..-1].join
+    cfg
```
robertcheramy: cfg.cuth_both should be enough, as it returns the modified config

robertcheramy:
Closing stale PR

PR #3485, opened 2025-04-09
Support for complete inventory

garryshtern:
"show inventory all" provides additional module details, such as optics.

## Pre-Request Checklist

- [ ] Passes rubocop code analysis (try `rubocop --auto-correct`)
- [ ] Tests added or adapted (try `rake test`)
- [ ] Changes are reflected in the documentation
- [ ] User-visible changes appended to [CHANGELOG.md](/CHANGELOG.md)

## Description

Additional inventory information can be seen with "show inventory all", such as optics, and various other modules.

Issue #3680, opened 2025-11-28
Rewrite fortios model


robertcheramy:
- [X] I have read and understood [docs/Issues.md](https://github.com/ytti/oxidized/blob/master/docs/Issues.md).

**Is your feature request related to a problem? Please describe.**
I added `diagnose hardware deviceinfo psu` and `get system interface transceiver` to the FortiOS model in a monkey patch.
The handling of `cmd :all` is not working properly, and sending all the commands in `post do` makes the model difficult to read and maintain.

**Describe the solution you'd like**
I plan to:
- rewrite the fortios model to make it simpler.
- produce YAML simulation files
- remove `| grep .` from the commands as we can handle the pager
- probably add an if statement on `cmd` that is evaluated at runtime, something like `cmd 'show full-configuration', if: -> { vars(:fullconfig) }`

Other ideas are welcome and can be added to this issue.

I have no fortigates with vdoms, so I won't be enable to test everything. As this change could break the FortiOS model for some users, I plan to announce it in the Release notes of 0.35, and make the change in 0.36, so that interested users can test the model (and provide YAML simulation files) before the 0.36 release. => Setting the Milestone to 0.35 for now, in order to remember it when releasing 0.35.

**Describe alternatives you've considered**
Adding more commands into to model without rewriting it. I think cleaning the model is a better solution and worth the effort.

Serazio:
Hi,

we're backing up a lot of fortigates and facing the problem with the changing hashes for every backup. As we need password hashes inside our backups, we implemented a temporary external solution for backing up FortiGates. With the hashes oxidized and git as backend it's generating a ton of unnecessary commits. 

If its somehow possible to read existing backups, it would be great to implement a workflow like the following: 

- Retrieve full backup
- Remove all hashes from backup
- If existing: Compare "hashless" backup with previous "hashless" backup
- If "hashless" backups differ from each other 
- - Save "hashless" backup
- - Save original backup
- else
- - do nothing


In our current solution we are using the following regex:
```- regexp: ".+(conf_file_ver).+"
  replace: "\\1 <line removed>"
- regexp: "(set .+ ENC) .+"
  replace: "\\1 <configuration removed>"
- regexp: "(set .*secret) .+"
  replace: "\\1 <configuration removed>"
- regexp: "(set (?:passwd|password|key|group-password|auth-password-l1|auth-password-l2|rsso|history0|history1)) .+"
  replace: "\\1 <configuration removed>"
[...]

PR #3681, opened 2025-12-01
Introduce conditional commands and fix regession on nxos (show inventory all not supported)

### LINKED ISSUE #3657: NX-OS 5.1(3)N2(1b) doesn't have `show intentory all`
- [x] I have read and understood [docs/Issues.md](https://github.com/ytti/oxidized/blob/master/docs/Issues.md).

**Describe the bug**

**To Reproduce**
1. Configure a Cisco Nexus switch running NX-OS 5.1(3)N2(1b)
2. Use model nxos
3. Run '....'
```
! show inventory all
!                           ^
! % Invalid command at '^' marker.
```

**Expected behavior**
A working config fetched and saved

**Configuration**
```
```

**Logs**
```
```

**Running environment (please complete the following information):**

- OS: [e.g. Debian Bookworm, official container version xxx, ...]
- oxidized version: [e.g. 0.32.2]
- oxidized-web version: [e.g. 0.15.1, if applicable]
- Manufacturer model an software version:
- oxidized model name:

**Additional context**

`show inventory all` was added in https://github.com/ytti/oxidized/commit/42d95640216eef62373115dc7f77bc6c2d0d1581

From the NX-OS cli on our switches:
```
# show inventory ?
  <CR>   
  >      Redirect it to a file
  >>     Redirect it to a file in append mode
  |      Pipe command output to filter


# show inventory all
                         ^
% Invalid command at '^' marker.
```

robertcheramy:
## Pre-Request Checklist

## Description
Closes #3657
PR #3485 introduced a regression as `show inventory all` is not supported in older Nexus versions.

In order to fix this, I introduced the possibility to run a command only if conditions at runtime are met. This has been resolved differently in the past (`cmd` within `cmd` or `cmd` within `post`), but I think having an `:if` argument in `cmd` is  more elegant and flexible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - aoscx: rework environmental data anonymization. Fixes #3568 (@robertcheramy, inspired by PR #3653 by @martadams89)
 - netgear: fix prompt issues caused by ANSI escape codes. Fixes #3287 (@robertcheramy)
 - remove redundant dependency on bundler producing a CI failure on ruby-head (@robertcheramy)
+- nxos: use "show inventory" when "show inventory all" is not supported. Fixes #3657 (@robertcheramy)
 
 
 ## [0.34.3 - 2025-08-05]
--- a/lib/oxidized/model/nxos.rb
+++ b/lib/oxidized/model/nxos.rb
@@ -25,6 +25,16 @@ class NXOS < Oxidized::Model
   end
 
   cmd 'show inventory all' do |cfg|
+    if cfg.include? "% Invalid parameter detected at '^' marker."
+      @show_inv_all_failed = true
+      ""
+    else
+      comment cfg
+    end
+  end
+
+  # 'show inventory all' isn't supported on older versions (See Issue #3657)
+  cmd 'show inventory', if: -> { @show_inv_all_failed } do |cfg|
     comment cfg
   end
 
--- a/spec/model/data/nxos#dummy_no_show_inventory_all#output.txt
+++ b/spec/model/data/nxos#dummy_no_show_inventory_all#output.txt
@@ -1,0 +1,15 @@
+! Cisco Nexus Operating System (NX-OS) Software
+! (...)
+! ...show version output...
+! (...)
+! 
+! (...)
+! ... show inventory - without SFPs ...
+! (...)
+
+!Command: show running-config
+(...)
+... show running config output ...
+(...)
+
+
--- a/spec/model/data/nxos#dummy_no_show_inventory_all#simulation.yaml
+++ b/spec/model/data/nxos#dummy_no_show_inventory_all#simulation.yaml
@@ -1,0 +1,39 @@
+---
+init_prompt: |-
+    
+    (... init prompt ...)
+    \rTEST-12345#\x20
+commands:
+  - "terminal length 0\n": |-
+      terminal length 0\r
+      \rTEST-12345#\x20
+  - "show version\n": |-
+      show version\r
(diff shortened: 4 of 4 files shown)


ytti:
This isn't strictly speaking needed. 

We can already do 

```ruby
pre/post do 
  cmd foo if @foo 
end
```

Only reason `junos.rb` model has this, is to show case it, it was never interesting to me, I just didn't want to spend time thinking of pragmatic example of conditional command. So has been possible day0.

robertcheramy:
Absolutely, I agree that we can already achieve this using pre/post or even by nesting `cmd` calls. And looking at `nxos.rb`, a much simpler implementation could have been:
```ruby
  cmd 'show inventory all' do |cfg|
    if cfg.include? "% Invalid parameter detected at '^' marker."
      # 'show inventory all' isn't supported on older versions (See Issue #3657)
      cfg = cmd 'show inventory'
    end
    comment cfg
  end
```

However, I still feel that pre/post blocks are not in the declarative spirit of Oxidized models. Since you can't place a pre/post between two `cmd` methods, you end up into putting most of the logic into a post/pre block (`fortios.rb`) . In some cases, you can't write the model without a pre/post block at all (`eatonnetwork.rb`), while it is in others it's unnecessary (`linuxgeneric.rb`, `airfiber.rb`).

Taking the junos model as an example:
```ruby
  cmd 'show version' do |cfg|
  (...)
  post do
    out = String.new
    case @model
    when 'mx960'
      out << cmd('show chassis fabric reachability') { |cfg| comment cfg }
    when /^(ex22|ex3[34]|ex4|ex8|qfx)/
      out << cmd('show virtual-chassis') { |cfg| comment cfg }
    when /^srx/
      out << cmd('show chassis cluster status') do |cfg|
        cfg.lines.count <= 1 && cfg.include?("error:") ? String.new : comment(cfg)
      end
    end
    out
  end

  cmd('show chassis hardware') { |cfg| comment cfg }
```

I prefer the declarative alternative:
```ruby
cmd 'show chassis fabric reachability', if: -> { @model == mx960 } do |cfg|
  comment cfg
}

cmd 'show virtual-chassis', if: -> { @model =~ /^(ex22|ex3[34]|ex4|ex8|qfx)/  } do |cfg|
  comment cfg
}

cmd 'show chassis cluster status', if: -> { @model.start_with?("srx") } do |cfg|
  cfg.lines.count <= 1 && cfg.include?("error:") ? String.new : comment(cfg)
}
```

This keeps everything in the expected order (between `cmd 'show version'` and `cmd 'show chassis hardware'`), instead of appending it all at the end.


So, to sum up:
- We don't need the `:if` argument for commands for nxos -- nested `cmd` blocks are the better solution.
- I still believe that the :if argument would be useful for certain models (as in Issue #3680 - fortios rewrite, which is the use-case I had in mind when writing this code).

I propose to split this PR in two:
- Address issue #3657 with nested `cmd` calls
- Create a dedicated PR for the `:if` argument, so we can discuss its value - if it doesn't benefit Oxidized enough, I'm fine leaving it out and keeping `model.rb`simpler.

What do you think?

PS: thank you for your feedback - I was hoping you would take a look at this PR ;-)

robertcheramy:
Update - Fix for NX-OS goes into PR #3684 (without conditional `cmd`).
Depending on @ytti 's feedback on this PR, I will just create a PR with the unit tests for `model.rb` or create a PR with conditional `cmd`.

ytti:
I'm by no means opposing it, it is quite simple and clean so won't add significant maintenance overhead and I'm generally not against multiple ways to achieve same, as use-cases are variant and different solutions are more maintainable for different problems.

robertcheramy:
This PR has been split into #3684 and #3686 and therefore can be closed.

PR #3742, opened 2026-02-07
Create dlinkdgs125052x.rb

vasyna:
add support Dlink DGS-1250-52X Gigabit Ethernet Smart Managed Switch

### DIFF
--- a/lib/oxidized/model/dlinkdgs125052x.rb
+++ b/lib/oxidized/model/dlinkdgs125052x.rb
@@ -16,7 +16,11 @@ class DlinkDgs125052x < Oxidized::Model
     cfg.each_line.to_a[2..-2].map { |line| line.delete("\r").rstrip }.join("\n") + "\n"
   end
 
-  cmd 'show startup-config'
+  cmd 'show vlan' do |cfg|
+    comment cfg
+  end
+
+  cmd 'show running-config'
 
   cfg :telnet do
     username /\r*([\w\s.@()\/:-]+)?([Uu]ser[Nn]ame|[Ll]ogin):/


systeembeheerder:
There is a complete model for [DLINK](https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/dlink.rb). Why do we need a new `dlinkdgs125052x` model? What doesn't work with the `DLINK` model on a DGS125052 ?

vasyna:
> There is a complete model for [DLINK](https://github.com/ytti/oxidized/blob/master/lib/oxidized/model/dlink.rb). Why do we need a new `dlinkdgs125052x` model? What doesn't work with the `DLINK` model on a DGS125052 ?

This switch does not have "show switch" and "show config current". And I deleted "show vlan" because it doesn't seem to me to provide any useful information.


We also need these options for ssh/telnet
```
    post_login 'terminal width 0'
```
otherwise you won't be able to get the output and the logs will show the info
```
INFO -- net.ssh.connection.session[1900]: channel_data: 0 1b
```

systeembeheerder:
> Full CLI support available at end 2019 via firmware update
https://www.dlink.com/uk/en/-/media/business_products/dgs/dgs-1250/dgs_1250_series_datasheet_en_eu.pdf

You're up-to-date?

I doubt oxidized will accept a PR for just one single switch model. Nowhere can i find any clue that this DLINK would have any other OS/cli. It's a 8 year old model, not something fancy/new that oxidized users haven't used before.

vasyna:
> > Full CLI support available at end 2019 via firmware update
> > https://www.dlink.com/uk/en/-/media/business_products/dgs/dgs-1250/dgs_1250_series_datasheet_en_eu.pdf
> 
> You're up-to-date?
> 
> I doubt oxidized will accept a PR for just one single switch model. Nowhere can i find any clue that this DLINK would have any other OS/cli. It's a 8 year old model, not something fancy/new that oxidized users haven't used before.

Yes, updated.

https://support.dlink.com/ProductInfo.aspx?m=DGS-1250%20Series

https://support.dlink.com/resource/PRODUCTS/DGS-1250-SERIES/REVA/DGS-1250%20Series_A1%20A2_CLI%20Manual_v2.04%28US%29.pdf

image

I changed "show startup-config" to "show running-config", but it's probably more important to save the running config than to save the same one for launch.

vasyna:
Sorry, I didn't see DlinkNextGen. Everything works fine with it. Even the SSH connection works without errors.

Issue #1883, opened 2019-08-14
Fortios always changing


Edzilla2000:
On fortios (at least on the fortigate that we manage) oxidized stores the configuration every hour believing it has changed.
It seems that it is caused by fortigate changing the encryption key every time the configuration is exported.

I understand that I can remove the secrets with remove_secret, but that makes the configuration file useless for restoring.
Is there a way to instead ignore secrets when comparing versions?

davama:
How does removing secrets make backup useless? 
Manually configuring 1 line/5 lines in a disaster recovery situation  vs 1000 lines makes it useless?

Edzilla2000:
I used the sentence "useless for restoring", not useless in general.
Right now, if I have a dead firewall, I only have to get the last backup, restore it on a spare firewall and it's back up in less than 10 minutes.
If I strip the secrets, the firewall doesn't accept the config file as is. That means I have to replace the missing secrets with nonsense that meet the firewall's required format, and then manually reconfigure each secret in the running firewall. I have never done that and I'm not actually sure it's possible.


So a interesting compromise in my mind would be for the diff to ignore secrets and to only trigger a change when something else changes.

The only limitation that I see with this is that if the password or certificate actually changes then the backup would not reflect it as oxidized would not trigger a change. 
But in my opinion it is better having a backup with some outdated passwords rather than nothing to restore, or having a 31GB backup folder for 170 devices for a year:
![image](https://user-images.githubusercontent.com/5339038/63023889-5c115880-bea6-11e9-9d68-ab928132690f.png)

davama:
One option (not solution for your problem):

If im not mistaken you can increase the time to daily or weekly to mitigate the diff count.

Edzilla2000:
I could do that, but then I might not have an up to date backup, which kinda negates the whole point of having config backups.

raunz:
As FortiOS reencrypts the secrets every time the "show full-configuration" command is executed, I'm also looking for option to save full configuration, when configuration with removed_secret has been altered. I understand, that change of any native secret via cli/ui could be missed, but thats minor side effect.

I looked at hooks, but I could figure out how to achieve the following scenario:
- start off with default remove_secret for fortios and make git commit of remove_secret_conf
- now a real change of config is done on FortiOS device
- oxidized detect that remove_secret config has been changed
- now I would like to commit to git the config before remove_secret (the native full-config with ENC hashes).
[...]

PR #3743, opened 2026-02-10
Rewrite fortios.rb and split it into fortigate and fortios

robertcheramy:
## Pre-Request Checklist

## Description
Closes issue #3680 

The fortios model has been split into fortigate and fortios. You need the new fortigate model for FortiGate firewalls. Be sure to check the [Fortinet model notes](docs/Model-Notes/Fortinet.md) before upgrading.
- fortios: model rewrite and split into fortios and fortigate.
- fortigate: Add PSU & SFP inventory
- fortigate: move var fortios_autoupdate (deprecated) to fortigate_autoupdate

I plan to merge this PR in one week (after 17.2.2026) if no one objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ fortigate model for FortiGate firewalls. Be sure to check the
 ### Added
 - String refinements: introduce `keep_lines` and `reject_lines` methods (@robertcheramy)
 - Support for storing configurations only on significant changes (@robertcheramy)
+- Add support for Ivanti Secure Connect ISA models (@candleflip)
+- smartbyte: new model for SmartByte switches (@freddy36)
 
 ### Changed
 - Refactored models: Use `keep_lines` and `reject_lines` in aosw, arubainstant, asa, efos, firelinuxos, fsos, ironware, mlnxos and perle to (@robertcheramy)
@@ -20,14 +22,17 @@ fortigate model for FortiGate firewalls. Be sure to check the
 - fortios: model rewrite and split into fortios and fortigate. Fixes #3680 (@robertcheramy)
 - fortigate: Add PSU & SFP inventory (@robertcheramy)
 - fortigate: move var fortios_autoupdate (deprecated) to fortigate_autoupdate (@robertcheramy)
+- netgear: extended login and pager detection to add support for GS728TPv2 and GS752TPv2 (@weberc)
 
 ### Fixed
+- VyOS: Only remove SNMP community, not route-maps. Fixes #3735 (@systeembeheerder)
 - apc_aos: set comment to "; " to match comments in config.ini (@robertcheramy)
 - h3c: fix overly permissive prompt regexp causing false matches. Fixes #3673 (@robertcheramy)
 - extra/device2yaml.rb: fix \r being removed at end of line (@robertcheramy)
 - perle: remove trailing \r (the device sends \r\r\n) (@robertcheramy)
 - Reintroduce support for Ruby 3.0. Fixes #3688 (@robertcheramy)
 - githubrepo: fix authentication with ssh-agent not working. Fixes #3420 (@robertcheramy)
+- fastiron: adjust prompt to account for stacks, remove time from stack output. Fixes #3106 (@ManoftheSea)
 
 
 ## [0.35.0 - 2025-12-04]
--- a/docs/Inputs.md
+++ b/docs/Inputs.md
@@ -177,6 +177,35 @@ input:
     passive: false
 ```
 
+## HTTP
+### Supported HTTP Methods
+
+The HTTP input supports the following HTTP methods:
+- `:get`  - for GET requests
+- `:post` - for POST requests
+
+These methods are used internally by models that require HTTP-based 
+configuration retrieval. Models can use `get_http()` and `post_http()` methods 
+provided by the HTTP input.
+
+Example usage in a model:
+
+```ruby
+cfg :http do
+  post_response = post_http('/some/path', payload, 'Some-Extra-Header' => 'value')
+  get_response  = get_http('/some/path')
+end
+```
+
+HTTP input can be enabled by adding this block to the configuration file:
+
+```yaml
+input:
+  http:
+    scheme: https
+    ssl_verify: true
+    timeout: 30
+```
 
 ## Debugging
 
--- a/docs/Model-Notes/IvantiConnectSecure.md
+++ b/docs/Model-Notes/IvantiConnectSecure.md
@@ -1,0 +1,59 @@
+### Ivanti Connect Secure (ICS)
+
+#### Overview
+
+This model provides support for Ivanti Connect Secure (ICS) appliances using REST API ([official documentation](https://help.ivanti.com/ps/help/en_US/ICS/22.x/22.7R2/22.xICSAG.pdf)).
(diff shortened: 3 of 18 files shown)


raunz:
In the "get system status" section I would like to see the following 3 lines, which weren't excluded in previous fortios model:
```
# License Expiration Date: 2026-09-05
# License Status: Valid
# VM Resources: 1 CPU/1 allowed, 1993 MB RAM
```
___
In my customized fortios model I ignore dynamically generated and constantly changing firewall internet-service-names.
`    cfg.gsub! /(config firewall internet-service-name\n).*?(\nend\n)/m, '\\1 <configuration removed>\\2'`
Coule this be implemented in new fortigate.rb and avoid unnecessary fluctuation. The list in the device gets always anyway updated from Fortinet online service and doesn't need to be backupped at all. 
___
I tested the model with 25 different Fortigate/FortiSwitch device hardware models and all of them reported:
`No PSU serial numbers available`
If the power supply information isn't available, then it would be nice to silently ignore it. For curiosity, which hardware models actually provide that information?
___
Should we ignore the unoccupied ports in  "get system interface transceiver" block and not report undetected interfaces?
```
+# Interface port15 - Transceiver is not detected.
+# Interface port16 - Transceiver is not detected.
```

robertcheramy:
@raunz - Thank you for the feedback, I will implement it.

Could you provide an example of the `config firewall internet-service-name` section, so that I can add it to a YAML simulation file? You need it for fortigate.rb only?

I think the PSU results come with bigger firewalls with field-replaceable PSUs, like the FortiGate 3001F:
https://github.com/ytti/oxidized/blob/61e77d5a13789fb1646e634ecc56b7c8e681aee9/spec/model/data/fortigate%23FortiGate-3001F_7.4.8_HA%23simulation.yaml#L92-L107

robertcheramy:
> Could you provide an example of the `config firewall internet-service-name` section

Sorry, I have it in my configs too.

robertcheramy:
> In the "get system status" section I would like to see the following 3 lines, which weren't excluded in previous fortios model:
> 
> ```
> # License Expiration Date: 2026-09-05
> # License Status: Valid
> # VM Resources: 1 CPU/1 allowed, 1993 MB RAM
> ```

I don't have devices that show these lines. What type of models have them? Is this for the fortigate or the fortios model?

raunz:
Thanks for the power-supply and tranceiver cleanup. That looks great!
My hardware range is from Fortigate 40's series up to 200's.

in my previous comment I forgot to add, that I remove internet-service-name section only within the redacted secrets output:
```
  cmd :secret do |cfg|
    # Remove private key for encrypted configs
    ...
    cfg.gsub! /(set csr ).*?-+END CERTIFICATE REQUEST-+"$/m, '\\1<configuration removed>'
    cfg.gsub! /(config firewall internet-service-name\n).*?(\nend\n)/m, '\\1 <configuration removed>\\2'
    cfg
  end
```
As I still use "double-backup" for Fortigates with hook explained in https://github.com/ytti/oxidized/issues/1883
I keep the redacted version nice and tidy, but he full backup made with hook,  still contains the secrets and also the internet-service-name block.
The current PR removes internet-service-name block from every output and it looks even better after all.


My updated hook to match new fortigate model and related Fortigate config looks like this:
(Hook also now supports per node username/password from router.db).

```
# oxidized/config

hooks:
  fortigatehash:
    type: fortigatehook
    events: [post_store]
    group: "fortigates"
    model: "fortigate"
groups:
  fortigates:
    vars:
      remove_secret: false
models:
  fortigate:
    vars:
      fullconfig: true
      remove_secret: true


# oxidized/hook/fortigatehook.rb

class FortigateHook < Oxidized::Hook
  def validate_cfg!
    raise KeyError, 'hook.group is required' unless cfg.has_key?('group')
    raise KeyError, 'hook.model is required' unless cfg.has_key?('model')
  end

  def run_hook(ctx)
    return unless ctx.node
    return unless ctx.event.to_s == "post_store"
    return unless ctx.node.model.class.name.to_s.downcase == cfg.model

    node = Oxidized::Node.new(name: ctx.node.serialize[:name], ip: ctx.node.serialize[:ip], group: cfg.group, model: cfg.model, output: 'git', username: ctx.node.auth[:username], password: ctx.node.auth[:password])
    gitoutput = node.output.new

    status, config = node.run

    if status == :success
      msg = "update #{node.group}/#{node.name}"
      msg += " from #{node.from}" if node.from
      msg += " with message '#{node.msg}'" if node.msg

      if gitoutput.store node.name, config, msg: msg, email: node.email, user: node.user, group: node.group
        logger.info "Configuration updated in hook for #{node.group}/#{node.name}"
      end
    end
  end
end
```

raunz:
> > In the "get system status" section I would like to see the following 3 lines, which weren't excluded in previous fortios model:
> > ```
> > # License Expiration Date: 2026-09-05
> > # License Status: Valid
> > # VM Resources: 1 CPU/1 allowed, 1993 MB RAM
> > ```
> 
> I don't have devices that show these lines. What type of models have them? Is this for the fortigate or the fortios model?

These 3 lines appear in Virtual Fortigate models FortiGate-VM64-KVM and FortiGate-VM64. e.g. Version: FortiGate-VM64 v7.4.8.
This would be for fortigate.rb model.

robertcheramy:
> I forgot to add, that I remove internet-service-name section only within the redacted secrets output

I understand why you do this, but I don't like the idea of removing something in :secrets which isn't a secret, at least in the official model.
I've moved the lines to (significant changes)[https://github.com/ytti/oxidized/blob/master/docs/Configuration.md#store-configuration-only-on-significant-changes], which may be a simpler alternative to your hook (although slightly different).

raunz:
> I've moved the lines to (significant changes)[https://github.com/ytti/oxidized/blob/master/docs/Configuration.md#store-configuration-only-on-significant-changes], which may be a simpler alternative to your hook (although slightly different).

This is excellent!
In my testing I noticed, to use successfully **output_store_mode: on_significant**, the private keys should also be wiped in  :significant_changes. I copied the "set private-key" line from secrets cleanup to achieve it. To retain coding style, I removed parentheses after "gsub!". 
```
  cmd :significant_changes do |cfg|
    cfg = cfg.reject_lines [
      /^ +set \S+ ENC \S+$/
    ]
    cfg.gsub! /(set private-key ).*?-+END (ENCRYPTED|RSA|OPENSSH) PRIVATE KEY-+\n?"$/m, ''
    cfg.gsub! /(config firewall internet-service-name\n).*?(\nend\n)/m, ''
    cfg
  end
```
After a significant change in configuration, new commit contains also new ENC and private-key hashes. Current test with ~100 nodes is working like a charm now.

robertcheramy:
> To retain coding style, I removed parentheses after "gsub!".

The parentheses are there because the ruby parser can't differentiate between a regexp and a division. It falls back to regexp, so it is OK not to use parentheses and the rubocop warning is deactivated for the models. I prefer using parentheses, but both are OK.

I've added certificate private keys to (un)sinificant changes in the fortigate model.

raunz:
I've tested the latest [b161acb](https://github.com/ytti/oxidized/pull/3743/commits/b161acb5da6ae01e322fe1426eedaca30709c5b2) for a few days and looks good to me. No more objections by me.

Issue #3746, opened 2026-02-20
Fix fetch config at start when interval is 0


tgr229:
Hi,

We use oxidized to backup 20k devices and at each start of the service 20k nodes are added to the queue so its hard to handle this.

So we have modified the `node.rb` and `config.rb` to have the possibility to disable fetch at start. I create PR for this https://github.com/ytti/oxidized/pull/3747

ytti:
I'm bit confused. Can you explain me what the problem is.

My thinking is, that without this, you immediately start all threads that you allow, and all these threads keep working, until you have completed a cycle.

If you never want to have that amount of threads working, wouldn't the solution to be to reduce thread count, and let the box work on the thread count you observe after the fix?

tgr229:
It's too long with 20k devices to process when service is starting. We have 300 threads configured and the duration is 3 hours to complete the backup. We have very slow devices, they can take 30min to 1 hour to backup. So when we have to modify a configuration in `config` file or ajust a model with monkey patch for our needs, 20k devices are process... (we use docker).

At the end, the goal is to disable completely auto backup with interval 0 and disable at start. And we configure fetch backup via /node/next with event from syslog.

ytti:
I still don't follow.

Why do you have 300 threads, if you don't want to launch 300 threads at start? Why not have 1 thread? You can have /next spawn another thread in addition, so you can always call next, while very slowly 1 thread is crawling?

tgr229:
I missed another information, we need to trigger backup via event log AND manually. So 1 thread is very very slow and can take more than one day to backup 20k devices at start. Imagine 900 devices, each take 30-45min. And it's too slow if we trigger backup manually. We need threads to speed up the process, but just we don't want backup at start.

We can't permit backup at start if our container crash or something else.

Other guys need this https://github.com/ytti/oxidized/issues/433

tgr229:
And when I said manually, it's by group, by model. So we trigger backup of 1k device in same time.

ytti:
I'm sorry. We are having communication problem, since I still don't understand the behavioural change that you want with the PR.

How I understand your problem, which is likely misunderstanding
a) Oxidized starts, all devices need to be collected, as they haven't been collected in configured interval
b) Oxidized measures after each thread how many threads are needed to reach interval goal
c) Oxidized spawns too many threads for your liking to hit that interval goal
d) By setting last fetched on nodes you essentially force oxidized into stable state, by pretending all nodes have been fetched

[...]

PR #3747, opened 2026-02-20
Add option disable_fetch_at_start

tgr229:
## Pre-Request Checklist


ytti:
Closing as discussed in #3746 

We want fix the original code introducing support for 0 interval.

Likely implementation will be @flag which will collect once, and reset itself to false. The flag would be set true by calling /next. If interval is 0 and flag is false, collection is skipped.

PR #3752, opened 2026-02-27
Fix fetch config at start when interval is 0

tgr229:
## Pre-Request Checklist

## Description

With interval 0, a fetch continue to be done at service start. Fetch should not be executed when service is starting with interval 0. Interval 0 should disable fetching device.

The PR corrects implementation of interval 0.

Closes issue #3746

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ fortigate model for FortiGate firewalls. Be sure to check the
 - Reintroduce support for Ruby 3.0. Fixes #3688 (@robertcheramy)
 - githubrepo: fix authentication with ssh-agent not working. Fixes #3420 (@robertcheramy)
 - fastiron: adjust prompt to account for stacks, remove time from stack output. Fixes #3106 (@ManoftheSea)
+- interval: fix fetching device configuration at oxidized start when interval is 0. Fixes #3746 (@tgr229)
 
 
 ## [0.35.0 - 2025-12-04]
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -82,7 +82,7 @@ a regular expression.
 username: oxidized
 password: S3cr3tx
 model: junos
-interval: 3600 #interval in seconds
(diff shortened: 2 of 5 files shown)


ytti:
I think the direction is correct. 

I'm bit worried about the self-documentation aspect of calling it @collect. It seems confusing if you're not deeply familiar with this. Because of course we are collecting boxes, despite them being false.

Perhaps more self documenting to tie the variable name to next? @next? Also I think resetting the variable should be reset anyhow, regardless if interval zero is used or not. As we wouldn't know what other uses for it occur, which are not related to interval zero, and in those cases over time, all boxes might be next==true.

```ruby
break if Oxidized.config.interval.zero? && not nextnode.next
nextnode.next = false
```

I'm not making any request, I'm just thinking aloud, I may be way off. I also want @robertcheramy to take a peek. But for sure this or something very much like this will be committed.

tgr229:
Sorry I amend the commit during your comment. I renamed @collect_interval_zero

tgr229:
Changes done

ytti:
Sorry for bike shedding, I am thinking is nexted better or worse?

Problem with next is that obviously next node is next.
Problem with nexted is that it is highly irregular even thought grammatically correct. I can actually find nexted in similar context in other source codes.

Would nexted help or hurt person reading the code?

Also because this is boolean, please create alias to ```nexted?``` and use the ? version on the comparison.  Add yourself to the changelog and ensure the commit has Fixes #nnn. And I think we're good to go.

robertcheramy:
Please first document in this PR description what you are trying to do.
Also update CHANGELOG.md and the documentation under docs/

When you have documented what you are trying do to, we can discuss the name of the variable @collect / @next / whatever. It should be self explanatory (@next is not and is overloaded with ruby next - I'm unhappy with nodes.next, and don't want to overload this more).

ytti:
Please mention the issue and/or PR in the change log.

Otherwise I think we're good to go!

ytti:
I'll fix the rubocop issues.

Issue #3702, opened 2025-12-11
Add SQL-based persistent state management


robertcheramy:
- [X] I have read and understood [docs/Issues.md](https://github.com/ytti/oxidized/blob/master/docs/Issues.md).

**Is your feature request related to a problem? Please describe.**
- Oxidized needs a lot of memory to store its state. On an instance with 5k Devices, the Resident Set Size is 350 MB.
- When using the [cached git history](https://github.com/ytti/oxidized/blob/70d7ea26817ce74bc2c75d02a6f8f9c1961a94d3/lib/oxidized/output/git.rb#L138) to display versions in oxidized-web, the Resident Set size goes up to 1,5 GB (git Size: 600MB).
- Compute intensive operations on oxidized block oxidized-web, and compute intensive operations on oxidized-web block oxidized (probably due to the thread lock). Having a SQL persistent Database could better decouple a front end from the back end.
- @ytti noted this in Issue #812 (first item)
- PR #3666 shows the need for an SQL-based persistent state (not directly usable because raw output of LLM)

**Describe the solution you'd like**
- Oxidized save and reads it states to an SQL backend
- The choice of the SQL backend is flexible
- The git history is cached in the SQL backend
- The SQL backend can be used as an interface to extensions (oxidized-web, Node.js based frontend...)

**Describe alternatives you've considered**
Keep the state in memory.
Cache git logs only to some extend (last 10 Versions...) and only get more with user interaction.

**Additional context**
Setting this to Milestone 1.0.0 and assigning it to myself. I'm not sure how much work this will need.

PR #3760, opened 2026-03-07
perf: parallelize node loading to eliminate startup/reload freeze

garryshtern:
## Pre-Request Checklist

- [x] Tests added or adapted

### Problem

When `resolve_dns` is enabled (the default), `Nodes#load` performs a
synchronous DNS lookup for every node **sequentially**, while holding
`@mutex` the entire time. With large node lists (1000+ devices) this
causes two visible problems:

1. **Startup delay** — oxidized does not accept any web API requests
   until the full initial load completes (minutes with 1300 nodes).
2. **Web UI freezes on reload** — any SIGHUP causes the web API to
   become unresponsive for the full duration of the reload, because
   every API endpoint (`/nodes`, `/node/…`, etc.) competes for the same
   mutex that `load` holds throughout.
3. **Full teardown on every reload** — every SIGHUP discarded all
   runtime state (queue position, last job, stats) and forced a DNS
   lookup and `Node.new` for every device, even those whose source
   configuration had not changed.

### Solution

**`lib/oxidized/nodes.rb`**
- All slow I/O (HTTP source fetch, DNS resolution, `Node.new`
  construction) is moved **outside** the mutex.
- `Node` objects are now built concurrently via a simple `Queue`-based
  thread pool. The pool size is controlled by the new
  `node_load_threads` config option (default: 20).
- Only the final atomic list swap (`replace` / `update_nodes`) is
  performed under the lock, keeping the critical section to
  microseconds.
- **Delta-based reload**: `build_nodes_parallel` snapshots existing
  nodes and compares source opts (`Node#source_opts`). Unchanged nodes
  are reused as-is — no DNS lookup, no object reconstruction. Only
  genuinely added, removed, or config-changed nodes are rebuilt.
  `update_nodes` is rewritten as a proper delta merge: remove deleted
  nodes, reuse identical objects in-place, replace changed nodes
  carrying over `last` job and `stats`, append new nodes, then
  re-sort by schedule priority.

**`lib/oxidized/node.rb`**
- New `source_opts` attribute stores a frozen copy of the raw source
  hash at construction time, used for efficient equality comparison
  during delta reload.

**`lib/oxidized/manager.rb`**
- `Manager#loader` is now guarded by a per-instance `Mutex` with
  double-checked locking so concurrent `Node.new` threads cannot race
  when loading the same model/input/output for the first time.

**`lib/oxidized/core.rb`**
- `Core#reload` now spawns a background `Thread` so the main worker
  loop continues processing jobs while a reload is in progress.
- A `@reloading` guard flag (cleared in `ensure`) prevents concurrent
  duplicate reload threads if multiple SIGHUPs arrive.

**`lib/oxidized/config.rb` / `docs/Configuration.md`**
- New `node_load_threads` option (default: 20) documented.

### Impact

With 1300 nodes and ~50 ms DNS latency, sequential loading took ~65 s.
With 20 threads the same load completes in ~3–4 s. The web API remains
responsive throughout.

On subsequent reloads (SIGHUP), unchanged nodes skip DNS entirely —
only the delta is processed, making incremental reloads near-instant
for stable node lists.

### Tests

24 tests, 59 assertions, 0 failures:
- `Nodes#load`: populates list, full reload replace, preserves
  last/stats on unchanged reload, skips bad nodes (ModelNotFound,
  ResolvError), continues after single failure, node_want filter,
  parallel multi-node correctness, source order preservation.
- Delta cases: reuses same object when opts unchanged, replaces object
  when opts change, carries over last/stats on config change, removes
  disappeared nodes, adds new nodes.
- Concurrency: mutex not held during source fetch, multiple concurrent
  readers, atomic swap.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ fortigate model for FortiGate firewalls. Be sure to check the
 
 ### Changed
 - Node list reload (SIGHUP) now runs in a background thread so the worker loop and web API remain responsive during reload. (@shtern)
+- Delta-based node reload: unchanged nodes are reused without DNS lookup or re-construction; only added, removed, or config-changed nodes are rebuilt. Runtime state (last job, stats) is preserved across reloads. (@shtern)
 - Refactored models: Use `keep_lines` and `reject_lines` in aosw, arubainstant, asa, efos, firelinuxos, fsos, ironware, mlnxos and perle to (@robertcheramy)
 - Refactor SSH and SCP into a common class SSHBase. Fixes #3597 (@robertcheramy)
 - Modified models to support store mode on significant changes: ios, fortios, perle (@robertcheramy)
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -7,13 +7,15 @@ module Oxidized
   class Node
     include SemanticLogger::Loggable
 
-    attr_reader :name, :ip, :model, :input, :output, :group, :auth, :prompt, :timeout, :vars, :last, :repo
+    attr_reader :name, :ip, :model, :input, :output, :group, :auth, :prompt, :timeout, :vars, :last, :repo,
+                :source_opts
     attr_accessor :running, :user, :email, :msg, :from, :stats, :retry, :err_type, :err_reason, :nexted
     alias running? running
     alias nexted? nexted
 
     # opt is a hash with the node parameters given in the source (:name, :group, :ip...)
     def initialize(opt)
+      @source_opts = opt.dup.freeze
       logger.debug 'resolving DNS for %s...' % opt[:name]
       # remove the prefix if an IP Address is provided with one as IPAddr converts it to a network address.
       ip_addr, = opt[:ip].to_s.split("/")
--- a/lib/oxidized/nodes.rb
+++ b/lib/oxidized/nodes.rb
@@ -146,9 +146,15 @@ module Oxidized
 
     # Constructs Node objects from +candidates+ (an Array of option hashes) using
     # a thread pool so that DNS lookups run concurrently instead of serially.
+    # Nodes whose source options are identical to the existing Node object are
+    # reused as-is — no DNS lookup, no object reconstruction.
     # Results preserve the original source ordering.
     def build_nodes_parallel(candidates)
       return [] if candidates.empty?
+
+      # Snapshot existing nodes outside the mutex for delta comparison.
+      # A stale read here is harmless: at worst we reconstruct a node unnecessarily.
+      existing = each_with_object({}) { |n, h| h[n.name] = n }
 
       num_threads = [Oxidized.config.node_load_threads || 20, candidates.size].min
       results     = []
@@ -160,14 +166,22 @@ module Oxidized
         Thread.new do
           loop do
             i, node = queue.pop(true)
-            begin
-              built = Node.new(node)
-              results_mu.synchronize { results << [i, built] }
-            rescue ModelNotFound => e
-              logger.error "node %s raised %s with message '%s'" % [node, e.class, e.message]
-            rescue Resolv::ResolvError => e
-              logger.error "node %s is not resolvable, raised %s with message '%s'" % [node, e.class, e.message]
-            end
+            existing_node = existing[node[:name]]
+            built = if existing_node&.source_opts == node
+                      existing_node # unchanged: reuse without DNS or reconstruction
+                    else
+                      begin
+                        Node.new(node)
+                      rescue ModelNotFound => e
+                        logger.error "node %s raised %s with message '%s'" % [node, e.class, e.message]
+                        nil
+                      rescue Resolv::ResolvError => e
+                        logger.error "node %s is not resolvable, raised %s with message '%s'" \
+                                     % [node, e.class, e.message]
+                        nil
+                      end
+                    end
+            results_mu.synchronize { results << [i, built] } unless built.nil?
           rescue ThreadError
             break # queue is empty
           end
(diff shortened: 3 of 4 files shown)


ytti:
This is worthwhile feature. 

However the contribution comes from someone with very little to no GitHub history of contributing. It would likely take me about as long time to review it well as it would take me to write it myself with Claude Code, and the latter is preferable as I have some idea how I want to refactor the code to make it cleaner as well as I would combine the work with moving non-running nodes from memory to on-disk sqlite. 

I could consider suggesting non-trivial commit from someone who has decent history of usefully contributing to projects. But right now I'm not seeing that, and I just don't have sufficient trust in Claude. 

I am very unlikely to accept this PR.

garryshtern:
You have done a great job with this project, and you are sole decision maker on what makes sense or doesn't to accept, andhow a contribution should be composed.

Having said that, do you really believe that being condescending is the best way to encourage contributions?

ytti:
Condescension wasn’t my intention.

What options do I have in face of LLM? It truly does generate faster than it takes to review it.

1. Review it
2. Rewrite it yourself with LLM
3. Trust

Are there more options? Which should I choose?

garryshtern:
The way I work with it for reviews is similar to how I used to do it myself. Clearly lay out criteria of I am checking for, what the guidelines are, what tests would cover this functionality. And then analyze the response.

robertcheramy:
I review every PR with the quality of Oxidized in mind, and I invest some time in them.
When I know the code is generated by a LLM, I am particularly careful, because a don't know how much the author of the PR understands the code. Thank you @garryshtern for having indicated that this PR has been generated by Claude, I was pretty sure when reading the description, you confirmed it.

That being said, I think the root of the problem is that oxidized-web is too tightly coupled with oxidized. Working on threaded DNS Resolution treats the symptom, not the cause. Oxidized needing 60 seconds to make its DNS resolution is probably not a problem if it doesn't block oxidized-web. And DNS resolution *may* possibly better be done when fetching the node, not when reading the sources.

I'll postpone this PR for 0.37 for now, and will evaluate what to do with it when I have made plans for #3702.

ytti:
> I'll postpone this PR for 0.37 for now, and will evaluate what to do with it when I have made plans for #3702.

We should not accept this. The problem is real and should be addressed. But if we are going to rework this, it should be included in the work where non-running nodes state is moved to persistent SQLite. 

Solving just this, without moving non-running state to disk would risk producing solution which is not conducive to fixing the SQLite part of the problem. 

The solution should consider both issues simultaneously. I have fairly good idea how it should look like since I never intended current design to be long term, but I've found myself too lazy to fix it. 

LLM can implement this, but it cannot refactor it the 'right way', not without someone having clear idea what the clean implementation looks like. 

Anyone wanting to tackle this shouldn't produce any code at all, they should first explain the design they're thinking of, before moving forward with implementation. The design would use Sequel to represent the nodes, it would have transactions and locking where needed, it would support schema migrations. It would separate concerns well, DB implementation would be in its own files and business logic would usually contain just 1 line method call towards DB class instance, meaning DB class implements all possible complexity related to data handling, keeping call-sites clean.

garryshtern:
## Design Proposal: SQLite-backed node state with deferred DNS

Thank you for the detailed feedback. I understand the concern: fixing DNS latency at load time treats the symptom, not the cause. The right fix addresses both problems simultaneously — persistent state and deferred DNS — via a Sequel-backed `NodeStore`. Here is the design I would propose before writing a single line of code.

---

### Root problems (as I understand them)

1. `Nodes < Array` is simultaneously the scheduler queue, the config store, and the runtime state store. There is no separation of concerns.
2. All runtime state (`last` job, `stats`, resolved IP) is ephemeral — lost on restart or reload.
3. DNS resolution happens at load time, blocking everything, because `Node#initialize` is the only place that knows a node's name and needs an IP.
4. oxidized-web competes with the worker on the same Ruby `Mutex`, because the only way to read node state is through the in-memory `Nodes` array.

---

### Proposed architecture

#### New layer: `lib/oxidized/db/`

```
lib/oxidized/db/
  schema.rb       — Sequel migrations (versioned, forward-only)
  node_store.rb   — owns all data-handling complexity; caller sites are 1-line calls
```

`NodeStore` is the single source of truth for persistent state. It wraps a Sequel connection (SQLite by default, any Sequel-supported backend by config). All locking that today happens via `@mutex` in `Nodes` is replaced by Sequel transactions.

#### Schema (draft)

```sql
CREATE TABLE nodes (
  id               INTEGER PRIMARY KEY,
  name             TEXT    NOT NULL UNIQUE,
  ip               TEXT,                   -- resolved, cached
  ip_resolved_at   DATETIME,               -- TTL anchor for re-resolution
  group_name       TEXT,
  model            TEXT,
  source_opts_json TEXT,                   -- JSON of raw source hash (delta key)
  active           BOOLEAN NOT NULL DEFAULT 1,
  -- runtime state (survives restart)
  last_start       DATETIME,
  last_end         DATETIME,
  last_status      TEXT,
  last_duration    REAL,
  -- stats as JSON ring buffer (avoids a second table for now)
  stats_json       TEXT
);
```

Migrations are versioned in `schema.rb` using `Sequel.migration` blocks. Oxidized runs `Sequel::Migrator.run` on startup — safe to call repeatedly.

#### `NodeStore` responsibilities (all complexity lives here)

```ruby
# Every public method is the entire call-site in business logic

NodeStore#sync_from_source(candidates)    # delta upsert inside a transaction
NodeStore#update_job_result(name, job)    # persist last/stats after a fetch
NodeStore#next_due                        # return name of next node to poll
NodeStore#resolved_ip(name)              # return cached IP or nil (TTL check)
NodeStore#store_ip(name, ip)             # write resolved IP + timestamp
NodeStore#serialize_all                  # for /nodes API response
NodeStore#serialize_one(name)            # for /node API response
```

`NodeStore` never returns rich `Node` objects to callers — it returns plain hashes or Sequel row objects. Business logic does not touch SQL.

#### `Node` becomes a connection-only object

`Node` keeps only what is needed to establish a connection and run a model: `name`, `ip`, `model`, `input`, `output`, `auth`, `vars`, `prompt`, `timeout`. It no longer holds `@last`, `@stats`, `@retry`, `@running`, or `@nexted` as mutable state — those all move to the DB.

DNS resolution moves out of `initialize`. Instead:

```ruby
def ip
  @ip ||= begin
    cached = Oxidized.node_store.resolved_ip(name)
    cached || resolve_and_cache
  end
end
```

`resolve_and_cache` calls `Resolv` and writes the result back to the DB. This means DNS is deferred to first actual fetch, runs on a worker thread, and the result is cached with a TTL. No DNS at load time at all.

#### `Nodes` becomes a thin orchestrator

```ruby
def load(node_want = nil)
  candidates = fetch_source(node_want)
  Oxidized.node_store.sync_from_source(candidates)   # 1-line DB call
  # Nodes array stays in memory only as the scheduler queue;
  # it holds lightweight Node objects built from the DB, not from source opts
  replace build_queue_from_store
end
```

The scheduler queue (the `Array`) holds lightweight `Node` objects. Their mutable scheduling state (`running`, `nexted`, queue position) remains in-process because it changes at sub-millisecond frequency and does not need persistence. Everything else is in the DB.

#### `Worker` becomes thinner

```ruby
def process(job)
  node = job.node
  Oxidized.node_store.update_job_result(node.name, job)  # 1-line DB call
  node.running = false
  job.status == :success ? process_success(node, job) : process_failure(node, job)
end
```

#### oxidized-web decoupling

oxidized-web's read-heavy endpoints (`/nodes`, `/node/:name`, `/version`) can query `NodeStore` directly (or via a read-only Sequel connection to the same SQLite file). They never need to acquire the worker's mutex. The only shared in-process state is the scheduler queue, and that is only written by the worker thread.

---

### What this solves

| Problem | Current | Proposed |
|---|---|---|
| DNS at load time | Blocks for minutes | Deferred to fetch; cached in DB with TTL |
| State lost on restart | Yes | No — DB persists across restarts |
| State lost on reload | Partial (our delta fix) | No — reload = DB sync, no object teardown |
| oxidized-web lock contention | Competes on `@mutex` | Reads DB directly, no shared mutex |
| Memory (350 MB / 5k nodes) | All state in RAM | Runtime state in SQLite; RAM holds queue only |

---

### What I am NOT proposing to implement without further alignment

- The git history cache in SQLite (mentioned in #3702) — that is a separate concern
- Any specific Sequel adapter beyond SQLite — the adapter is a config option
- Any API changes in oxidized-web — that is a follow-on

---

### Questions before moving forward

1. Is `sqlite` the right default, or should it be an in-process SQLite with an option to point at PostgreSQL/MySQL?
2. Should the scheduler queue (in-process Array) be fully replaced by a DB-ordered query, or kept in-process for performance?
3. Is there a preference on where the DB file lives (e.g., alongside the config, or XDG data dir)?

Happy to iterate on this design before writing any code.

robertcheramy:
Please stop producing LLM bloat, I don't have time to review LLM produced data, I can produce it myself with my own prompts.

I'm closing this PR as it doesn't bring us further. I will address persistent state in #3702, and I won't introduce thread-based DNS resolution because of oxidized-web blocking. This is a design + refactoring issue, and LLM can't solve this.

PR #3762, opened 2026-03-09
Dynamic Model Selection

alexnet123:
This PR introduces a new feature: **dynamic model selection** via `model_rules`.

When `model_rules_enable: true` is set in the configuration, Oxidized evaluates a list of rules to determine the appropriate model for each node. Each rule can match on any device field (e.g., `vendor`, `group`, `type`, `ip`, etc.) and assign a specific model. This is particularly useful when nodes are sourced from external systems like NetBox or CSV, which provide rich metadata.

Rules are processed in order; the first matching rule wins. If no rule matches, the traditional model resolution (explicit `model` field, group setting, or global `model`) is used as a fallback.

Documentation has been updated in `docs/Configuration.md` and a CHANGELOG entry has been added.

Closes # (if applicable, otherwise remove this line)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ fortigate model for FortiGate firewalls. Be sure to check the
 - Support for storing configurations only on significant changes (@robertcheramy)
 - Add support for Ivanti Secure Connect ISA models (@candleflip)
 - smartbyte: new model for SmartByte switches (@freddy36)
+- Added `model_rules` for dynamic model selection based on device attributes (e.g., vendor, group, type, IP). Controlled by `model_rules_enable` flag. (@alexnet123)
+
 
 ### Changed
 - Refactored models: Use `keep_lines` and `reject_lines` in aosw, arubainstant, asa, efos, firelinuxos, fsos, ironware, mlnxos and perle to (@robertcheramy)


ytti:
I'm skeptical of the use-case. Your first example is just vendor mapping, which already is done.
Your second example is vendor+ip exception, and I have to wonder, why would you want this? Why wouldn't you then in your single source of truth expose this information for all consumers? Why should only oxidized know about this? It seems like potentially anti pattern. 

I'm also not entirely comfortable with the code quality or idea that we should implement our DSL on top of YAML syntax. If this should exist, I think plugin would be more preferable, we have other use cases for plugin based solution.

However I am unlikely to accept PR even with plugin, unless it comes from someone with some GitHub history of contributing meaningfully to projects. As generating it on LLM myself the way I want, would be less work that thorough review from stranger with no history.

alexnet123:
@ytti, thank you for the feedback.

The main reason I proposed this feature is that **I am not a NetBox administrator** and cannot add a `model` field to every device. However, I can leverage the existing data (vendor, device type, group) already present in NetBox. `model_rules` allows me to assign the correct Oxidized model based on those fields without modifying the source. This is useful for users who don't control the source of truth but still want to automate backups based on available metadata.

The IP example was just for illustration – the real value is in mapping by vendor and device type.

If you think a plugin would be a better approach, I'm open to discussing it. Otherwise, if this doesn't fit Oxidized's direction, I'll close the PR. No hard feelings.

Thank you for your time and for maintaining Oxidized.

ytti:
https://netboxlabs.com/docs/netbox/models/dcim/platform/ - is this relevant?

We’ve had another issue where user wants to conditionally use different data field from sources depending on other data. The wanted sometimes to use oob_ip sometimes mgmt_ip depending on other key/value information.

Likely other more exotic needs exist, and instead of solving all these in YAML, we could call plugin when creating nodes where the plugin can manipulate what information node ends up having. This approach would solve your problem, the other users problem and likely any other future need for logic in node creation.

alexnet123:
@ytti, thank you for the explanation. I understand that a built-in YAML solution isn't the right direction. I'll close this PR for now and wait for the plugin system you mentioned. Thanks again for your time.

ytti:
https://github.com/ytti/oxidized/tree/feat-srchook

This represents the direction I am looking for. I'm not saying this actually works, as I didn't test any of it. But if you can work towards solution like this or ensure this works, I'm very likely to accept the PR .

robertcheramy:
Note: Definitely LLM Generated Code.

I don't think we should include the complexity into oxidized. If you need to modify the source coming from Netbox, put an intelligent http proxy (a simple python script will do) in between. Ask your favourite LLM something like: "please generate an http python proxy that converts a csv into another csv to transport model+ip into model_ip". Maybe you'll be happy with the code.

@ytti - you reopened the PR, so I let you choose the best thing to do with it ;-)

ytti:
@robertcheramy i already indicated to @alexnet123 that they are not only one needing complex logic to load nodes. Another poster wanted sometimes to use oob_ip sometimes mgmg_ip from source.

I LLMd direction that this should be, which is hook in sources, which satisfy both users use cases and any future ones. 

Either of the users should test it, I’m happy to work on it based on feedback. 

The use cases are valid and well addressed by hooks.

ytti:
@alexnet123 I'm bit confused, are you saying that you don't currently have time to test the plugin based system to solve your problem?

alexnet123:
I tried to look into the plugin/hooks-based approach, but I couldn’t get it working on my side — it’s a bit too complex for me at the moment.
That’s why I went with a simpler solution for now.

ytti:
I’m happy to address the problems, if you have time to tell what issues you ran into, I’m happy work to address them.

alexnet123:
I'm experimenting with your `feat-srchook` branch to implement dynamic model selection via a hook (`model_rules`). However, the hook doesn't seem to be triggered – I see no debug output, and nodes still get the global `junos` model.

**Setup:**
- Branch: `feat-srchook` (latest as of today)
- Source: HTTP (NetBox API) with a `hooks` section inside the source config.
- Hook file: `/etc/oxidized/hooks/model_rules.rb` (mounted via Docker volume).
- The hook file is present and readable inside the container.

**What I've tried:**
- Verified the hook file is in place (`ls -l` inside container).
- Added `$stderr.puts` at the top of the hook file and inside the `source_node_transform` method – no output in logs.

**Attached config:**
```yaml
username: oxidized_ssh_user
password: oxidized_ssh_password

resolve_dns: false
interval: 3600
rest: 0.0.0.0:8888

log: /home/oxidized/.config/oxidized/logs/oxidized.log
debug: true

next_adds_job: false

output:
  default: git
  git:
    user: Oxidized
    email: oxidized@example.com
    repo: "/home/oxidized/.config/oxidized/oxidized.git"

source:
  default: http
  http:
    url: http://192.168.0.120:8000/api/dcim/devices/?status=active&has_primary_ip=true
    scheme: http
    delimiter: !ruby/regexp /:/

    headers:
      Authorization: Token XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX

    map:
      name: name
      vendor: device_type.manufacturer.name
      type: device_type.model
      group: role.slug
      ip: primary_ip.address

    secure: false
    hosts_location: results
    pagination: true
    pagination_key_name: next

    hooks:
      - name: model_rules
        type: ruby
        events: [source_node_transform]
        file: /etc/oxidized/hooks/model_rules.rb
        rules:
          - vendor: Mikrotik
            model: routeros
          - vendor: Cisco
            group: switch
            model: ios
          - vendor: Mikrotik
            group: switch
            ip: 192.168.0.180/24
            model: eltex

groups:
  ro_ud: {}
  ro_bd: {}
  switch: {}
```

Any idea why the hook isn't being called? Thanks for your help!

ytti:
The hooks isnt array, it is hash with the hook name as key. 
I’d also like to see the script if further problems.

alexnet123:
@ytti, I fixed the config format (changed array to hash), but the hook is still not being called – no debug output appears in the logs. Here is the hook script I'm using (`/etc/oxidized/hooks/model_rules.rb`):

```ruby
$stderr.puts ">>> ModelRulesHook: LOADED"

def source_node_transform(ctx)
  $stderr.puts ">>> source_node_transform called for #{ctx.node[:name] || ctx.node['name']}"
  
  rules = config[:rules] || []
  node = ctx.node

  matched_model = nil
  rules.each_with_index do |rule, idx|
    match = true
    rule.each do |key, value|
      next if key == 'model'
      node_value = node[key.to_sym] || node[key.to_s]
      if node_value.to_s.strip.downcase != value.to_s.strip.downcase
        match = false
        break
      end
    end
    if match
      matched_model = rule['model']
      break
    end
  end

  if matched_model
    node = node.merge(model: matched_model)
  end
  node
end
```

The logs show no `>>>` messages at all. The hook file is present and readable inside the container (`/etc/oxidized/hooks/model_rules.rb`). Any idea why it's not being triggered? Could it be that the `source_node_transform` event isn't wired for HTTP sources yet? Thanks for your help!

ytti:
The hooks is top level key, not under sources. 

Further instead of config[:rules] it is cfg.rules. But it actually doesn't work right now, I'll fix it if this at least 1 person will use this solution. Right now, ignore the `cfg.rules` and just use rules right in the ruby code.

Issue #3334, opened 2024-12-03
Allow models to use multiple inputs


robertcheramy:
In some situations, you may need to get a config per scp / ftp and execute some commands per ssh to get the inventory or serial number.

In the case APC devices (Uninterruptible Power Supply - model apc_aos.rb), I want `config.ini` per ftp/scp and execute following commands per ssh:
```
about
upsabout
```

robertcheramy:
Notes:
- This may imply to merge scp.rb into ssh.rb, see input from @ytti in https://github.com/ytti/oxidized/commit/0bfddc0f0804faf3d9bfbe8ae8c0ee694ce99257#r162861892
- I'm thinking of adding `download` oder `get_file` in parallel to `cmd` to differentiate between (ssh) commands and downloading files.
- This may mean refactoring code if we really mean mutliple inputs like ssh + ftp. Currently, one input module is chosen in Oxidized::Node.run_input. My Idea is to allow to download a file with scp/sftp within the ssh module (any maybe download a file within http, I haven't looked at it), but not to mix inputs.

ytti:
```cmd``` does take parameters, I'm thinking we configure >1 input methods in model and then in the config block we give them execution order. If execution order is unset it is executed first. First execution order input is also default for ```cmd``` if no input is not specified.


```ruby
# this gets via scp
cmd "/config/plop"
# this logins via ssh 
cmd "about" input: ssh

cfg :scp
cfg :ssh do
  order 2
end
```

tgr229:
Hi,

I'm interested by this issue. In my case, I need to execute a script after ssh execution to complete device configuration : I trigger a backup ftp device through input ssh to upload file on an ftp gateway then the gateway upload the config device to oxidized on filesystem. I need to fetch the config device on filesystem and add it to the config device in Oxidized

robertcheramy:
Closed by PR #3763

PR #3763, opened 2026-03-13
Add multiple inputs to oxidized + ApcAos with SSH + SCP capatibilities

robertcheramy:
## Pre-Request Checklist

## Description
Add multiple inputs to oxidized 
New ApcAos model with SSH + SCP capabilities.
The SCP gem is now optional because it is rarely used (as TFTP).
APC: update model notes
model: move DSL definition into separate modules

Closes Issue #3334


ytti:
This looks fine to me. Perhaps while refactoring model.rb, we should separate the actual models from the the infrastructure. One approach could be:

```
model.rb
model/dsl...
models/ios.rb
```

or maybe model_abstract/x is less work and breakage than plural for models. 

I kind of always regretted putting model.rb along the device models. Moving the model.rb file up would be entirely fine, but then it would be less clear what is obvious place for the additional files it now has. 

Adding more non-device models in that directory feels worst option. 


I guess cleanest would be to move all of manager assets into pluralised directories, and for now load both singular and plural version, and later drop singular. And keep actual concrete input/output/source/model/hook in pluralised, and the abstract in singular. Not sure if worth the chore, since it's really just model where it might become realistically issue the more we split model.rb into separate files.

robertcheramy:
Thanks to unit tests, such a change can be done without too much fear breaking things.

I didn't changed the path for local models (`~/.config/oxidized/model`), but it would be the logical consequence of the move. I suppose this is what you mend with "load both singular and plural version"? For the official oxidized models in the gem we don't need to support both names.

robertcheramy:
I didn't change the hooks either - it doesn't belong into this PR and it can wait until there is a bigger change there.

ytti:
I'm not asking to include any of this work into this commit, just thinking aloud while splitting model.rb into multiple files. 

Some optionality here

```shell
model.rb
model/addtitions_to_model.rb
models/ios.rb
```

```shell
model/model.rb
model/additions_to_model.rb
models/ios.rb
``` 

```shell
model/model.rb
model/additions_to_model.rb
model/models/ios.rb
``` 

 I don't feel too strongly, and clearly model is most important, since risk of collision is highest. But of course if we change this for model, we should change it for everything that manager manages. 

And indeed we don't need to support plural/singular for gem files, but likely user files in manager we need to load from plural.chop and plural, and possibly logger.warn if plural.chop loads.

robertcheramy:
In d4f630383cd0a19e4ec797b70de1423686d47da2, I opted for
```
model.rb
model/addtitions_to_model.rb
models/ios.rb
```

if we want to make this for other modules, we need to change a lot of things. Current tree (including commit d4f630383cd0a19e4ec797b70de1423686d47da2):
```
lib/oxidized
- hook.rb
+ hook/    <- only hooks
+ input/  <- input.rb, cli.rb and sshbase.rb AND inputs
- model.rb
+ model/
+ models
+ output/ <- output.rb AND outputs
+ source/ <- source.rb AND sources
(and more dirs/files)
```

on the local directory side, manager.rb uses the singlular version (Example: https://github.com/ytti/oxidized/blob/27e4c07a4cffc136c7f2e2d293269796046aa89a/lib/oxidized/manager.rb#L41):
```
~/.config/oxidized/
+ hook/
+ input/
+ output/
+ source/
+ model/
```

We have at least four options:

1. use singular in the gem and the config / status quo (This PR before d4f630383cd0a19e4ec797b70de1423686d47da2)
2.  use plural for models in the gem and singular in the config (This PR so far)
3.  use plural for models in the gem and the config, support singular in the config but warn with a deprecation
4.  use plural for al modules (inputs, sources, outputs, models, hooks) in the gem and the config, support singular in the config but warn with a deprecation. 
   When at it, use a proper structure/naming convention for module/class paths (`Oxidized::Hook::GithubRepo instead of `GithubRepo`). This will break every custom model and hook.

Option 3 feels wrong, 2 feels dirty.
Option 4 might be the cleanest any will pay some technical debt back, but it will break a lot of things for our users. This may be mitigated in a clever migration strategy.

I tent to revert d4f630383cd0a19e4ec797b70de1423686d47da2 and postpone 4.

What do you think?

ytti:
I share same sentiment. 

I’m not too worried of breaking. Directory migration seems straightforward, namespace trickier.

One approach to namespace is:

The abstract class still available via old namespace. The old namespace links to the new namespace, emits warning and also assigns the top level name to the right namespace. This should all be possible by having the old namespace abstract class define inherited method.

robertcheramy:
> :warning: Warning: force-pushing to feat/3334-multiple-inputs to revert d4f630383cd0a19e4ec797b70de1423686d47da2

robertcheramy:
The structure change is postponed to release 0.37.0 and tracked in Issue #3765.

PR #3766, opened 2026-03-17
exalink model add

obol89:
## Pre-Request Checklist

## Description

Adding very specific Cisco model for basic support. It has been confirmed working with Cisco Nexus 3550-F which was formerly Exablaze and it is running custom Cisco / Exalink OS.


lib/oxidized/model/exalink.rb
```diff
@@ -0,0 +1,43 @@
+class Exalink < Oxidized::Model
+  using Refinements
+
+  PROMPT = /^([\w.@()-]+[#>]\s?)$/
+
+  prompt PROMPT
+  comment '! '
+
+  def filter(cfg)
```
robertcheramy: I suggest you use [cmd :all](https://github.com/ytti/oxidized/blob/master/docs/Ruby-API.md#cmd) and [cut_tail](https://github.com/ytti/oxidized/blob/master/docs/Ruby-API.md#cut_tail) or `cut_both`. Also see https://github.com/ytti/oxidized/blob/master/docs/Creating-Models.md#typical-tasks-and-solutions

This way, there is no need for saving PROMPT in an extra variable.

robertcheramy (changes requested):
Please update CHANGELOG.md

PR #3796, opened 2026-04-09
Full configuration support for modern TrueNAS SCALE devices

### LINKED ISSUE #3795: TrueNAS SCALE gaps in configuration and constant updates
- [X] I have read and understood [docs/Issues.md](https://github.com/ytti/oxidized/blob/master/docs/Issues.md).

**Describe the bug**
* TrueNAS SCALE supports App configurations (docker containers running on the system), but the model doesn't capture their configurations.
* TrueNAS SCALE updates the configuration of the CloudSync tasks every time they run with the updated status of the last run, so it changes regularly despite no configuration changes.
* TrueNAS CORE and SCALE both used a similar behavior for ZFS replication configurations, so the model just ignored all those configurations to keep them from being too noisy.

**To Reproduce**
1. Configure a TrueNAS SCALE model to be scraped by oxidized for configuration.
2. Inspect configuration, see that apps and replication configurations aren't present.
3. Run a CloudSync operation, and then see that it's "configuration" has changed because the last-run output was added to the configuration database.

**Expected behavior**
1. I expect that apps configured on the NAS and their configurations are captured by Oxidized.
2. I expect that replication and CloudSync configurations are captured, but only change when actually reconfigured.

**Running environment (please complete the following information):**
- Manufacturer model an software version: TrueNAS SCALE 25.04.2.3
- oxidized model name: truenas.rb

**Additional context**
I will submit a PR soon with the updated model fixing these bugs.

neilschelly:
This is an update to the model configuration for TrueNAS devices (in particular for SCALE generation). Fixes #3795.

## Pre-Request Checklist

## Description
Most of the configuration is stored in a configuration database in sqlite. It used to ignore the storage_replication table because that included "status updates" from the last run of a replication that changed regularly. That same patter was also recently adopted for the configuration of CloudSync tasks, causing them to appear "changed" every time the job ran.

Those tables will now be captured without the status update columns now, so that they are captured without the noise.

Modern TrueNAS SCALE machines also support "Apps" configurations that can manage some docker containers running on the device, and those configurations are not present in the configuration database. They are a filesystem with YAML files describing docker-compose environments instead. Those will now be captured for a more complete configuration snapshot of the device.

Closes issue #3795


robertcheramy:
I think this update will break previous configurations, as the sudoers has to be changed. Could you warn the Users about in in CHANGELOG.md after having resolved the conflict?

neilschelly:
It shouldn't break existing configurations. It will attempt getting the configuration without sudo first, just like it used to. If that works, it will wok with no changes to the sudoers file.

Anyone already using this will not have to make any changes. I _am_ using this already on older devices that have always worked without sudo, and it's still working fine on them.

lib/oxidized/model/truenas.rb
```diff
@@ -9,19 +9,70 @@ class TrueNAS < Oxidized::Model
...
+      cfg = cmd("sudo sqlite3 -readonly 'file:/data/freenas-v1.db' .dump")
     end
     cfg.lines.reject do |line|
-      line.match(/^INSERT INTO storage_replication /) ||
+      line.match(/^INSERT INTO storage_replication /) || # ignore storage_replication because repl_status field changes on every run
         line.match(/^INSERT INTO system_alert /) || # ignore system alerts in db
-        line.match(/^INSERT INTO sqlite_sequence VALUES\('system_alert',/) # ignore system alerts in db
+        line.match(/^INSERT INTO sqlite_sequence VALUES\('system_alert',/) || # ignore system alerts in db
+        line.match(/^INSERT INTO tasks_cloudsync /) # ignore cloudsync tasks because job field changes on every run
     end.join
```
robertcheramy, line 24: Not asking for a change - using cfg.reject_lines (https://github.com/ytti/oxidized/blob/master/docs/Creating-Models.md#keep-or-remove-lines-returned-from-a-command) is easier - maybe for next time.

Issue #3793, opened 2026-04-07
0.36.0 broke Vyos SNMP Community String detection


nicolasberens:
- [x] I have read and understood [docs/Issues.md](https://github.com/ytti/oxidized/blob/master/docs/Issues.md).

**Describe the bug**
This commit https://github.com/ytti/oxidized/commit/0519fb38326ccc3332ffd7932f8b54e368ead73f changed the snmp community string detection.

This breaks the detection for SNMP traps since their syntax is:

`set service snmp trap-target 1.2.3.4 community communitystring`

**To Reproduce**
1. Add SNMP Trap to vyos router.
2. Run oxidized
3. SNMP Community is not removed

**Expected behavior**
Community string gets redacted.

**Configuration**
```
set service snmp trap-target 1.2.3.4 community communitystring
```

**Logs**
```
If applicable, add logs to help explain your problem.
```

**Running environment (please complete the following information):**

- OS: [e.g. Debian Bookworm, official container version xxx, ...]
- oxidized version: [e.g. 0.32.2]
- oxidized-web version: [e.g. 0.15.1, if applicable]
- Manufacturer model an software version:
- oxidized model name:

**Additional context**

nicolasberens:
fixed in https://github.com/ytti/oxidized/pull/3799

PR #3799, opened 2026-04-14
Add secret detection support for SNMP traps

nicolasberens:
## Pre-Request Checklist

## Description

Closes Issue #3793 

This Change adds support to detect SNMP community strings in the Vyos config.

It also extends the tests so this regression will get caught in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - input/ssh: validate that cmd is a String. See #3700 (@robertcheramy)
 
 ### Fixed
-
+- VyOS: detect community string in SNMP traps. Fixes: #3793 (@nicolasberens)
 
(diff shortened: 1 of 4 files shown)


ytti:
Can you update changelog also please.

Additionally, if you wish, you can add unit test for the secret, so that there won't be regression.

nicolasberens:
Updated Changelog.

I added the tests in the first commit.

ytti:
What I meant, is testing for feature you added:
https://github.com/ytti/oxidized/blob/master/docs/ModelUnitTests.md#secrets

This is not mandatory, just if you want to ensure no regression.

nicolasberens:
Oh, maybe i am understanding this wrong.

I thought that this is already covered by: https://github.com/ytti/oxidized/blob/master/spec/model/data/vyos%23Supermicro_1.4.3%23secret.yaml#L2 since it uses the same "secret"

ytti:
Yes, I think you are right. Less precision in our testing, but should stop regression.

PR #3813, opened 2026-05-05
Additional models, for additional Palo Alto models

athompson-merlin:
## Pre-Request Checklist
Not a dev, so no idea what rubocop or rake are.  (Obviously I can guess, but never seen them before.)  Learning one iota more about Ruby than I need to is *wildly* out of scope for me, but I still wanted to upstream these models if possible.

## Description
These three new models address two use cases the original Palo Alto models did not, (1) obtaining an XML output without needing to use the API, and (2) obtaining output in two different formats from the Panorama management server, which has slightly different syntax.  The "XML" format can be used (with some work) to restore a firewall, but is largely un-parseable; the "set" format is more useful for every other scenario as it's parseable and pasteable.


robertcheramy:
Closing stale PR. Doesn't pass CI Checks.

Issue #1378, opened 2018-06-11
Import from Zabbix via API


Misairu:
Trying to import list of hosts from Zabbix with API.
Script returns those json code:
```
{
	"jsonrpc":"2.0",
	"result": [
		{
		"hostid":"12041",
		"host":"sw_dlink",
		"inventory": 
			{
				"hostid":"12041",
				"model":"dlink",
				"tag":"admin",
				"asset_tag":"secretpassword",
				"notes":"enable"
			},
		"interfaces":[
			{
				"ip":"10.10.10.10"
			}]
		}
	],
	"id":1
}
```
"Source" part of oxidized config looks like this:
```
source:
  default: http
  debug: true
  http:
    url: http://127.0.0.1/zabbix-host-get-api.php
    scheme: http
    delimiter: !ruby/regexp /:/
    map:
      name: host
      model: inventory.model
      ip: interfaces.ip
      username: inventory.tag
      password: inventory.asset_tag
    vars_map:
      enable: inventory.notes
```
But it crushes with message
`no implicit conversion of String into Integer [TypeError]`

Tried to map all variables with adding prefix "result." but getted same error.
Need your help

ytti:
Your interfaces is an array, we don't have array navigation support. So what it is complaining is you're trying to navigate array interfaces with key 'ip'.

Misairu:
It's not a complaint, just a feedback and an ask. Cause I have a problem and " you're my only hope".
Is it possible to make this supportnes of array navigation? Both services has API and I really do not want come to the Dark side and use database connection.

ytti:
Try this branch https://github.com/ytti/oxidized/tree/feature-string_navigation

ytti:
If you can't test, or it doesn't work, can you open sanitised API access for me. I'd come from 91.198.120.1.

ytti:
I suspect I need to further need to add ```http.device_root``` navigation key, to describe where the device array exists in the JSON.

ytti:
You probably should do something like

```
http:
  hosts_location: result
  map:
    ip: interfaces[0].ip
```

Misairu:
Thank you a lot for your support 
**feature-string_navigation** works great with importing.
But now prompt detection failed.
```
I, [2018-06-13T15:44:31.867347 #21350]  INFO -- : Oxidized starting, running as pid 21350
I, [2018-06-13T15:44:31.891462 #21350]  INFO -- : lib/oxidized/nodes.rb: Loading nodes
[...]

PR #3818, opened 2026-05-10
Experimental Zabbix integration

ziotibia81:
## Pre-Request Checklist

## Description
This is an early-stage integration between Zabbix and Oxidized that enables dynamic device discovery and configuration backup management. It works but is not yet battle-tested in large production environments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - routeros: support store mode `on_significant` (@infabo)
 - model for Grandstream HT8xx (@mklopocki)
 - Add --support option to gather system diagnostics (@robertcheramy)
+- Documentation and scripts for an early-stage integration between Zabbix and Oxidized
 
 ### Changed
 - input/ssh: validate that cmd is a String. See #3700 (@robertcheramy)
--- a/extra/zabbix_integration/README.md
+++ b/extra/zabbix_integration/README.md
@@ -226,7 +226,7 @@ curl -s 'http://127.0.0.1/oxidized-api/get-nodes.php?token=YOUR_TOKEN' | jq .
 # [
 #   {
 #     "name": "switch-01",
-#     "model": "dxs",
+#     "model": "dlinknextgen",
 #     "group": "DXS",
 #     "ip": "192.168.1.100"
 #   },
--- a/extra/zabbix_integration/README.md
+++ b/extra/zabbix_integration/README.md
@@ -1,0 +1,540 @@
+# Zabbix-Oxidized Integration
+
+**Status: Experimental/Embryonic** ⚠️
+
+This is an early-stage integration between Zabbix and Oxidized that enables dynamic device discovery and configuration backup management. It works but is not yet battle-tested in large production environments.
+
+## Overview
+
+This integration allows you to:
+- **Dynamically discover** network devices in Oxidized based on Zabbix host tags
+- **Manage backup strategies** through Zabbix host inventory and device groups
+- **Monitor backup status** with Zabbix alerts and historical data
+- **Reload device lists** automatically without restarting Oxidized
+
+### How It Works
+
+```
+┌─────────────────────────┐
+│  Zabbix Database        │
+│  (Host inventory,       │
+│   Tags, Credentials)    │
+└────────────┬────────────┘
+             │
+             ▼
+┌─────────────────────────┐
+│  get-nodes.php          │
+│  (Zabbix API Client)    │
+└────────────┬────────────┘
+             │
+             ▼
+┌─────────────────────────┐
+│  Oxidized Web API       │
+│  (http://localhost:8888)│
+└────────────┬────────────┘
+             │
+             ▼
+┌─────────────────────────┐
+│  Device Configs         │
+│  (Git Repository)       │
+└─────────────────────────┘
+```
+
+## Requirements
+
+### System Requirements
+- **Single server deployment**: Oxidized and Zabbix must run on **the same machine**
+  - Both services accessible via `127.0.0.1`
+  - No remote Zabbix API access (token-based only)
+- **OS**: Linux (Ubuntu/Debian recommended)
+- **PHP**: 7.4+ with `file_get_contents()` enabled
+- **Apache**: 2.4+ with mod_proxy enabled
+- **Oxidized**: Latest stable version with `oxidized-web` gem installed
+
+### Software Versions Tested
+- Zabbix 6.0 LTS
+- Oxidized 0.29.x
+- Apache 2.4.x
+- PHP 7.4, 8.0, 8.1
+
+## Installation
+
+### Step 1: Create API Directory Structure
+
+```bash
+# Create the directory for the PHP API script
+sudo mkdir -p /usr/local/share/oxidized-api
+
+# Ensure proper ownership (adjust user/group if different)
+sudo chown www-data:www-data /usr/local/share/oxidized-api
+sudo chmod 755 /usr/local/share/oxidized-api
+```
+
+### Step 2: Install the PHP API Script
+
+```bash
+# Copy the get-nodes.php script
+sudo cp get-nodes.php /usr/local/share/oxidized-api/
+
+# Secure the script
+sudo chmod 640 /usr/local/share/oxidized-api/get-nodes.php
+sudo chown www-data:www-data /usr/local/share/oxidized-api/get-nodes.php
+```
+
+### Step 3: Configure Apache
+
+```bash
+# Copy the Apache configuration
+sudo cp oxidized-get.conf /etc/apache2/conf-available/
+
+# Enable the configuration
+sudo a2enconf oxidized-get
+
+# Verify configuration syntax (important!)
+sudo apache2ctl configtest
+# Should output: Syntax OK
+
+# Create the password file for web UI access
+# You'll be prompted to enter a password
+sudo htpasswd -c /etc/apache2/.oxidized_pwd oxidized
+
+# Reload Apache
+sudo systemctl reload apache2
+```
+
+### Step 4: Configure Oxidized
+
+```bash
+# Copy the Oxidized configuration
+sudo cp config /home/oxidized/.config/oxidized/config
+
+# Adjust permissions
+sudo chown oxidized:oxidized /home/oxidized/.config/oxidized/config
+sudo chmod 600 /home/oxidized/.config/oxidized/config
+
+# **IMPORTANT**: Edit the configuration and replace:
+# - `ZabbixTokenHere` with your actual Zabbix API token (see Step 6)
+# - Default credentials to match your environment
+sudo nano /home/oxidized/.config/oxidized/config
+```
+
+Minimum configuration changes:
+```yaml
+url: http://127.0.0.1/oxidized-api/get-nodes.php?token=YOUR_ZABBIX_TOKEN_HERE
+username: oxidized
+password: your_device_password
+```
+
+### Step 5: Set Up Automatic Node Reload
+
+Add to the **oxidized** user's crontab to reload the device list every hour:
+
+```bash
+# Edit oxidized user's crontab
+sudo -u oxidized crontab -e
+
+# Add this line:
+0 * * * * /usr/bin/curl -s http://127.0.0.1:8888/oxidized/reload.json > /dev/null
+```
+
+**Why is this needed?**
+- Without this cron job, Oxidized only reads the device list at startup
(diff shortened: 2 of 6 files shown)


ytti:
Can you explain why you chose this approach?

For example

- does input/http.rb not work for this case, why not?
- if input/http does not work, and cannot be fixed reasonably, why not input/zabbix.rb

ziotibia81:
Hi! Thanks for the feedback.

The reason behind this approach is mainly related to the specific requirements of the Zabbix API and my current technical constraints:

1. POST Request & Filtering: To get the required data from Zabbix, I needed to send a specific JSON filter via POST. I couldn't find a straightforward way to handle this custom POST logic using the standard input/http.rb without modifying the core. If there are any solution it isn't explained in documentation.
2. Data Transformation: The JSON returned by Zabbix needs some restructuring (especially regarding group mapping) to be compatible with Oxidized. Handling this logic inside a Ruby module was quite a challenge for me.
3. Ruby Knowledge: I am not a Ruby developer, so modifying the existing HTTP input or writing a native zabbix.rb module was currently out of my reach.
4. Speed & Community: I needed a working solution quickly and, since I found many users asking for Zabbix integration in various forums, I decided to share my PHP-based approach even if it's "external".

I’ve labeled this as Experimental/Embryonic in the README for this very reason. It’s a bridge solution that works, and I hope it can either help someone immediately or serve as a logical base for someone with better Ruby skills to port it into a native module.

ytti:
It is very good that you were able to get your integration working, but I don’t think this is right direction for the project nor should we promote this as solution seeing as our answers for why not modify http.rb or create source/zabbix.rb are not well researched, and are very likely the solutions we’d offer should demand arise.

ytti:
https://github.com/ytti/oxidized/issues/1378

People seem to be using http.rb as-is with zabbix api

ziotibia81:
I’m sorry to disappoint you, but if you look closely at #1378, that approach is actually identical to mine: it also relies on an external PHP script to fetch and format the data before Oxidized can consume it. It does not use the native HTTP module to handle the Zabbix API logic directly. The Zabbix API interface is `api_jsonrpc.php`, the issue call `zabbix-host-get-api.php`.

This confirms that, currently, a direct integration using only input/http.rb is not as straightforward as it seems, which is why multiple people have ended up building this kind of bridge.

Anyway, I respect your decision to keep the codebase clean. My fork remains available for anyone who needs this specific workaround.

ytti:
If you want to expose some lab zabbix environment for my consumption, I can either change source/http.rb to support it, or introduce source/zabbix.rb if changing http.rb doesn't seem reasonable. 

If you are willing to do this, please also tell me what is expectation from the API, that is your zabbix instance returns something, I want to also know what nodes data you would expect to be returned.

ziotibia81:
I really appreciate your openness to look into this! Unfortunately, I cannot expose my production Zabbix environment for security reasons, but I can provide you with everything you need to reproduce the logic: the exact API call, the required payload, and the expected output format.

Zabbix uses JSON-RPC 2.0. To fetch hosts, you must send a POST request to `http://zabbix-web-url/[prefix]/api_jsonrpc.php` with a JSON body. A simple GET won't work.

This is the Oxidized request: 

```
GET /oxidized-api/get-nodes.php?token=zabbix_api_token_here
HTTP/1.1
Accept-Encoding: gzip;q=1.0,deflate;q=0.6,identity;q=0.3
Accept: */*
User-Agent: Ruby
Connection: close
Host: 127.0.0.1
```

This is that came from my php script:
```
POST /zabbix/api_jsonrpc.php HTTP/1.1
Host: 127.0.0.1
Connection: close
Content-Length: 300
Content-type: application/json

{
   "jsonrpc":"2.0",
   "method":"host.get",
   "params":{
      "output":[
         "host"
      ],
      "selectInventory":[
         "software_app_a",
         "model"
      ],
      "selectInterfaces":[
         "ip"
      ],
      "filter":{
         "status":"0"
      },
      "tags":[
         {
            "tag":"Backup",
            "value":"Oxidized",
            "operator":1
         }
      ]
   },
   "auth":"zabbix_api_token_here",
   "id":1
}
```

I only fetch hosts that have the tag `Backup: Oxidized`. This is crucial for large Zabbix environments where you don't want to back up every single monitored asset.

I'm pulling `software_app_a` and `model` from the Zabbix Inventory. In my setup `software_app_a` countains the Oxidized model to use. `model` is the the item model, in my script I use some logic to map models and oxodozed groups (see my README). It coult also be possibile write directly the group in `software_app_b`, and use  `default` if empty.  The `ip` must be extracted from the `selectInterfaces` array.

To make this integration robust, a native Zabbix source should ideally allow users to define in the Oxidized config which inventory fields map to which Oxidized variables.

Zabbix offers a wide range of inventory fields (documented here: [Zabbix 6.0 Host Object](https://www.zabbix.com/documentation/6.0/en/manual/api/reference/host/object)). A dedicated source could look like this in the config:

```yaml
source:
  default: zabbix
  zabbix:
    url: https://zabbix.example.com/api_jsonrpc.php
    token: zabbix_api_token
    tag: "Backup"
    tag_value: "Oxidized"
    map:
      model: software_app_a
      group: software_app_b
```
Here te Zabbix api response:

```json
{
   "jsonrpc":"2.0",
   "result":[
      {
         "hostid":"10341",
         "host":"switch-1.local.lan",
         "inventory":{
            "software_app_a":"dlinknextgen",
            "model":""
         },
         "interfaces":[
            {
               "ip":"192.168.12.5"
            }
         ]
      },
      {
         "hostid":"10327",
         "host":"switch-2.local.lan",
         "inventory":{
            "software_app_a":"dlinknextgen",
            "model":""
         },
         "interfaces":[
            {
               "ip":"192.168.12.3"
            }
         ]
      },
      {
         "hostid":"10304",
         "host":"switch-3.local.lan",
         "inventory":{
            "software_app_a":"dlinknextgen",
            "model":""
         },
         "interfaces":[
            {
               "ip":"192.168.12.8"
            }
         ]
      },
      {
         "hostid":"10296",
         "host":"switch-4.local.lann",
         "inventory":{
            "software_app_a":"dlinknextgen",
            "model":""
         },
         "interfaces":[
            {
               "ip":"192.168.12.6"
            }
         ]
      },
      {
         "hostid":"10279",
         "host":"switch-5.local.lan",
         "inventory":{
            "software_app_a":"dlinknextgen",
            "model":""
         },
         "interfaces":[
            {
               "ip":"192.168.12.7"
            }
         ]
      },
      {
         "hostid":"10283",
         "host":"switch-6.local.lan",
         "inventory":{
            "software_app_a":"dlinknextgen",
            "model":""
         },
         "interfaces":[
            {
               "ip":"192.168.12.1"
            }
         ]
      },
      {
         "hostid":"10315",
         "host":"switch-7.local.lan",
         "inventory":{
            "software_app_a":"dlinknextgen",
            "model":"DXS-1210-16TC"
         },
         "interfaces":[
            {
               "ip":"192.168.12.9"
            }
         ]
      },
      {
         "hostid":"10316",
         "host":"switch-8.local.lan",
         "inventory":{
            "software_app_a":"dlinknextgen",
            "model":"DXS-1210-16TC"
         },
         "interfaces":[
            {
               "ip":"192.168.12.10"
            }
         ]
      }
   ],
   "id":1
}
```

Edit
Someone wrote a [zabbix.rb](https://gist.github.com/SerWax/e4a9f89985c9e6754966534aa07c16a4#file-zabbix-rb), it filters by template id. I prefer the tag based filter approach.

ytti:
I feel like filter should be configurable, to cover all practical uses cases. All of this is very hard to get right without access to real instance with real data. 

Should should opportunity arise, I’m happy to work on this, in absence of access, we can continue with solutions people have.

ytti:
Another option if no one can provide me access to zabbix is that that @SerWax upstreams their implementation, and possibly adds ability to configure filtering to satisfy your use-case too.

ziotibia81:
Please have a look here: [zabbix.rb](https://github.com/ziotibia81/oxidized/blob/zabbix-integration-native/lib/oxidized/source/zabbix.rb)

It works for me. If it's ok I will update documentation and I will try a PR.

ytti:
I will accept it, even though the code quality is unsatisfactory, the @SerWax version is very reasonable. But next person unhappy about it, who has access to zabbix can rewrite it. 

- You should run 'rubocop' against it, as it will fail CI.
- You should include documentation and changelog
- You don't have to add any testing, but I will accept complete rewrite from next person, and without testing, no telling if there will be regression and no attempt will be made to guarantee no-regression. 

Just couple quick things I noticed, but I'm not asking for a changes, as the problem isn't these specifically, problem is broader. You can't LLM your way out of this, this would require human caring, which is a tall order.

- You don't need 'respond_to?', we have @cfg.has_key?(x) and @cfg.x? Applies to map/cfg_str, ssl_verify and elsewhere
- @cfg.foo == true to get boolean of it, feels bizarre, like this is not misunderstanding ruby, this feels misunderstanding programming
- you can use .map(&:method(arg)) if you plan to call just single method on every element
- hash (not cfg) has #fetch, so. you can do hash.fetch(key, default) when you want either key value, or default value if no key exists
- not sure your type coercion to string makes sense, like why test 'x.empty?' if you can test 'unless x', if we expect x to be nil/false or string?
- pick_address seems quite awful 
- if else end can be done by condition ? true_branch : false_branch

SerWax:
Sorry for the delayed response. I've submitted a pull request not just for my original implementation, but for an updated one that includes filtering by tags.

My gist was never published because I don't consider the implementation to be production worthy, as it requires you to place plaintext credentials in Zabbix.

It should only be implemented if your Zabbix instance is well secured. That said the .rb itself poses no risk, and it is under the each admin's responsibility to understand how credentials are secured.

ziotibia81:
@SerWax you are absolutely amazing! 🥇
Please consider my approach regarding the device password: leave the password optional on the Zabbix side, use groups, and specify the group passwords in the Oxidized config file.

SerWax:
In my environments devices tend to have different credentials so it has made more sense to be able to set them zabbix side
Setting them in Oxidised if possible should be a better approach

I'll try to think of something in the next days

ziotibia81:
You can check the logic in my botched implementation here: [zabbix.rb](https://github.com/ziotibia81/oxidized/blob/zabbix-integration-native/lib/oxidized/source/zabbix.rb)

The logic itself works, but since I don't know Ruby, I hacked it together.

What it does:

-Filters by tag
-Groups and Models by inventory field

No password logic, so it's hadled by groups in the Oxidized configuration file.

ytti:
@ziotibia81 it is not clear to me, if this pull request would solve your problem as well. And I won't merge it until both your and @SerWax use-cases are covered, with hopes that this will increase probability that it covers also the the next person's zabbix use-case. 

If I understood correctly, your use-case is already supported. You can send "nil" password from zabbix side, which will fall back to groups eventually. But I may be confused.

ziotibia81:
Hi @ytti,

Many thanks for your interest. @SerWax's implementation and mine actually use different approaches. Let me try to describe the differences (@SerWax, please correct me if I miss or misrepresent anything).

**@ziotibia81's  Approach**
In my implementation, device passwords are never stored inside the Zabbix database; they are kept exclusively in the Oxidized configuration file. The Oxidized model and group are stored in Zabbix Inventory fields using a configurable mapping. This approach executes only one call to the Zabbix API, filtering hosts via Zabbix tags.

**@SerWax's Approach**
@SerWax's approach fetches all hosts by filtering either by Zabbix template or Zabbix tag (it is not entirely clear to me if it filters only by tag name, or by both tag name and value like mine). After that, it executes:

- Another API request to retrieve the IP and hostname for the previously found hosts.
- A call to retrieve all parent templates for each template discovered.
- A final API call to get all Zabbix macros defined for each host, which is then used to build the node list (including passwords) for Oxidized.

To summarize, my approach is very straightforward, while @SerWax's code is more complex and tightly integrates the Zabbix templates associated with the hosts.

Because of these fundamental differences, the two approaches are not entirely compatible out of the box. To cover both use cases, the two code logics would need to be merged together.

ytti:
Sorry I don't follow. Can you not build node list from @SerWax's implementation? Mind you, you can set password in Zabbix "nil", at which point Oxidized will continue to resolution, and will use your group config for password. 

Certainly if you cannot build node list, we shouldn't merge, since perhaps then other Zabbix users have same problem.

ziotibia81:
@SerWax uses Zabbix macros to store Oxidized model an group, I use inventory filelds.
@SerWax filters host by template name or tag (it is not entirely clear to me if he filters only by tag name, or by both tag name and value), I filter using configurable tag_name and tag_value (Ex: tag_name = Backup AND tag_value = Oxidized).

I don't know if @SerWax code works setting a 'nil' password in Zabbix macro. I should try...

PR #3838, opened 2026-06-03
Add Zabbix source with Tags support

SerWax:
## Pre-Request Checklist

## Description

Closes issue #3818 and fixes #3837

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ source:
       model:    "{$OXIDIZED_MODEL}"
     vars_map:
       group:  "{$OXIDIZED_GROUP}"
-#      enable: "{$OXIDIZED_ENABLE}"
+      enable: "{$OXIDIZED_ENABLE}"
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
...
+    map:
+      username: "{$OXIDIZED_USERNAME}"
+      password: "{$OXIDIZED_PASSWORD}"
+      model:    "{$OXIDIZED_MODEL}"
+    vars_map:
+      group:  "{$OXIDIZED_GROUP}"
+      enable: "{$OXIDIZED_ENABLE}"
+
+- Update Zabbix source to implement filtering by tags are requested by other users.
+
```
robertcheramy, line 26: Write details in docs/ not in the CHANGELOG.

robertcheramy:
@ytti - I haven't looked in Detail in this PR. as I think you were involved in specific sources for Zabbix. Do you have an opinion on it ?

ytti:
@ziotibia81 has not confirmed if it addresses their use-case. There seems to be some disagreement if node discovery offered satisfies both users. 

I think having Zabbix support would be good, but it may be difficult to find right implementation without multiple zabbix users confirming their use-case is addressed.

Issue #437, opened 2016-05-26
F5 login into tmsh mode instead of bash shell


joshrain:
when i login to an f5 device i am given the tmsh shell instead of bash.

the tmos.rb model expects a bash shell.

i tried to modify tmos.rb to first run the following cmd:

cmd('run /util bash') { |cfg| comment cfg }  ## this is so it enters bash mode and run the remaining cmds but it hangs.

ytti:
I don't have any F5 device, so I can't really comment. Maybe @mikebryant can chime in.

mikebryant:
It's not something we've tested - we just configured the F5s to allow bash shell logins for the collector user.

ytti:
It would be terrible tax to pay, as we cannot run ```exec``` mode then, as we won't know which type of commands to send. Unless of we make some ```tmos_bash: false``` variable.

I would discourage this.

ytti:
@joshrain is it acceptable to make bash user?

nowell29morris:
This is a bit of a stretch, but what if calling bash was used like an enable password?  Correct me if I am wrong, but setting yourself to enable mode is like calling a new shell inside of your first shell.  Why not call bash?

I am facing this same difficulty with F5s on 11.X and priv users are domain driven.  I will have to give this a shot, and unfortunately go through about four people to ask them to set the shell of my user to bash in the AD unix utils.

ytti:
We can definitely do it, it just means 'exec' mode won't, and we need to do screen scraping. 
As long as it works for @joshrain @mikebryant and @nowell29morris I'm happy with it.

If it were my device, I'd make oxidized users shell bash, to avoid evils screen scraping.

nowell29morris:
The problem we are facing here is a design decision F5 made for this version.  You can either have privileged accounts locally or via the domain, but not both.  When you use a domain account, privileged or not, you cannot specify bash as your shell like you could in a traditional /etc/passwd sense.  Maybe we make tmos.rb and tmos11.rb so that in these specific instance/version of tmos you can sacrifice 'enable' for bash.  Or am I requesting too much at this point? :)

ytti:
I get that. I don't think we need two models. I'm ok changing this into screen-scraping, after which it's easy to enter bash shell, like you explained.

matt-wells-m451:
I've taken over this item for nowell29morris.  First thanks for all the help on this one.  
What action items can I do to assist on this one?

clay584:
[...]

PR #3839, opened 2026-06-08
Separate model for TMOS when using AAA

martinberg:
## Pre-Request Checklist

## Description
Modify existing tmos.rb as a new model to be able to run bash commands.
This is needed when using AAA on a F5 device due to it not allowing bash shell for a remote user.
See #437 for earlier Oxidized issue and https://my.f5.com/manage/s/article/K28660000 for F5 docs.

I have not ran any tests or code analysis but can try to do so if this seems like a viable approach.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - routeros: support store mode `on_significant` (@infabo)
 - model for Grandstream HT8xx (@mklopocki)
 - Add --support option to gather system diagnostics (@robertcheramy)
+- Add `tmos_aaa` model for F5 TMOS devices using AAA authentication
 
 ### Changed
 - input/ssh: validate that cmd is a String. See #3700 (@robertcheramy)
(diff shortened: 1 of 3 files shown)


robertcheramy:
Missing CHANGELOG and Supported OS changes.

robertcheramy:
This PR is a near-complete duplication of tmos.rb — the two files differ only in that each command is wrapped in `run /util bash -c "..."`. That should be solvable directly in tmos.rb without a second model.

Two problems with the current approach:

- Duplication: ~60 lines copied with a single mechanical difference.
- Class name collision: tmos_aaa.rb also declares class TMOS < Oxidized::Model. Since the constant is the same, the two definitions would clash depending on load order.

A cleaner approach [not tested]: gate the AAA behaviour behind a vars flag and build the commands at runtime inside a post do, wrapping each command dynamically. This is exactly the pattern routeros.rb uses to build its /export command based on vars(:remove_secret) and version:

```ruby
post do
  wrap = ->(c) { vars(:tmos_use_aaa) ? %(run /util bash -c "#{c}") : c }

  cmd(wrap.call('tmsh -q show sys version')) { |cfg| comment cfg }
  # ... remaining commands, each passed through wrap.call ...
end
```

Users then just set:

```YAML
vars:
  tmos_use_aaa: true
```

That keeps everything in a single tmos.rb, no duplicated code, and no separate model to maintain.

PR #3840, opened 2026-06-08
procurve: add model spec, fix secret redaction, simplify the model, rstrip comments

thanegill:
## Pre-Request Checklist

## Description
- spec: queue the init prompt in `MockSsh`
- spec: add procurve model unit test
- procurve: fix snmp-server host secret redaction, redact local user password hashes (possibly fixes #3774)
- procurve: rstrip commented sections
- models: replace slice and rstrip with `cut_*` and `rstrip_lines` refinement
- docs: add @thanegill as a procurve model maintainer, note that some additional HPE Aruba switches use the procurve model
- procurve: use clean `:escape_codes` and simplify the model

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 - procurve: rstrip trailing whitespace from commented command sections; add `String#rstrip_lines` refinement (@thanegill)
+- procurve: remove ANSI escape codes with `clean :escape_codes` and simplify the prompt regexp (@thanegill)
 
 ### Fixed
 - procurve: hide the SNMP community on `snmp-server host <ip> community "<name>"` lines (@thanegill)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 ### Added
+- procurve: model unit test (@thanegill)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Changed
 
 ### Fixed
+- procurve: hide the SNMP community on `snmp-server host <ip> community "<name>"` lines (@thanegill)
+- procurve: hide local user password hashes (`password ... sha1 "<hash>"`) when remove_secret is true (@thanegill)
 
 ## [0.37.0 - 2026-05-20]
 ### Added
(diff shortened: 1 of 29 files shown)


ytti:
You don't define rstrip_cfg method?

You probably intended something like:
```ruby
comment cfg.each_line.map(&:rstrip).join("\n") + "\n"
```

But then we might ask, why not add this to :all. block?

thanegill:
> You don't define rstrip_cfg method?

Oops, I thought that was defined in refinements. Added, copied from `aosw.rb`.

thanegill:
> But then we might ask, why not add this to :all. block?

We could. But I'm not certain there is any training whitespace that is needed to restore the config accurately.

ytti:
Maybe rather something like 

```ruby
def comment_clean str
  comment str.each_line.map(&:rstrip).join("\n") + "\n"
end
```

So that you're not repeating yourself that much. 

Also, if you can, please provide spec/model/data files for testing purposes. I'm uncomfortable as this was submitted with non-working code, so I don't know if it works at all, or if it breaks for everyone.

thanegill:
Added the spec model data for tests. Found a few other bugs along the way. Let me know if this would be better suited as multiple PRs.

lib/oxidized/model/procurve.rb
```diff
@@ -1,26 +1,26 @@
...
   using Refinements
 
-  # previous command is repeated followed by "\eE", which sometimes ends up on last line
-  # ssh switches prompt may start with \r, followed by the prompt itself, regex ([\w\s.-]+[#>] ), which ends the line
-  # telnet switches may start with various vt100 control characters, regex (\e\[24;[0-9][hH]), followed by the prompt, followed
-  # by at least 3 other vt100 characters
-  prompt /(^\r|\e\[24;[0-9][hH])?([\w\s.-]+[#>] )($|(\e\[24;[0-9][0-9]?[hH]){3})/
+  # The prompt is the device name followed by '#' or '>' and a
+  # space, optionally preceded by a carriage return on ssh.
+  prompt /(^\r)?([\w\s.-]+[#>] )$/
```
robertcheramy, line 6: I think placing ^ outside the parenthesis (making them unnecessary) would improve our prompt safety:
```ruby
prompt /^\r?([\w\s.-]+[#>] )$/
```
I tested against our simulation files - works.

What do you think?

CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 ### Added
+- procurve: model unit test (@thanegill)
+
+### Changed
+- procurve: rstrip trailing whitespace from commented command sections; add `String#rstrip_lines` refinement (@thanegill)
```
robertcheramy, line 12: The add rstrip_lines refinement should be in a separate line, and documented under [docs/Creating-Models.md]](https://github.com/ytti/oxidized/blob/master/docs/Ruby-API.md#refinements)

robertcheramy: And the updated models could be named (even if the code is equivalent)

robertcheramy:
Thank you for this well written PR. Please have a look at my comments, then I'll be happy to merge it into master.

PR #3849, opened 2026-06-13
fsos: add 'cmd :all' block to strip carriage returns and command echo

tlbeaty:
## Pre-Request Checklist

## Description

Add `cmd :all` block to strip carriage returns and command echo/prompts
from device output. Without it, `\r` characters were appearing in saved
configs and command echo/prompts were leaking into the output, causing
noisy diffs on every backup.

Tested against FS.com S5860-48SC and S5860-20SQ switches.

### DIFF
--- a/lib/oxidized/model/fsos.rb
+++ b/lib/oxidized/model/fsos.rb
@@ -10,6 +10,11 @@ class FSOS < Oxidized::Model
     data.sub re, ''
   end
 
+  cmd :all do |cfg|
+    cfg = cfg.delete("\r")
+    cfg.cut_both
+  end
+
   cmd :secret do |cfg|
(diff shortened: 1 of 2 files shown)


lib/oxidized/model/fsos.rb
```diff
@@ -10,6 +10,11 @@ class FSOS < Oxidized::Model
     data.sub re, ''
   end
 
+  cmd :all do |cfg|
+    cfg = cfg.delete("\r")
+    cfg.cut_both
```
robertcheramy, line 15: Does this conflict with cut_head in `cmd 'show running-config'?

Can you provide a Simulation file, so that we see what cut_both will change in the model behaviour?

PR #3860, opened 2026-07-03
Added the ability to explicitly specify the port on which the telnet, ssh is running for each host

rajven:
Added the ability to explicitly specify the port on which the telnet  or ssh service is running for each host, i.e.

in config:
source:
  default: csv
  csv:
     file: "/home/oxidized/.config/oxidized/router.db"
     delimiter: !ruby/regexp /\:/
     map:
       name: 0
       ip: 1
       username: 2
       password: 3
       model: 4
added:       input_port: 5
       input: 6
       group: 7
     vars_map:
       enable: 8
       comware_cmdline: 9

in router.db:
router:192.168.1.1:admin:password:routeros:60022:ssh:Gateways::

## Pre-Request Checklist

### DIFF
--- a/lib/oxidized/input/sshbase.rb
+++ b/lib/oxidized/input/sshbase.rb
@@ -28,6 +28,7 @@ module Oxidized
     end
 
     def make_ssh_opts
+      node_ssh_port = @node.input_port || vars(:ssh_port)
       ssh_opts = {
         number_of_password_prompts:      0,
         keepalive:                       vars(:ssh_no_keepalive) ? false : true,
@@ -35,7 +36,7 @@ module Oxidized
         append_all_supported_algorithms: true,
         password:                        @node.auth[:password],
         timeout:                         @node.timeout,
-        port:                            (vars(:ssh_port) || 22).to_i,
+        port: (node_ssh_port || 22).to_i,
         forward_agent:                   false
       }
 
--- a/lib/oxidized/input/telnet.rb
+++ b/lib/oxidized/input/telnet.rb
@@ -12,7 +12,7 @@ module Oxidized
 
       @text_debug = DebugText.new(Oxidized.config.input.debug, @node, config_name)
 
-      port = vars(:telnet_port) || 23
+      port = @node.input_port || vars(:telnet_port) || 23
 
       telnet_opts = {
         'Host'    => @node.ip,
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -7,7 +7,7 @@ module Oxidized
   class Node
     include SemanticLogger::Loggable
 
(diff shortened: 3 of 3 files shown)


ytti:
You can of course already do this for SSH, Telnet per node. And collapsing it as input port for node suggests we guarantee ssh and telnet and other inputs all listen on same port, as same node way traverse multiple inputs, this is not a safe assumption.

rajven:
> You can of course already do this for SSH, Telnet per node. And collapsing it as input port for node suggests we guarantee ssh and telnet and other inputs all listen on same port, as same node way traverse multiple inputs, this is not a safe assumption.

That's the whole point — so that for a specific equipment, you can explicitly specify the port on which SSH or Telnet is running.

ytti:
I dont follow. When would you need this? Why wont specifying telnet or ssh port as needed work?

rajven:
Well, look - I have a group of devices, for example mikrotik, for which ssh runs on a non-standard port, for example 5122. And all the other mikrotik are on standard port 22. What should I do in this situation?

ytti:
You set ssh_port for the 5122 device, using vars_map.

rajven:
Is it possible to do this selectively on the device?

ytti:
Technically depends on source, in CVS you point to an index which has to exist, in SQL you point to a row which has to exist, but even then it can return falsy value. If it returns falsy value, default 22 will be used. 

You can further customise what is returned between source and actual node by adding hook. Let's say you want 2222 for all, except 2223 for few, but you don't want to write 2222 to all. Then you could write a hook that'll return 2222 on falsy value, otherwise configured. value. 
Of course you could also write 2222 for e.g. group, and if source returns nil, I'll continue searching lower precedence until it finds 2222 from group. If source returns false, it'll terminate there, and won't consider any lower precedence values.

rajven:
Why such difficulties? And then how can you not forget what you did that hook for?

ytti:
It was just an example, you don't need to do hook. You can just not configure truthy value in source for node which should use 22, and only configure truthy value for node which should use it. 

In your case:
```
router:192.168.1.1:admin:password:routeros:60022:ssh:Gateways::
router:192.168.1.2:admin:password:routeros:nil:ssh:Gateways::
router:192.168.1.3:admin:password:routeros:22:ssh:Gateways::
```

Which is in no way different to what you propose, as the field has to exist in your case too. 
Both 2 and 3 will use 22. 
If you have SQL, and you have row here, it'll natively be falsy or integer, so it'll just work. Same with JSON file.

rajven:
I'll try it tomorrow, thanks.

robertcheramy:
Just set `ssh_port` and/or `telnet_port` in vars_map.  You can see an example in the [Netgear model notes](https://github.com/ytti/oxidized/blob/master/docs/Model-Notes/Netgear.md#oxidized-config).

PR #3863, opened 2026-07-08
Remove additional POE status comment from routeros config

hendrikbl:
## Pre-Request Checklist

## Description

Remove additional POE status comment `voltage_on_poe-in` from routeros config.

### DIFF
--- a/spec/model/data/routeros#CHR_7.16#simulation.yaml
+++ b/spec/model/data/routeros#CHR_7.16#simulation.yaml
@@ -39,6 +39,7 @@ commands:
     # serial number = XXXX
     /interface ethernet
     set [ find default-name=ether1 ] disable-running-check=no
+    # poe-out status: voltage_on_poe-in
     set [ find default-name=ether2 ] disable-running-check=no
     set [ find default-name=ether3 ] disable-running-check=no
     set [ find default-name=ether4 ] disable-running-check=no
--- a/spec/model/data/routeros#CHR_7.16#simulation.yaml
+++ b/spec/model/data/routeros#CHR_7.16#simulation.yaml
@@ -39,7 +39,6 @@ commands:
     # serial number = XXXX
     /interface ethernet
     set [ find default-name=ether1 ] disable-running-check=no
-    # poe-out status: voltage_on_poe-in
     set [ find default-name=ether2 ] disable-running-check=no
     set [ find default-name=ether3 ] disable-running-check=no
     set [ find default-name=ether4 ] disable-running-check=no
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Changed
 - docker: set LANG=C.UTF-8. Fixes #3690 (@ytti)
+- routeros: remove intermittent POE `voltage_on_poe-in` comment (@hendrikbl)
 
 ### Fixed
 - siklu: allow parenthesis in prompt. Fixes #3841 (@ytti)
--- a/lib/oxidized/model/routeros.rb
+++ b/lib/oxidized/model/routeros.rb
@@ -47,7 +47,7 @@ class RouterOS < Oxidized::Model
       cfg.gsub! "# inactive time\r\n", '' # Remove time based system comment
       cfg.gsub! /# received packet from \S+ bad format\r\n/, '' # Remove intermittent VRRP/CARP collision comment
       cfg.gsub! "# poe-out status: short_circuit\r\n", '' # Remove intermittent POE short_circuit comment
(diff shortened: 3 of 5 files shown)


ytti:
This is not covered by test cases, this doesn't mean the work won't be merged, but it would be much appreciated if you could also contribute test case for this. You will also benefit, as there won't be regression for you in future.

```shell
❯ grep poe routeros\#*
routeros#L009UiGS_7.15.2#output.txt:/interface pppoe-client
routeros#L009UiGS_7.15.2#output.txt:add add-default-route=yes disabled=no interface=vlan6 name=pppoe-out1 password=password use-peer-dns=yes user=user
routeros#L009UiGS_7.15.2#output.txt:add interface=pppoe-out1 list=WAN
routeros#L009UiGS_7.15.2#simulation.yaml:    /interface pppoe-client
routeros#L009UiGS_7.15.2#simulation.yaml:    add add-default-route=yes disabled=no interface=vlan6 name=pppoe-out1 \\
routeros#L009UiGS_7.15.2#simulation.yaml:    add interface=pppoe-out1 list=WAN
```

hendrikbl:
I've added the comment to the simulation file for routeros 7.16. I'm using 7.19 so I'm not sure if that comment appeared in 7.16. If you prefer a completely new test case for 7.19 let me know,.

ytti:
Obviously what you committed does not appear on any device on any version, so it adds to confusion. 

It is fine to say that this is too big ask, if there is low appetite to spend time on this.

hendrikbl:
I can add a new test for my device and my version later, no problem. I'm happy to help :)

All I have to do for that is put the 3 yml files (output, simulation, significant changes) inside the data folder, right?

ytti:
You can generate the output file from the simulation file. And you can get the simulation file from live device. Let me know if the documents are not clear.

Your changes have no impact on significant changes, so there is no need for that file.

spec/model/data/routeros#CHR_7.16#simulation.yaml
```diff
@@ -39,6 +39,7 @@ commands:
     # serial number = XXXX
     /interface ethernet
     set [ find default-name=ether1 ] disable-running-check=no
+    # poe-out status: voltage_on_poe-in
```
systeembeheerder: CHR means CloudHostedRouter. It doubt a VM supports PoE.

ytti: It's not that, it's that it is missing the variable part, it's impossible string for any device, any version. It bothers me greatly how anyone could arrive here, what possible testing utility could impossible string give.

hendrikbl: I got that. I already said I'll create a new test for my specific device.
I did put it in a random file there to check if it was working at all. And frankly I did not read the documentation for tests in every detail to know the naming scheme, that was my mistake.

I just don't know when I find some time to create the files and update the PR.

ytti: You dont have to, it is entirely fair to say it is more commitment than you can afford. It is something that is very nice to have, not requirement, and it is better not to do at all, than to do without understanding it.

systeembeheerder:
@ytti i have a hard time believing a random comment just show up on  `export show-sensitive`

```
      # 2025-03-10 15:04:33 by RouterOS 7.18.2
      # software id = QMLS-64F2
      #
      # model = RB5009UPr+S+
      # serial number = 123456789AB
      /interface bridge
      add admin-mac=AA:11:AA:11:AA:11 auto-mac=no comment=defconf name=bridge
      /interface ethernet
      set [ find default-name=ether2 ] mtu=1514
      set [ find default-name=ether3 ] mtu=1514
      # poe-out status: voltage_on_poe-in
      set [ find default-name=ether4 ] mtu=1514
```
https://help.mikrotik.com/docs/spaces/ROS/pages/19136769/PoE-Out#PoEOut-PoE-OutMonitoring
shows the command is `interface ethernet poe monitor` not `export`.


edit:
"unfortunately" I don't have a Mikrotik with "unexpected voltage input". Personally i would fix the electrical issues and not work around it in Oxidized

ytti:
@hendrikbl comments on the @systeembeheerder concern?

hendrikbl:
@ytti @systeembeheerder it actually does, see the screenshot. It's hard to "prove" it otherwise. You're right that the issue should be fixed on the physical side, but in some cases it is not even clear why this warning appears and where this is coming from (see https://forum.mikrotik.com/t/ether6-poe-error-detected-poe-out-status-voltage-on-poe-in/265396 for others having similar issues).
Screenshot 2026-07-13 162121

As you can see in the screenshot, there are already almost 100 changes for the device. 90 of these changes happend in the last 26 days because the warning appears and disappears randomly on those interfaces, resulting in new commits everytime oxidized "catches" it.

This may be a bug, intended behaviour or a hardware defect. But whatever it is, I don't think it should be in the config saved by oxidized.

PR #3871, opened 2026-07-24
Make 'ip' field resolvable

kosherclerk:
enabled and ip_addr is present, falling back to resolving `name` only when ip_addr is nil or empty, preserving prior behavior for that case.

## Pre-Request Checklist

### DIFF
--- a/lib/oxidized/node.rb
+++ b/lib/oxidized/node.rb
@@ -20,7 +20,10 @@ module Oxidized
       logger.debug 'IPADDR %s' % ip_addr.to_s
       @name = opt[:name]
       @ip = IPAddr.new(ip_addr).to_s rescue nil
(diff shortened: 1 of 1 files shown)


ytti:
What are you trying to solve? 

I feel like this is similar as late another issue. Where the problem is actually that you want to have different name than DNS name for devices stored?

To me it seems like overloading IP by providing name for IP, it would make more sense to differentiate between dns name and display name. 

We can't probably rename 'name', so 'name' will have to remain 'dns name', but we could maybe add new field 'display_name' or 'alias' or something else. 

Logic would be, we use 'name' as 'name', unless if 'display_name' is given we use 'display_name' as 'name', after we've resolved 'name' into 'ip'? This way you provide your reachability DNS information via' name', and additionally you provide 'alias', 'display_name' or something equivalent, which ends up being used as 'name' post DNS resolution?

PR #3874, opened 2026-07-29
Improve SixWind config collection and secret filtering

hcaldicott:
## Pre-Request Checklist

## Description
This change consists of a small bugfix for password and secret redaction, as well as a new feature to support multiple config backup formats.

I received a request from another ISP to implement support for backups in the fullpath format, and this change adds some variables that allow this to be configured, while retaining the hierarchical human-readable format by default.

_I have also implemented some config value validation that I have not seen implemented in other models. When invalid values are entered, I raise Oxidized::InvalidConfig. Nothing mentions this being an issue in the contributing guidelines, but I would like feedback if there are concerns about this approach to variable validation._


robertcheramy:
Does the behaviour of the model change? `show config` is replaced with `show config nodefault`.

It would be great to include a YAML simulation file if possible.

hcaldicott:
> Does the behaviour of the model change? `show config` is replaced with `show config nodefault`.
> 
> It would be great to include a YAML simulation file if possible.

I will investigate the YAML simulation.

The nodefault command fixes a bug as well related to reapplying configuration, but I might make a quick change to make that configurable.

robertcheramy:
Producing a YAML simulation file from an existing model should be easy: https://github.com/ytti/oxidized/blob/master/docs/Issues.md#submit-a-yaml-simulation-file

PR #3880, opened 2026-08-05
source/jsonfile: handle gpg-decrypted String from open_file

### LINKED ISSUE #3879: JSONFile source with GPG fails to load
**Describe the bug**
Can not load json (encrypted) file when gpg used.

**To Reproduce**
1. Create router.db: 
```
cat >> router.db
[{"name": "device1", "ip": "192.0.2.1", "model": "fortigate", "location": "location1", "username": "user", "password": "pass"}]
```
2. Create encrypted router.db.gpg (passphrase: test): `gpg --symmetric --output router.db.gpg router.db`
3. Use part of the config below
4. Restart Oxidized
5. See `crash`

**Configuration**
```
source:
  default: jsonfile
  jsonfile:
    file: "/home/oxidized/.config/oxidized/router.db.gpg"
    gpg: true
    gpg_password: test
    map:
      name: name
      ip: ip
      model: model
      group: group
      username: username
      password: password
```

**Logs**
```
crash
--------------------------------------------------
2026-08-04 08:12:59 UTC
undefined method `read' for "[{\"name\": \"device1\", \"ip\": \"192.0.2.1\", \"model\": \"fortigate\", \"location\": \"location1\", \"username\": \"user\", \"password\": \"pass\"}]\n":String [NoMethodError]
--------------------------------------------------
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/source/jsonfile.rb:29:in `load'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/nodes.rb:18:in `block in load'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/nodes.rb:151:in `synchronize'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/nodes.rb:151:in `with_lock'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/nodes.rb:13:in `load'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/nodes.rb:146:in `initialize'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/core.rb:16:in `new'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/core.rb:16:in `initialize'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/core.rb:4:in `new'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/core.rb:4:in `new'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/lib/oxidized/cli.rb:19:in `run'
/var/lib/gems/3.2.0/gems/oxidized-0.37.0/bin/oxidized:5:in `<top (required)>'
/usr/local/bin/oxidized:25:in `load'
/usr/local/bin/oxidized:25:in `<main>'
--------------------------------------------------
```

**Running environment**
- oxidized version: 0.37.0

**Additional context**
It seems load() is trying to read a file, but when gpg is used the object is a string?
From source.rb:
```
def open_file
    file = File.expand_path(@cfg.file)
    if @cfg.gpg?
        crypto = GPGME::Crypto.new password: @cfg.gpg_password
        crypto.decrypt(File.open(file)).to_s
    else
        File.open(file)
    end
end
```

youdie006:
## Pre-Request Checklist

- [ ] Changes are reflected in the documentation (N/A - bug fix, no behavior/config change)

## Description

Closes #3879.

`Source#open_file` returns a `File` on the plain path but the decrypted contents as a **String** on the gpg path (`crypto.decrypt(...).to_s`). `JSONFile#load` called `open_file.read` unconditionally, so `source: jsonfile` combined with `gpg: true` crashed on startup with `NoMethodError: undefined method 'read' for an instance of String` instead of loading the device list. The `CSV` source avoids this because its `#load` uses `#each_line`, which both `String` and `File` respond to.

This reads the contents only when the object responds to `#read`, otherwise uses it as-is, so both the plain and gpg paths work.

Added a `#load` spec that stubs `open_file` to return a JSON String (the gpg case) and asserts the nodes parse. Verified red before the fix (`NoMethodError` at `jsonfile.rb`) and green after. All `spec/source/*` specs pass and `rubocop` reports no offenses locally.

Developed with the assistance of Claude Code (AI); reviewed and verified by me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - fortigate: prompt can contain HA cluster status. Fixes #3846 (@robertcheramy)
 - aoscx: hide power consumption on rows where PSU output is N/A. Fixes #3864 (@FusionBrah)
 - aoscx: hide input voltage readings in "show environment power-supply input-voltage". Fixes #3864 (@FusionBrah)
-- source/jsonfile: load the device list when combined with `gpg` encryption instead of crashing with a `NoMethodError`. Fixes #3879 (@youdie006)
+- source: return the decrypted contents of a `gpg`-encrypted source as an IO object, so JSONFile (and any source reading via `#read`) loads it instead of crashing with a `NoMethodError`. Fixes #3879 (@youdie006)
 
 ## [0.37.0 - 2026-05-20]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - fortigate: prompt can contain HA cluster status. Fixes #3846 (@robertcheramy)
 - aoscx: hide power consumption on rows where PSU output is N/A. Fixes #3864 (@FusionBrah)
 - aoscx: hide input voltage readings in "show environment power-supply input-voltage". Fixes #3864 (@FusionBrah)
+- source/jsonfile: load the device list when combined with `gpg` encryption instead of crashing with a `NoMethodError`. Fixes #3879 (@youdie006)
 
 ## [0.37.0 - 2026-05-20]
 ### Added
--- a/lib/oxidized/source/jsonfile.rb
+++ b/lib/oxidized/source/jsonfile.rb
@@ -26,11 +26,7 @@ module Oxidized
       end
 
       def load(*)
-        file = open_file
-        # open_file returns a File on the plain path but a String on the gpg
-        # path (see Source#open_file), so only call #read when it is supported.
-        contents = file.respond_to?(:read) ? file.read : file
-        data = JSON.parse(contents)
+        data = JSON.parse(open_file.read)
         data = string_navigate_object(data, @cfg.hosts_location) if @cfg.hosts_location?
 
         transform_json(data)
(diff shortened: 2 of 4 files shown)


ytti:
I think the openfile should work consistently, either return string (probably bad idea) or IO object (probably good idea). If GPG cannot return IO natively (it probably can), we can always use IOString to turn string back into IO object, which can be then read.

Unfortunately fixing openfile means we have to fix all call sites as well, if call sites do what you're proposing to do.

youdie006:
Good call - reworked to fix `open_file` itself. It now wraps the gpg-decrypted string in a `StringIO`, so it always returns an IO object like the plain `File.open` path.

No call sites need changing: `JSONFile` reads via `#read` and `CSV` via `#each_line`, and `StringIO` supports both, so every source consumes `open_file` uniformly. I reverted the `JSONFile#load` guard, and the test now asserts `open_file` returns a readable IO for a gpg-encrypted source. `rubocop` and the source specs pass.

ytti:
I have strong reason to suspect the author is not real human who actually uses oxidized, and is possibly AI shop using OSS project for training purposes or some other motivation rather.

This should be disclosed, expectation is that real people are fixing actual problems they've had running the software. 

The problem is real, and contribution is fine. But motivation and not being human should be disclosed. 

Writing OSS software makes whole lot less sense, when your work is being consumed by trillion dollar LLM shops, rather than real human with real problem they want to solve.

PR #3886, opened 2026-08-09
model/fastiron: answer pager prompt and strip ANSI escape sequences

### LINKED ISSUE #3819: Handle pager in FastIron model
- [x] I have read and understood [docs/Issues.md](https://github.com/ytti/oxidized/blob/master/docs/Issues.md).

**Is your feature request related to a problem? Please describe.**
We are experiencing issues when backing up the config of a Ruckus ICX 7150-48P with the FastIron model, because there is a prompt from the pager in the way which cannot be skipped.

Changing to a different model has not resolved the problem we experienced and other devices of the same line were able to be handled.

**Describe the solution you'd like**
Adding the following code to the FastIron Model has resolved the issue for us.

```
  expect /--More--|Press any key to continue|Press <space> to continue/i do |data, re|
    send ' '
    data.sub re, ''
  end

  expect /\e\[[\d;]*[A-Za-z]/ do |data, re|
    data.gsub(re, '')
  end
```

**Describe alternatives you've considered**
Any other solution which can handle a pager (which cannot be disable and using "enable" is also not an option because of our security.)

**Additional context**
Other device types which were solved as well with the above code:
- `ICX 7850-48F`
- `ICX 7150-48PF`

Unaffected devices where of the type: `ICX 7850-48F` (yes... same model as above)

FusionBrah:
## Pre-Request Checklist

## Description

On Ruckus ICX devices, the pager cannot always be disabled with `skip-page-display`, for example when the user has no enable rights. This adds two `expect` blocks to the fastiron model, as proposed and tested by @cr-bruderer in #3819: answer the pager prompt with a space and strip the ANSI escape sequences the pager uses to redraw the screen. The existing `skip-page-display` post_login is kept, as it still works for users with sufficient rights.

Tests:
* `pass_with_expect` prompt tests with ANSI-prefixed prompts
* a simulation of an ICX 7150-48P where `skip-page-display` is rejected and `show running-config` is interrupted by a `--More--` pager prompt

Note: the simulation is hand-built, not captured from a real device with device2yaml, as I have no access to an affected device. I will ask the reporter to test this branch on real hardware.

Fixes #3819

### DIFF
--- a/lib/oxidized/model/fastiron.rb
+++ b/lib/oxidized/model/fastiron.rb
@@ -2,7 +2,19 @@ class FastIron < Oxidized::Model
   using Refinements
 
   prompt /^([\w.@\/()-]+[#>]\s?)$/
-  comment  '! '
+  comment '! '
+
+  # The pager cannot always be disabled with 'skip-page-display', for example
+  # when the user has no enable rights. See issue #3819
+  expect /--More--|Press any key to continue|Press <space> to continue/i do |data, re|
+    send ' '
+    data.sub re, ''
+  end
+
+  # Remove ANSI escape sequences used by the pager to redraw the screen
+  expect /\e\[[\d;]*[A-Za-z]/ do |data, re|
+    data.gsub(re, '')
+  end
 
   cmd :all do |cfg|
     # cfg.gsub! /\cH+\s{8}/, ''         # example how to handle pager
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Fixed
 - eos: hide the community string / v3 user in snmp-server host lines. Fixes #3882 (@FusionBrah)
 - aoscx: mask the community string in snmp-server host lines instead of the token after the host address. Fixes #3881 (@FusionBrah)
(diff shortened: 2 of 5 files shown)


robertcheramy:
Should we wait for a reaction of @cr-bruderer ?

lib/oxidized/model/fastiron.rb
```diff
@@ -2,7 +2,19 @@ class FastIron < Oxidized::Model
...
+  # when the user has no enable rights. See issue #3819
+  expect /--More--|Press any key to continue|Press <space> to continue/i do |data, re|
+    send ' '
+    data.sub re, ''
+  end
+
+  # Remove ANSI escape sequences used by the pager to redraw the screen
+  expect /\e\[[\d;]*[A-Za-z]/ do |data, re|
+    data.gsub(re, '')
+  end
```
robertcheramy: Consider using `clean :escape_codes` https://github.com/ytti/oxidized/blob/master/docs/Creating-Models.md#remove-ansi-escape-codes

FusionBrah: Fixed in 02809dbf: replaced the FastIron ANSI-stripping expect block with clean :escape_codes, keeping the pager response unchanged. FastIron tests, full test suite, and RuboCop pass.

PR #3896, opened 2026-08-25
tmos: strip churn-only fields from all command output

dahlmo:
## Pre-Request Checklist

## Description

TMOS command output includes several fields — revision counters, checksum SHA1 hashes, size, create-time/last-update time timestamps, and updated-by/created-by identities — that change on every poll without any real configuration change, producing noisy diffs.

Add a cmd :all block so these are stripped unconditionally from every command's output, rather than only when remove_secret is enabled.

### DIFF
--- a/lib/oxidized/model/tmos.rb
+++ b/lib/oxidized/model/tmos.rb
@@ -3,6 +3,12 @@ class TMOS < Oxidized::Model
 
   comment '# '
 
+  cmd :all do |cfg|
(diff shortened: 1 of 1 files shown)


dahlmo:
Any chance someone could take a look at this? Happy to address any feedback.

robertcheramy:
Hi @dahlmo and thank you for your PR.
- Please document your change in CHANGELOG.md
-  are the fields present in each cmd output or only in a few one? If present in each cmd output, `cmd :all` is the right place. If not, you should clean the output of the affected command.
- Please include a YAML simulation file (just run the model with input.debug=yaml and clean the YAML file from sensitive data)

PR #3908, opened 2026-09-08
Add Cisco FTD model

mgrocock-cwcs:
## Pre-Request Checklist

## Description

Add a new model for backing up Cisco FTD firewalls via the HTTP API. Includes supporting changes to the HTTP input:

* Support overriding port
* Support DELETE method


ytti:
I'm not familiar with the model. Why is the delete of the config needed?

Also, why are we using vars(ip) instead of node.ip? I think we have few other cases where problem is that users want to use name, that is not resolved for connecting, but files are stored at such name.

And this is valid issue, and if this is what is going on, then I think right solution is, that for the cases where people want some display_name that issued and another IP or mgmt_name that is used to connect, we need to support this on node level, not vars. 
So it would be separate issue.

For example we could have node.mgmt_name, and if that is defined, we leave node.name alone, and resolve (if name) this to IP we'll connect to.

mgrocock-cwcs:
The configexport API call saves a backup to the device as a Zip file with the specified filename (oxidized.zip in this case). This file is then downloaded, and the JSON configuration extracted from it. The first delete is to ensure that there is no existing backup with the same filename, otherwise the configexport would fail, and the second is to tidy up afterwards.

I considered including a timestamp or random component in the filename, but decided against this as it could result in a buildup of old backups on the device if jobs are interrupted for any reason, and instead took the approach of using a static filename that was unlikely to be used for anything else.

We use vars(:ip) as the devices we're backing up generally have SNMP responding on the outside interface, so this is the address we use when adding them to LibreNMS, but the HTTP API that Oxidized needs to connect to is listening on the management interface. Our approach is to define mapping rules in LibreNMS that assign each device to its own group, and then to set the ip variable for each group in /etc/oxidized/config.

If vars(:ip) is undefined, then it will default to using @node.ip.

ytti:
Sorry I think you need to elaborate more.

why cannot you use node.ip and/or node.name? 

The common reason seems to be that 
a) name does not resolve, but is desirable to use as filename
b) management name does resolve, but is not desirable to use as filename

currently we are not handling that case, and that case could justify adding node.mgmt_name, which is then resolved into node.ip. But node.name still used for storage. 

So I just want to understand specifics of your case, so that however we solve the existing gap also contains solution to your problem. 
It almost guaranteed solution isn't models using vars(:ip). I'm struggling to see reason to do that, having node.ip that works seems to be minimum requirement. I could potentially understand if you need node.ip for some connection and you need additional ip for some other case, but that doesn't seem to be case here, you're either using node.ip or vars(:ip), at which point, why not just use node.ip. 

We already today support hooks for sources, where you can pick-up e.g. node.ip or node.name conditionally from source, which could also cover your issue.

mgrocock-cwcs:
You're right, we can simply use a mapping rule in LibreNMS to ensure that @node.ip is set appropriately. I have updated the pull request so that it no longer uses vars(:ip).

ytti:
Can you check the Dockerfile and make a judgement call if the ruby zip should be added there or not. It'll either use distribution packaged if added there, or gem if not. 
I think idea is, if distribution has it, it is preferred. But I'm not making any requirements.

ytti:
Are you sure raising the error is prudent? I don't think this will be rescued by input during collection? That is, I think entire oxidized will crash when you raise, and I feel like you only intended the particularly run to fail, not oxidized to fail?

I think it is reasonable to raise error from class, but there probably should be some subclass which then gets :warn, instead of crash curing fetch. Like InputFailed, GetFailed, FetchFailed (I don't know if these are good names, I didn't think about it much). 
Possibly inheritance such as ThisError inherits InputError, which Inherits OxidizedError. And then ThisError recovered if raised.

ytti:
Tiny nitpick, but generally I like small methods, with focused logic and as little repetition as possible. So things you could consider (not making any asks). Are like:

```ruby
def check_job_status(job)
  return if job['status'] == 'SUCCESS'

  message = case
            when job['status'] == 'FAILED'
              job['statusMessage']
            when job['error']
              job.dig('error', 'messages', 0, 'description')
            else
              'unknown error'
            end

  raise Oxidized::SomeSoftErrorThatWeCatchAndWarnInsteadOfCrash, message
end
```

There are some other opportunities for what I perceive as clarity and readability, but again, not making any mandates here, it'll be merged anyway once we sort the raising and docker issue. 

For example / isn't valid variable name, so I think brackets are redundant and you can have just "foo/#@bar/baz" and it'll work, because / terminates the bar.

docs/Supported-OS-Types.md
```diff
@@ -55,6 +55,7 @@
 |                    |Cisco Catalyst Express        |[ciscoce](/lib/oxidized/model/ciscoce.rb)
 |                    |ExaLink Fusion (Nexus 3550-F) |[exalink](/lib/oxidized/model/exalink.rb)        |@obol89          |[ExaLink](Model-Notes/ExaLink.md)
 |                    |FireLinuxOS                   |[firelinuxos](/lib/oxidized/model/firelinuxos.rb)
+|                    |FTD                           |[ftd](/lib/oxidized/model/ftd.rb)                |                 |[FTD](Model-Notes/FTD.md)
```
robertcheramy, line 58: As FireLinuxOS and FTD are the same device an one is accessed per ssh and the other per http, it should appear here.

Maybe FireLinuxOS => FTD via SSH and FTP => FTD via HTTP ?

oxidized.gemspec
```diff
@@ -41,6 +41,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'net-ssh',              '~> 7.3'
   s.add_dependency 'net-telnet',           '~> 0.2'
   s.add_dependency 'psych',                '~> 5.0'
+  s.add_dependency 'rubyzip',              '~> 1.2'
```
robertcheramy, line 44: rubyzip in debian Trixie is version 2.3.2. Upstream already has 3.6.0

The dependency should be >= 2.3.2, < 4

lib/oxidized/model/ftd.rb
```diff
@@ -0,0 +1,84 @@
...
+      'doNotEncrypt'        => true,
+      'deployedObjectsOnly' => true
+    }.to_json
+
+    obj_id = JSON.parse(post_http("#{@api_endpoint}/action/configexport", payload))['jobHistoryUuid']
+    job_status = nil
+
+    @retries.times do
+      job_status = JSON.parse(get_http("#{@api_endpoint}/jobs/configexportstatus/#{obj_id}"))
+      break if %w[SUCCESS FAILED].include?(job_status['status'])
```
robertcheramy, line 36: job_status could be nil and should be checked before accessing it.

lib/oxidized/model/ftd.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+
+    if job_status['status'] != 'SUCCESS'
+      if job_status['status'] == 'FAILED'
+        raise Oxidized::OxidizedError, job_status['statusMessage']
+      elsif job_status['error']
+        raise Oxidized::OxidizedError, job_status['error']['messages'][0]['description']
+      else
+        raise Oxidized::OxidizedError, 'unknown error'
+      end
```
robertcheramy, line 48: More defensive coding:
```ruby
  else
    description = job_status.dig('error', 'messages', 0, 'description')
    raise Oxidized::OxidizedError, description || 'unknown error'
  end
```

lib/oxidized/model/ftd.rb
```diff
@@ -0,0 +1,84 @@
...
+    job_status = nil
+
+    @retries.times do
+      job_status = JSON.parse(get_http("#{@api_endpoint}/jobs/configexportstatus/#{obj_id}"))
+      break if %w[SUCCESS FAILED].include?(job_status['status'])
+
+      sleep(@retry_secs)
+    end
+
+    if job_status['status'] != 'SUCCESS'
```
robertcheramy, line 41: defensive coding:
```ruby
status = job_status&.dig('status')
raise Oxidized::OxidizedError, 'unknown error' unless status

if status != 'SUCCESS'
```

lib/oxidized/model/ftd.rb
```diff
@@ -0,0 +1,84 @@
...
+    delete_http("#{@api_endpoint}/action/configfiles/#{@config_filename}")
+
+    # generatedOn contains the timestamp of the config export. Delete it to avoid unnecessary differences.
+    config[0].delete('generatedOn')
+
+    # The distiniguishedNames list seems to change order between exports. Sort it deterministically to avoid unnecessary differences.
+    index = config.find_index { |element| element['type'] == 'identitywrapper' and element['data']['type'] == 'distinguishednamegroup' }
+
+    if index
+      config[index]['data']['distiniguishedNames'].sort_by! { |element| Digest::MD5.hexdigest(element['id']).to_i(16) }
```
robertcheramy, line 62: Do you mean `distinguished` ?

lib/oxidized/model/ftd.rb
```diff
@@ -0,0 +1,84 @@
...
+      if job_status['status'] == 'FAILED'
+        raise Oxidized::OxidizedError, job_status['statusMessage']
+      elsif job_status['error']
+        raise Oxidized::OxidizedError, job_status['error']['messages'][0]['description']
+      else
+        raise Oxidized::OxidizedError, 'unknown error'
+      end
+    end
+
+    config = JSON.parse(Zip::File.open_buffer(get_http("#{@api_endpoint}/action/downloadconfigfile/#{@config_filename}")).read('full_config.txt'))
```
robertcheramy, line 51: This might crash if we get something else than a zip or it does not contain full_config.txt

robertcheramy (changes requested):
It would also be great to update the tests for input/http for the new features (delete + custom port).

lib/oxidized/model/ftd.rb
```diff
@@ -0,0 +1,84 @@
...
+    index = config.find_index { |element| element['type'] == 'identitywrapper' and element['data']['type'] == 'distinguishednamegroup' }
+
+    if index
+      config[index]['data']['distiniguishedNames'].sort_by! { |element| Digest::MD5.hexdigest(element['id']).to_i(16) }
+    end
+
+    JSON.pretty_generate(config)
+  end
+
+  cmd cfg_cb
```
ytti, line 68: Perhaps the configuration manipulation code should be here, instead of in the callback itself?

mgrocock-cwcs:
> Can you check the Dockerfile and make a judgement call if the ruby zip should be added there or not. It'll either use distribution packaged if added there, or gem if not. I think idea is, if distribution has it, it is preferred. But I'm not making any requirements.

I've added ruby-zip to the Dockerfile.

mgrocock-cwcs:
> Are you sure raising the error is prudent? I don't think this will be rescued by input during collection? That is, I think entire oxidized will crash when you raise, and I feel like you only intended the particularly run to fail, not oxidized to fail?
> 
> I think it is reasonable to raise error from class, but there probably should be some subclass which then gets :warn, instead of crash curing fetch. Like InputFailed, GetFailed, FetchFailed (I don't know if these are good names, I didn't think about it much). Possibly inheritance such as ThisError inherits InputError, which Inherits OxidizedError. And then ThisError recovered if raised.

I've now realised that I can fail the run without causing a crash by simply returning false. The approach I've taken is to create and FTDError class that inherits from OxidizedError, raise that where appropriate, and then rescue it at the end of the callback, log the message and return false.

mgrocock-cwcs:
> Tiny nitpick, but generally I like small methods, with focused logic and as little repetition as possible. So things you could consider (not making any asks). Are like:
> 
> ```ruby
> def check_job_status(job)
>   return if job['status'] == 'SUCCESS'
> 
>   message = case
>             when job['status'] == 'FAILED'
>               job['statusMessage']
>             when job['error']
>               job.dig('error', 'messages', 0, 'description')
>             else
>               'unknown error'
>             end
> 
>   raise Oxidized::SomeSoftErrorThatWeCatchAndWarnInsteadOfCrash, message
> end
> ```
> 
> There are some other opportunities for what I perceive as clarity and readability, but again, not making any mandates here, it'll be merged anyway once we sort the raising and docker issue.
> 
> For example / isn't valid variable name, so I think brackets are redundant and you can have just "foo/#@bar/baz" and it'll work, because / terminates the bar.

I've separated the code out into smaller methods, split some longer lines up to make them more readable, and amended check_job_status along the lines suggested.

The suggested change from "foo/#{@bar}/baz" to "foo/#@bar/baz" upsets rubocop, so I've left those alone.

mgrocock-cwcs:
> It would also be great to update the tests for input/http for the new features (delete + custom port).

I've added tests for the custom port and DELETE method in spec/input/http_spec.rb. I re-worked the way I'd implemented the custom port slightly to fit better with the existing tests.

docs/Supported-OS-Types.md
mgrocock-cwcs: I've updated docs/Supported-OS-Types.md as suggested.

oxidized.gemspec
mgrocock-cwcs: I've updated oxidized.gemspec with the correct dependency.

lib/oxidized/model/ftd.rb
mgrocock-cwcs: I've added checks for job_status being nil.

lib/oxidized/model/ftd.rb
mgrocock-cwcs: I've re-worked the error-checking code to use dig.

lib/oxidized/model/ftd.rb
mgrocock-cwcs: I've re-worked check_job_status to be more defensively coded.

lib/oxidized/model/ftd.rb
mgrocock-cwcs: It is actually 'distiniguishedNames' not 'distinguishedNames' in the JSON output from the FTD. I've added a comment to confirm that this is intentional.

lib/oxidized/model/ftd.rb
mgrocock-cwcs: I've added some error-checking here for it not being a valid zip file, the zip file not containing full_config.txt, and JSON parsing failing.

lib/oxidized/model/ftd.rb
mgrocock-cwcs: I've moved this code from the callback to a block passed to command. I also found another instance of a list not being deterministically sorted, so I've generalised the code and added this instance.

oxidized.gemspec
robertcheramy: It should be  should be `>= 2.3.2`, not `~> 2.3.2.` (which fixes the gem to 2.3.*).

robertcheramy: Fixed in 8d28311a1c8fbe76639b232dcc5133c64f1740dd

robertcheramy (approved):
The PR ist OK for me. @ytti - for you OK?

Note: the sleep(`@poll_wait`) in the polling loop runs inside the global `timelimit` (default: 300s), which wraps the entire node run.
If a user increases values, ftd_polls × ftd_poll_wait (plus HTTP round-trips) could exceed timelimit, causing the job to be aborted with status timelimit. I'm sure if it worth noting this in FD.md, as the default values are way below the global limit.

ytti:
From style perspective, I’m not happy. It feels strange how the methods are made, and wrapping them inside method feels ugly to me.
But these are subjective, and models arent beaty competition. 

If @robertcheramy is fine by it, I’m fine by it.

robertcheramy:
I want @ytti to be happy ;-)

I looked at the code under your comment, and I agree. I've changed the structure with last commit. Better ?

@mgrocock-cwcs - can you test against a real device ?

ytti:
Looks cleaner now, thank you.

mgrocock-cwcs:
I can confirm that this is working with a real device.

